PR #52, opened 2020-08-20
Add dependencies and config for auto documentation.

Tanmay06:
Modified poetry project and lock files to install sphinx dependencies. Also modified README.md with directions to generate documentations. There are still some changes to be done to capture package level docstrings.


ethanjli:
Should we try to finish this PR by just merging in what we have now and creating a future Github Issues item to look into using a Makefile to simplify building the docs, or would you prefer to just close this PR without merging?

Tanmay06:
I'm almost done making a MAKEFILE for this PR. I believe we can close this today.

Tanmay06:
I think this is ready now. One note, the module starting with `_`, like `_trio.py`, is ignored by the auto-module plugin. In future we need to find a way overcome or rename the modules.

Tanmay06:
Sphinx auto-documentation requires `python3-sphinx` to be present in the system. It can be done using `apt-get install python3-sphinx`. But since it is a dev dependency rather than production, I'm not sure whether it should be added to the deployment process or not. I guess @Praneet9 can suggest a better solution for this.

Praneet9:
> Sphinx auto-documentation requires `python3-sphinx` to be present in the system. It can be done using `apt-get install python3-sphinx`. But since it is a dev dependency rather than production, I'm not sure whether it should be added to the deployment process or not. I guess @Praneet9 can suggest a better solution for this.

We have a development environment setup.
As this is a dev env requirement we can add this to poetry installation script as it already installs all other packages.
Another way of doing it would be to just add it to setup.sh which does the development environment setup.

Tanmay06:
Another possible way that I see could be to just add the installation step in `backend/docs/README.md`. Since this not a necessary dependency for development it could be skipped from deployment process and explicitly installed whenever the need arises.

Tanmay06:
I not sure why but `pylint` has started throwing this `raise-missing-from` warnings during the checks. This happened a couple of times before on my local machine but some how it got magically resolved.

ethanjli:
The docs build works for me. Can you see if the instructions provided at https://stackoverflow.com/a/63852262 will solve the pylint complaints?

Before you merge, can you add/modify a `.gitignore` to remove any auto-generated build files?

For records-keeping:

Tanmay06:
For records-keeping:

Tanmay06:
I have made the necessary changes to the branch and also fixed the `pylint` warnings.

ethanjli (approved):

Issue #79, opened 2020-08-29
Low-level driver for barometric pressure sensor


ethanjli:
We need a low-level driver for talking to the Bosch BMP388 for measuring barometric pressure.

PR #93, opened 2020-09-02
Low-level driver for barometric pressure sensor (#79)

hemanthgs:
Added files BMP388.h and BMP388.cpp
class SPIBMP388 is created and updated with following methods
1. getChipId
2. setIIRFilter
3. setPowerCtrlSettings
4. write
5. writeRead
6. spiReset
7. ReadCalibrationData
8. ReadTemperatureData
9. ReadPressureData


firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+
+#pragma once
+
+#include "Pufferfish/HAL/STM32/HALSPIDevice.h"
+
+namespace Pufferfish {
+namespace Driver {
+namespace BMP388 {
+
+class SPIBMP388 {
```
ethanjli: Is there a specific reason to include `SPI` in the name of this class, and is there a specific reason to put this class inside a `BMP388` namespace? I ask because I don't remember seeing anything like this for any of the other drivers.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+   */
+  SPIBMP388Status write(uint8_t *buff, size_t count);
+
+  /**
+   * @brief  Writes and read the data from the register
+   * @param  txbuf data to write into DI pin
+   * @param  rxbuf data read from DO pin
+   * @return SPIBMP388Status returns ok/readError
+   */
+  SPIBMP388Status writeRead(uint8_t *txbuf, uint8_t *rxbuf, size_t count);
```
ethanjli: Do these need to be public methods, or can they be private?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+   * @return returns the temperature data
+   */
+  uint32_t ReadTemperatureData();
+
+  /**
+   * @brief  reads the pressure data from sensor
+   * @param  None
+   * @return returns the pressure data
+   */
+  uint32_t ReadPressureData();
```
ethanjli: Is it possible for any of these operations to fail? If so, do we need to return an error code or status code?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+    double par_p4;
+    double par_p5;
+    double par_p6;
+    double par_p7;
+    double par_p8;
+    double par_p9;
+    double par_p10;
+    double par_p11;
+    double t_lin;
+  };
```
ethanjli: Which of these can be made private?

Another thing which we have not yet decided, so I'd like to hear what you think are pros and cons of each choice: some of these structs and enums could be moved into the namespace level, or we could keep them in the public section of the class. I have done both options when implementing drivers for this project (but not in a self-consistent way); I think we should standardize how we do this. What is your opinion on where these kinds of structs and enums should go?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setIIRFilter(bool iirFilterEnabel, IIRFilterCoefficient IIRFilterCoeff);
+
+  /**
+   * @brief  Sets the power control settings using PWR_CTRL reg
+   * @param  bmp388Value structure contains the power control parameters
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setPowerCtrlSettings(struct bmp388Settings bmp388Value);
```
ethanjli: `bmp388Value` should be passed as a const reference if possible.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+   */
+  SPIBMP388Status getChipId(uint8_t &chip_id);
+
+  /**
+   * @brief  Enables/disables the setIIRFilter based on IIRFilter Coefficient
+   * @param  iirFilterEnabel Enables/disables the IIRFilter
+   * @param  IIRFilterCoeff input coefficient for IIR filer for enabling
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setIIRFilter(bool iirFilterEnabel, IIRFilterCoefficient IIRFilterCoeff);
```
ethanjli: `IIRFilterCoeff` should be passed as a const reference if possible.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status spiReset();
+
+  /**
+   * @brief  reads the CalibrationData data from NVM
+   * @param  QuantizedCalibData updates the structure bmp388QuantizedCalibData
+   * @return None
+   */
+  void ReadCalibrationData(struct bmp388QuantizedCalibData QuantizedCalibData);
```
ethanjli: `QuantizedCalibData` should be passed as a const reference if possible.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
+/*
+ * BMP388.h
+ * 
+ * Created on: Aug 29, 2020
+ *      Author: pradip.jm
+ */
+
+#pragma once
+
+#include "Pufferfish/HAL/STM32/HALSPIDevice.h"
```
ethanjli: Drivers should never include STM32-specific HAL files; instead, drivers should only include `Pufferfish/HAL/HAL.h`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,212 @@
...
+
+namespace Pufferfish {
+namespace Driver {
+namespace BMP388 {
+
+class SPIBMP388 {
+public:
+  SPIBMP388 (HAL::HALSPIDevice &spi):mSpi(spi){
```
ethanjli: Drivers should never take as parameters any HAL modules which depend on the STM32 HAL; instead, they should take the respective abstract interfaces.
```suggestion
  SPIBMP388 (HAL::SPIDevice &spi):mSpi(spi){
```

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Updated the namespace with SPI (As we are using SPI protocol).
Changed class name to BMP388.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Yes, moved the write and writeRead to private, as they are called only from the public methods of BMP388 class.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Updated with return values for read operation result.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Moved few structures to private.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Updated the bmp388Settings variabel input as constant.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Updated the IIRFilterCoeff input variable as constant.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: QuantizedCalibData is taken as private variable. As it updates the QuantizedCalibData in begining and need to use it in multiple methods.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: Updated with "Pufferfish/HAL/Interfaces/SPIDevice.h" inclusion.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
hemanthgs: updated HAL::HALSPIDevice to HAL::SPIDevice

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
ethanjli: Great! This will be consistent with the way we use an `I2C` namespace to hold all the I2C drivers.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+namespace Pufferfish {
+namespace Driver {
+namespace SPI {
+
+class BMP388 {
+public:
+  /**
+   * @brief bmp388 sensor structure for temperature and pressure and mode settings.
+   */
+  struct bmp388Settings{
```
ethanjli: If you declared this struct inside the `BMP388` class, so you can rename it to just `Settings`, so that external code accesses it through `Pufferfish::Driver::SPI::BMP388::Settings` instead of `Pufferfish::Driver::SPI::BMP388::bmp388Settings`. If you were to instead declare this struct outside the `BMP388` class (which is up to you, I think it's fine to declare it inside the class), then it would make sense to name this struct `BMP388Settings`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+    iirFilter63,
+    iirFilter127
+  };
+
+
+  /**
+  * @ Time standby in the Output Data Rate (ODR) register
+  */
+  enum class TimeStandby {
+    TimeStandby5Ms,
```
ethanjli: Would it be possible to use shorter names for these? I understand that `5Ms` is not an allowed identifier under C/C++, but maybe we can call it `time5Ms` something - right now, to compare a variable of this type `foo`, we would need to use `if (foo == TimeStandby::TimeStandby5Ms)` which is longer than we need.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+  BMP388 (HAL::SPIDevice &spi):mSpi(spi){
+   }
+  /**
+   * @brief  Settings
+   * @param
+   * @return SPIBMP388Status returns ok/writeError
+   */
+   SPIBMP388Status bmp388ConfigSettings(const bool enableIIRFilter,
+                                    const IIRFilterCoefficient IIRFilterCoeff,
+                                    const struct bmp388Settings setBMP388PwrControl);
```
ethanjli: You should be able to use `const bmp388Settings &`, right?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setIIRFilter(bool iirFilterEnable, const IIRFilterCoefficient IIRFilterCoeff);
+
+  /**
+   * @brief  Sets the power control settings using PWR_CTRL reg
+   * @param  bmp388Value structure contains the power control parameters
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setPowerCtrlSettings(const struct bmp388Settings setBMP388PwrControl);
```
ethanjli: You should be able to use `const bmp388Settings &`, right?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+
+  /**
+   * @brief  Resets the sensor device
+   * @param  None
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status spiReset();
```
ethanjli: Would it be descriptive enough to just name this method `reset`?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+    iirFilter31,
+    iirFilter63,
+    iirFilter127
+  };
+
+
+  /**
+  * @ Time standby in the Output Data Rate (ODR) register
+  */
+  enum class TimeStandby {
```
ethanjli: Would this name be more consistent with everything else if it were named `StandbyTime` instead?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+   */
+  SPIBMP388Status setTimeStandby(enum TimeStandby timeStandby);
+
+  /**
+   * @brief  Set the pressure and temperature over sampling
+   * @param  over sampling for pressure : OFF | X1 | X2 | X4 | X8 | X16 | X32
+   * @param  over sampling for temperature : OFF | X1 | X2 | X4 | X8 | X16 | X32
+   * @return SPIBMP388Status returns OK/writeError
+   */
+  SPIBMP388Status setOversamplingRegister(uint8_t presOversampling, uint8_t tempOversampling);
```
ethanjli: Can we use arbitrary `uint8_t` values for `presOversampling` and `tempOversampling`, or are the values restricted (the way they would be from an `enum class`?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+  SPIBMP388Status enableFIFO(bool tempEnable,
+                            bool pressEnable,
+                            bool timeEnable,
+                            bool stopOnFull);
+  /**
+   * @brief  Get the FIFO length
+   * @param  address of FIFO length variable
+   * @return SPIBMP388Status returns OK/writeError
+   */
+  SPIBMP388Status getFIFOLength();
```
ethanjli: How is the FIFO length returned from this method?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+   * @return SPIBMP388Status returns ok/readError
+   */
+  SPIBMP388Status writeRead(uint8_t *txbuf, uint8_t *rxbuf, size_t count);
+
+  /**
+   * @brief  reads the CalibrationData data from NVM
+   * @param  QuantizedCalibData updates the structure bmp388QuantizedCalibData
+   * @return None
+   */
+  void ReadCalibrationData();
```
ethanjli: Is it possible for this operations to fail? If so, do we need to return an error code or status code?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+  struct bmp388QuantizedCalibData QuantizedCalibData;
+
+  /* Uncompensated temperature read from sensor */
+  uint32_t uncompensatedTemperature;
+
+  /* Uncompensated pressure read from sensor */
+  uint32_t uncompensatedPressure;
+
+  /* Fifo buffer length of the BMP388 */
+  uint16_t fifoLength;
```
ethanjli: If at all possible, these private members (`uncompensatedTemperature`, `uncompensatedPressure`, and `fifoLength`) should be removed; and the methods which are currently writing to them should return the corresponding values as output parameters (e.g. `SPIBMP388Status readUncompensatedPressure(uint32_t &uncompensatedPressure)`); and the methods which are currently reading them should instead just get the value from the output parameters of the corresponding method calls (e.g. when `readCompensatedPressure` calls `readUncompensatedPressure`).

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,249 @@
...
+   * @return None
+   */
+  void ReadCalibrationData();
+
+  /* mSpi object is updated by constructor used to read/write data through SPI */
+  HAL::SPIDevice &mSpi;
+
+  /* QuantizedCalibData structure is updated from register used for
+   * pressure and temperature compensation */
+  struct bmp388QuantizedCalibData QuantizedCalibData;
```
ethanjli: Usually I would suggest that the `readCalibrationData` method should return any results (e.g. a `QuantizedCalibData`) as an output parameter, and then those results should be then be passed as input parameters into any methods which use them (i.e. `readCompensatedPressure` and `readCompensatedTemperature`). However, we should only do that if it's reasonable. Can you think of some reasons why it might make more sense to just save the calibration data in a private member, as you've done here?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,698 @@
...
+
+#include "Pufferfish/HAL/Interfaces/SPIDevice.h"
+#include <stdint.h>
+#include <type_traits>
+
+namespace Pufferfish {
+namespace Driver {
+namespace SPI {
+
+/**
```
ethanjli, line 29: Wherever possible, we should avoid using `union`s. For all of these `union`s, I would suggest instead just writing some small reusable functions to do any manipulations you need. Clang-tidy will complain about the use of unions, so we would need to justify every single union here.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,698 @@
...
+  highResolution,  /// 19bit / 0.33Pa
+  ultraHighResolution,  /// 20bit / 0.17Pa
+  highestResolution  /// 21bit / 0.085Pa
+};
+
+/**
+ * Trimming coefficients from nvm to calculate compensated data
+ */
+union TrimmingCoefficients {
+  struct __attribute__((packed)) {
```
ethanjli: We cannot depend on `__attribute__` or other GCC extensions.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,698 @@
...
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status getErrors(SensorFaults fault);
+
+  /**
+   * @brief  read sensor status register address is 0x03
+   * @param  status command ready or pressure ready, temperature ready
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status getSensorstatus(SensorStatus status);
```
ethanjli: If `fault` and `status` are supposed to be output parameters, they need to be passed by reference, not by value.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
```diff
@@ -0,0 +1,698 @@
...
+   */
+  SPIBMP388Status readRegister(RegisterAddress address, uint8_t &value);
+
+  /**
+   * @brief  write to register
+   * @param  RegisterAddress register address
+   * @param  value write data to address
+   * @return SPIBMP388Status returns ok/writeError
+   */
+  SPIBMP388Status setRegister(RegisterAddress address, uint8_t value);
```
ethanjli: These should be made private. If you intended to expose any other functionality through these methods, more specific methods with more more specific interfaces should be made for those functionalities.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -90,15 +97,33 @@ PF::HAL::AnalogInput ADC3Input(hadc3, adcPollTimeout);
...
 PF::HAL::DigitalOutput alarmRegMed(*ALARM1_MED_GPIO_Port, ALARM1_MED_Pin);
 PF::HAL::DigitalOutput alarmRegLow(*ALARM1_LOW_GPIO_Port, ALARM1_LOW_Pin);
 PF::HAL::DigitalOutput alarmBuzzer(*ALARM2_CNTRL_GPIO_Port, ALARM2_CNTRL_Pin);
+PF::HAL::DigitalInput pushButton(*B1_GPIO_Port, B1_Pin);
+PF::HAL::DigitalInput powerInput(*SET_PWR_ON_OFF_GPIO_Port, SET_PWR_ON_OFF_Pin);
+PF::HAL::DigitalInput inputButton(*Mem_Button_GPIO_Port, Mem_Button_Pin);
 
+PF::Driver::Button::Debouncer switchDebounce;
+PF::Driver::Button::EdgeDetector switchTransition;
+PF::Driver::Button::Button buttonMembrane(inputButton,switchDebounce);
```
ethanjli: I'm pretty sure these classes are not provided by this PR - I would not expect `main.cpp` to compile properly as is. If so, this absolutely would need to be fixed before the PR is merged into develop.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -63,6 +68,7 @@ I2C_HandleTypeDef hi2c2;
 I2C_HandleTypeDef hi2c4;
 
 SPI_HandleTypeDef hspi1;
+SPI_HandleTypeDef hspi2;
```
ethanjli: I see that you have added SPI2 in this PR. Is this compatible with the pin assignments of the Nucleo board when it is plugged into the control board?

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -37,6 +37,11 @@
 #include "Pufferfish/Driver/I2C/SFM3000.h"
 #include "Pufferfish/Driver/I2C/TCA9548A.h"
 #include "Pufferfish/Statuses.h"
+#include "Pufferfish/HAL/STM32/Time.h"
+#include "Pufferfish/Driver/Button/Button.h"
+#include "Pufferfish/Driver/SPI/SPIFlash.h"
```
ethanjli: Flash memory functionality is not provided by this PR, so I'm pretty sure the build of `main.cpp` would be broken right now in this PR as-is.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -630,15 +668,15 @@ static void MX_SPI1_Init(void)
...
   hspi1.Init.CLKPolarity = SPI_POLARITY_LOW;
   hspi1.Init.CLKPhase = SPI_PHASE_1EDGE;
   hspi1.Init.NSS = SPI_NSS_SOFT;
-  hspi1.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_2;
+  hspi1.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_256;
   hspi1.Init.FirstBit = SPI_FIRSTBIT_MSB;
   hspi1.Init.TIMode = SPI_TIMODE_DISABLE;
   hspi1.Init.CRCCalculation = SPI_CRCCALCULATION_DISABLE;
-  hspi1.Init.CRCPolynomial = 0x0;
+  hspi1.Init.CRCPolynomial = 0x7;
```
ethanjli: Are these changes necessary for the BMP388 sensor to work?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
RLaxmanrao: Updated naming conventions for TimeStandby enum.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
RLaxmanrao: hspi1.Init.DataSize = SPI_DATASIZE_4BIT; Required to change to 8 bit data.
hspi1.Init.BaudRatePrescaler = SPI_BAUDRATEPRESCALER_2; Baud rate needs to be define? Please confirm which baud rate is expected at system level.
hspi1.Init.CRCPolynomial = 0x0;  change is not required.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
RLaxmanrao: Removed SPIFlash.h inclusion.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
RLaxmanrao: SPI2 is added for testing on standalone Nucleo board. can be removed from main.cpp while merging.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
RLaxmanrao: Removed membrane button related objects from main.cpp

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
RLaxmanrao: Updated by reference

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
ethanjli: If readability is a big concern, where you use unions for type-punning (which is undefined behavior), you might instead use `reinterpret_cast` for type-punning. Refer to https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c183-dont-use-a-union-for-type-punning for more information.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
RLaxmanrao: Updates as private methods.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/SPI/BMP388.h
RLaxmanrao: Removed unions from the code

ethanjli:
I will take over this PR to work on it and start making changes to prepare it to merge into develop (i.e. merging develop into this branch, resolving the Github Actions checks). Can each of you answer the following questions for records-keeping for your contributions to this PR?

hemanthgs:
For my contributions on this PR:

This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? Yes
Were any of these contributions also part of work you did for an employer or a client? No
Does this work include, or is it based on, any third-party work which you did not create? No

ethanjli:
The develop branch has diverged, and merging this in will require taking a fresh look at the code in this PR; since we won't need this sensor for a while, I'm going to close this PR for now. We can reopen it later when needed.

Issue #51, opened 2020-08-20
Add filter for maintaining data integrity during protobuf file I/O. 


Tanmay06:
Test various hash functions and, build a filter to be used in file I/O protocol. This filter will check for bad sectors in the file and help us main the data integrity.

Issue #40, opened 2020-08-18
Async low-level file I/O driver


ethanjli:
We will need an asynchronous low-level file I/O driver (implemented using Trio) which simply takes a file path and byte contents to write to file, and which simply takes a file path to read byte contents from file. We will then build higher-level protocol buffer storage/loading on top of this file I/O driver.

PR #166, opened 2020-09-25
File I/O system.

Tanmay06:
This PR brings writing system states to `protobuf` files at scheduled intervals and the ability to read the states at the startup of the system. It also ensures data integrity in the in the `protobuf` filesystem by leveraging `crcelements` filter.
This PR adds async low-level file driver(`Handler`), high-level file protocol and necessary modifications for integration in the system. It addresses #40 and #51.
Note: It is advised to merge this after PR #165  has been resolved, since this is an old branch it may require to resolve the conflicts while merging.


Tanmay06:
This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? **Yes**
Were any of these contributions also part of work you did for an employer or a client? **No**
Does this work include, or is it based on, any third-party work which you did not create? **No**

ethanjli:
Can you see if the firmware checks get resolved after you merge the develop branch into this branch?

backend/tests/unit/io/trio/test_fileio.py
```diff
@@ -0,0 +1,136 @@
...
+    filename = "fileio_read_new"
+    if os.path.exists(os.path.join(FILES_PATH, filename + '.pb')):
+        os.remove(os.path.join(FILES_PATH, filename + '.pb'))
+    filehandler.set_props(filename=filename, mode='rb')
+    with pt.raises(OSError):
+        await filehandler.open()
+
+    filehandler.set_props(filename=filename, mode='wb')
+    await filehandler.open()
+    await filehandler.send(data)
```
ethanjli: Does this test need to call `filehandler.close()`?

backend/tests/unit/io/trio/test_fileio.py
```diff
@@ -0,0 +1,136 @@
...
+@pt.mark.parametrize("data_list", [examples[:2]])
+async def test_fileio_write_read_ow(
+        data_list: List[Tuple[bytes, str]]) -> None:
+    """"""
+    filehandler = fileio.Handler()
+    filehandler.props.filedir = FILES_PATH
+    filename = "fileio_write_read_ow"
+    for data, _ in data_list:
+        filehandler.set_props(filename=filename, mode='wb')
+        await filehandler.open()
```
ethanjli, line 73: Since this `fileio.Handler` has been designed to require an open+close for a read/write on every separate file, I think we should consider adding `__aenter__` and `__aexit__` methods to `fileio.Handler` so that we can handle the file operations in a context manager style - e.g. something like:
```python
async with filehandler.file(filename=filename, mode='wb') as f:
  await f.send(data)
```
or the following pattern (maybe this is already supported? It looks like I am using it in `process_io_persistently`):
```python
await filehandler.open()
async with filehandler:
  await filehandler.send(data)
```
so that we don't have to worry about leaking file handles. We can address this in a future PR

Note that I am suggesting this different interface here because `fileio.Handler` is being used in a very different way than the other drivers: in the other drivers, we just want to op

backend/tests/unit/io/trio/test_fileio.py
```diff
@@ -0,0 +1,136 @@
...
+    filehandler.set_props(filename=filename, mode='wb')
+    await filehandler.open()
+    await filehandler.send(data)
+    await filehandler.close()
+
+    filehandler.set_props(filename=filename, mode='rb')
+    for _ in range(3): 
+        await filehandler.open()
+        output = await filehandler.receive()
+        assert output == data
```
ethanjli, line 118: If this assertion fails, then we leak the open file descriptor, right? (this relates back to my comment above about using context managers for file descriptors)

backend/tests/unit/protocols/test_file.py
```diff
@@ -0,0 +1,215 @@
...
+
+import pytest as pt  # type: ignore
+import betterproto
+
+from ventserver.protocols import exceptions
+from ventserver.protocols import file
+from ventserver.protocols.protobuf import mcu_pb
+
+
+examples_good = [
```
ethanjli, line 11: For maintainability if/when we modify the definitions of the protobuf messages in the future, I recommend storing the CRCs as separate bytestrings (e.g. instead of `b'U\xa3\x13\xae\x04\x08\xdd\xe1...` it looks like `b'U\xa3\x13\xae' + b'\x04\x08\xdd\xe1...`, or else the CRC is actually a separate element in the tuple)

backend/ventserver/integration/_trio.py
```diff
@@ -82,6 +86,23 @@ async def send_all_websocket(
                 'Illegal data type: %s', err
             )
 
+async def process_file_save_output(
+        protocol: server.Protocol,
+        filehandler: fileio.Handler
+) -> None:
+    """Saves protobuf data to files (prototype)"""
+
+    for message in protocol.send.file.output_all():
```
ethanjli: To ensure correctness of asynchronous operations, we should not be accessing `protocol.send.file`; rather, any outputs from the File protocol should be generated when `protocol.send.output` is called, and those outputs should just be in a `file_send` field in `protocols.server.SendOutputEvent`. The processing of that field should then be done in in a similar way as is done for the websocket endpoint in `integration._trio.process_protocol_send_output`: we make a `file_send` `DeqaueChannel`, then we start a `send_all_file` task in the nursery.

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +334,49 @@ async def process_all(
...
+        ]
+) -> None:
+    """Initialize state values from state store or default values."""
+    default_init = list()
+    for state in states:
+        try: # Handle fileio errors
+            filehandler.set_props(state.__name__, "rb")
+            await filehandler.open()
+            message = await filehandler.read()
+            protocol.receive.file.input(
```
ethanjli: Strictly speaking, we should not access `protocol.receive.file`; rather, any inputs to the File protocol should be passed through `protocol.receive.input`, and those inputs should just be in a `file_receive` field in `protocols.server.ReceiveInputEvent`. This would be symmetric with what I described above for outputs from the File protocol, and also it would be consistent with what we are doing in the rotary encoder.

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +334,49 @@ async def process_all(
...
+                file.StateData(state_type=state.__name__, data=message)
+                )
+        except OSError as err:
+            print(err)
```
ethanjli: We should use `logging.error` or `logging.warning` instead of `print`.

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +334,49 @@ async def process_all(
...
+        finally:
+            await filehandler.close()
+
+    while True:
+        try: # Handles data integrity and protocol error
+            event = protocol.receive.file.output()
+            if not event:
+                break
+
+            all_states[type(event)] = event
```
ethanjli: Would it be feasible to have the server protocol or the backend protocol do the work of filling in the value in `all_states` (e,g, maybe even just using the backend's `file_state_synchronizer` to do that work), instead of doing it in `integration._trio`?

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +334,49 @@ async def process_all(
...
+        except exceptions.ProtocolDataError as err:
+            print(err)
+
+    for state in states:
+        if not all_states[state]:
+            default_init.append(state)
+
+    for state in default_init:
+        if state is mcu_pb.ParametersRequest:
+            all_states[mcu_pb.ParametersRequest] = mcu_pb.ParametersRequest(
```
ethanjli: Would it be possible for us to pass in the default initializations as a parameter to `initialize_states`? Perhaps we could just initialize all those states at the start of the function, and then overwrite them with states from any successfully-loaded files? And then states left out of the list/map/etc. of default initializations remain as `None`, i.e. the backend will wait for a value to be provided by the firmware or the frontend.

backend/ventserver/io/fileobject.py
```diff
@@ -0,0 +1,15 @@
...
+from typing import Text
+
+import attr
+
+@attr.s(auto_attribs=True)
+class FileProps:
+    """ Defines protobuf file details."""
+
+    filename: str = "state.pb"
+    filedir: str = os.path.join(os.getcwd(), "ventserver", "statestore")
```
ethanjli: I am thinking it might be helpful to make the file handler to take a path to a root directory (which defaults to `os.path.join(os.getcwd(), "ventserver", "statestore")` but can be set to something else), and then `FileProps.filedir` specifies the child directory path within the root directory (and maybe it defaults to `""`). This way, we can change where the root directory is more easily (since in production mode it will need to be on a USB drive).

backend/ventserver/io/trio/fileio.py
```diff
@@ -0,0 +1,96 @@
...
+        if (self._fileobject is None) and (not self.is_open):
+            return
+
+        assert self._fileobject is not None
+        await self._fileobject.aclose()
+        self._fileobject = None
+        self._logger.info('File connection closed.')
+        self._connected = trio.Event()
+
+    async def receive(self) -> bytes:
```
ethanjli, line 79: Do we need to have both `receive` and `read`?

backend/ventserver/io/trio/fileio.py
```diff
@@ -0,0 +1,96 @@
...
+
+    async def read(self) -> bytes:
+        """Reads data from protobuf file and returns a bytes object."""
+        if self._fileobject is None:
+            raise RuntimeError("No file object defined to read.")
+        # TODO: Recognize and handle async File I/O exceptions
+        _data = await self._fileobject.read()
+        return _data
+
+    async def send(self, data: Optional[bytes]) -> None:
```
ethanjli, line 87: Do we need to have both `send` and `write`?

backend/ventserver/protocols/file.py
```diff
@@ -0,0 +1,147 @@
...
+from ventserver.protocols import messages
+from ventserver.protocols import crcelements
+from ventserver.sansio import protocols
+from ventserver.sansio import channels
+
+
+# Classes
+
+@attr.s(auto_attribs=True)
+class StateData:
```
ethanjli: I believe this class should inherit from `protocols.events.Event`? Then we can use the `has_data` method to check inputs in the `FileReceiver.input` method.

backend/ventserver/protocols/server.py
```diff
@@ -100,6 +101,9 @@ class ReceiveFilter(protocols.Filter[ReceiveEvent, ReceiveOutputEvent]):
         factory=rotary_encoder.ReceiveFilter
     )
     _backend: backend.ReceiveFilter = attr.ib(factory=backend.ReceiveFilter)
+    _filerw: file_filters.ReceiveFilter = attr.ib(
```
ethanjli: What does `filerw` mean in this context?

ethanjli (commented):
I did a pretty close review of all code - the main thing I would suggest changing is the files-related interface for the server protocol, and the integration code; they should be made more similar to the pattern established with the mcu, frontend, and rotary encoder protocols and endpoints. If those changes are feasible, do you want to make those changes (as well as any other small changes) in this PR, or would you prefer to do them in future PRs after this one gets merged in?

backend/ventserver/integration/_trio.py
Tanmay06: `initialize_states()` was in `application.py` and `simulation.py` but since `pylint` was throwing duplicated code warnings, I had to move it in `_trio.py`. Otherwise, it is being called at the startup in `application.py` and `simulation.py`. 
In this case `all_states` is being initialized at the startup, I'm not sure how `file_state_synchronizer` can be used to fill in `all_states`.  May be I'll take a look at it and see what can be done.

backend/tests/unit/protocols/test_file.py
Tanmay06: If I understand your concern correctly, then if the definitions of the protobuf messages are changed in the future; anyhow we'll have re-compute the CRC header and the payload of the message. Apart from this, I also wanted to keep the `file` protocol completely unaware of the CRC header. The only thing it should deal with is the `StateData`, i.e `state_type` and `file_data`, and `protobuf.Message`; and I wanted to maintain same thought process while writing the tests as well.

Tanmay06:
I believe that when I first wrote this code, I was still new to the project and was trying to understand the code. That's why the integration wasn't as robust as expected. Anyways, I've made all the necessary changes that you've suggested. Let me know if there is anything that can be improved any further.

backend/tests/unit/protocols/test_file.py
ethanjli: Ok, that's fair!

backend/ventserver/protocols/file.py
ethanjli: You  marked this resolved but I don't see any difference in the code here - is there a commit you forgot to push up, or are you thinking `StateData` shouldn't inherit from `protocols.events.Event`?

backend/ventserver/protocols/file.py
```diff
@@ -54,7 +59,7 @@ def init_message_receiver(self) -> messages.MessageReceiver:  # pylint: disable=
     def input(self, event: Optional[LowerEvent]) -> None:
         """Handle input events."""
         if event:
```
ethanjli: I would suggest restructuring this as:
```python
if event is None:
  return

if not event.has_data():
  raise ...
```

backend/ventserver/io/fileobject.py
```diff
@@ -11,5 +11,6 @@ class FileProps:
     """ Defines protobuf file details."""
 
     filename: str = "state.pb"
-    filedir: str = os.path.join(os.getcwd(), "ventserver", "statestore")
+    rootdir: str = os.path.join(os.getcwd(), "ventserver", "statestore")
```
ethanjli: `rootdir` should be an attrib of `io.trio.fileio.Handler` rather than of `FileProps` - this way we can reuse the same `FileProps` even if we want to change `rootdir` (which we change by just giving a different value to the `fileio.Handler` constructor)

backend/ventserver/statestore/Alarms.pb
```diff
@@ -0,0 +1 @@
+��R
```
ethanjli: I just realized that `ventserver/statestore` is being included in our git tracking. Would it be reasonable to the `.pb` files from git tracking and put `ventserver/statestore` under `.gitignore`? It could make sense to have a folder of the default `.pb` files (which would include `ParametersRequest` but exclude `Alarms`, `CycleMeasurements`, and `SensorMeasurements`).

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +336,26 @@ async def process_all(
...
+        filehandler: fileio.Handler
+) -> None:
+    """Initialize state values from state store or default values."""
+    for state in states:
+        try: # Handle fileio errors
+            filehandler.set_props(state.__name__, "rb")
+            await filehandler.open()
+            async with  filehandler:
+                message = await filehandler.receive()
+                protocol.receive.input(
```
ethanjli, line 353: Since the server `ReceiveFilter`'s `_process_file` method gets called from its `output` method, I believe we will need to call the `output` method until it returns `None`.

backend/ventserver/integration/_trio.py
```diff
@@ -312,3 +336,26 @@ async def process_all(
             nursery.start_soon(
                 process_clock, protocol, channel, push_endpoint.clone()
             )
+
+
+async def initialize_states(
```
ethanjli: Now that this is just for initializing states from file, it would help to rename this function to reflect that, e.g. `initialize_states_from_file` or `load_file_states` or something.

ethanjli (commented):
Just a few small follow-up comments to address here, and then this will be good to merge.

ethanjli (approved):

PR #172, opened 2020-09-29
Feature/frontend event alerts

Sudhir-dev:
This PR includes
* Proto message for Event logs
* Event Log Listing -
   Functionality to list event logs from "NextLogEvents" 
   Increment & trigger "ExpectedLogEvent" based on list from NextLogEvents
   Update Status of Event to Active/Inactive based on ActiveLogEvents
* Event log popup - Alarm popup on header based on "ActiveLogEvents" 

This PR does not include
* Removal of "View Details" from event log listing
* Alarm popup design changes(Right now its shows everytime page is navigated)


frontend/src/store/controller/reducers.ts
```diff
@@ -46,12 +51,11 @@ const messageReducer = <T extends PBMessage>(
...
+        if (messageType === MessageType.NextLogEvents) {
+          return mergeEventList(action.state as NextLogEvents, state as NextLogEvents) as T;
+        }
         if (action.messageType === MessageType.RotaryEncoder) {
-          const updatedState = calculateStepDiff(
-            state as RotaryEncoder,
-            action.state as RotaryEncoder,
-          );
-          return updatedState as T;
+          return calculateStepDiff(action.state as RotaryEncoder, state as RotaryEncoder) as T;
```
ethanjli, line 58: Is there any reason you've put these `MessageType` case handlers inside the general `messageReducer`, rather than making their own reducers? It's not really consistent with the spirit of how `messageReducer` is parameterized - but I'm not sure if there is a way to compose two reducers (e.g. feed the output of `messageReducer<NextLogEvents>(MessageType.NextLogEvents, NextLogEvents) into a reducer which calls `mergeEventList`), either. If we want to fix this, we can do it in a future PR - this is okay for now.

frontend/src/store/controller/reducers.ts
```diff
@@ -61,9 +65,26 @@ const messageReducer = <T extends PBMessage>(
...
+  const events = [...oldEvents.logEvents];
+  if (!oldEvents || !oldEvents.logEvents.length) {
+    return newEvents as NextLogEvents;
+  }
+  newEvents.logEvents.forEach((logEvent: LogEvent) => {
+    const id = events.find((ev: LogEvent) => ev.id === logEvent.id);
+    if (id === undefined) {
+      events.push(logEvent);
+    }
+  });
```
ethanjli, line 78: There's definitely a more efficient way to merge two sorted arrays (`newEvents` and `oldEvents` should both have been sorted), but this implementation is fine for this PR.

ethanjli (approved):
I reviewed the Redux code and everything generally looks good to me. I've identified two places where we might wane to refactor things in the future (one is a question about organizing the code, the other is a potential performance optimization if we find we need it)

frontend/src/store/controller/reducers.ts
Sudhir-dev: Has been fixed on #190

frontend/src/store/controller/reducers.ts
Sudhir-dev: Has been fixed on #190

PR #175, opened 2020-10-01
HFNC feedback control of flow

ethanjli:
This PR:

- Ports over Hongquan's proportional-integral feedback controller for flow rate control in HFNC mode, and puts it into a new `Controller` unit.
- Decomposes `ParametersRequest`/`Parameters` servicing out of `Simulator`, into its own unit (which operates in different modes),
- Adds a `ControlLoop` unit for coordinating the activity and timings of sensor reads, `Controller` updates, and actuator writes.
- Applies some refactoring to the SFM3019 driver as discussed with Likhith in #173.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/SFM3019/Sensor.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/SFM3019/Sensor.h
@@ -17,51 +17,26 @@
 namespace Pufferfish::Driver::I2C::SFM3019 {
 
 /**
- * State management for Sensirion SFM3019 flow sensor, without I/O
+ * State management for Sensirion SFM3019 flow sensor, without I/O.
+ * This is basically a Moore machine, so the state consists of the
+ * next action to take, along with the current time.
  */
 class StateMachine {
  public:
-  enum class State {
-    uninitialized,
-    waiting_warmup,
-    checking_range,
-    measuring,
-    waiting_measurement
-  };
-  // Next Actions
-  enum class Output {
-    initialize,
-    wait_warmup,
-    check_range,
-    measure,
-    wait_measurement,
-    error_fail,
-    error_input
-  };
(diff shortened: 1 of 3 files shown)


ethanjli:
@alecaberg Following up from your email, this PR has a rough first pass at how I've split the control code up into separate modules:

- The pure-logic feedback control algorithm (which operates purely on input data and produces output data) is a `Controller` class (in the future, once we have multiple variables each with their own PI/PID/etc. feedback, such as in the more complex modes, we might want to split out parts those implementations for various modes into separate classes for single-input single-output control algorithms, e.g. one for PID). One type of `Controller` is for the HFNC mode, and we'll have other types for future modes. HFNC might be simple enough that it doesn't need states, but the other modse certainly will need their own states.
- Currently the output variables of `Controller` are provided through a non-`const` reference in the constructor, but I'm considering not passing it through the constructor - maybe through the `update` method instead. We are following the latter pattern for other parts of the code, except I name the method `transform` instead of `update` in those other cases.
- The code which integrates I/O reads and writes, timing, and the pure-logic algorithms is a `ControlLoop` class (I was having trouble coming up with descriptive+specific names). One type of `ControlLoop` is for the HFNC sensors+actuators+algorithm, and we'll have other types for future modes.
- We haven't yet started to prototype how we'll do sensor cross-checking using redundant sensors to detect faults, this needs some discussion.

ethanjli:
@pjstrohm Here's the pull request I mentioned in our meeting today, where I've split up the driver code for an I2C gas flow sensor (the [SFM3019](https://www.sensirion.com/fileadmin/user_upload/customers/sensirion/Dokumente/5_Mass_Flow_Meters/Datasheets/Sensirion_Mass_Flow_Meters_SFM3019_Datasheet.pdf)) into a low-level stateless wrapper around the blocking I2C read and write operations provided by the STM32 HAL (`Pufferfish::Drivers::I2C::SFM3019::Device`), and a higher-level driver (`Pufferfish::Drivers::I2C::SFM3019::Sensor`) which uses a state machine (`Pufferfish::Drivers::I2C::SFM3019::StateMachine`)  to keep track of the state of the sensor.

PR #179, opened 2020-10-02
save logs to usb

Praneet9:
This PR includes redirecting `backend`, `kiosk` and `nginx` logs to USB.
The USB is mounted by name so the name of USB being mounted is mandatory.
I've used **LOGS** as the USB Label.
It also does log rotation every hour
The configuration for log rotation is just random as the requirements aren't clear yet. We can change it accordingly once the requirements wrt USBs are clear.


ethanjli:
This works well for me! A few observations:

- If I run `./kiosk.sh` to set up the Chromium kiosk but I have never run Chromium before, then the service errors out with `/bin/sed: can't read /home/pi/.config/chromium/Default/Preferences:  No such file or directory`. This problem is resolved after I launch Chromium for the first time. Does this mean that our README needs to add instructions to manually run Chromium once during the setup process?
- When I press Ctrl+W to quit the Chromium kiosk-mode browser (i.e. Chromium exits normally based on user input), after a short delay systemd relaunches it. The only way I can prevent it from relaunching is to run `sudo systemctl stop kiosk.service`. This behavior is slightly annoying during development, but it seems desirable in production.

Can you answer the following questions for records-keeping:

Praneet9:
1. We don't need to open chromium for the first time. That doesn't sound right. I'll add a check to see if the file exists before replacing text which should handle that case.
2. I've used `Restart=always` for it to restart everytime it quits keeping in mind the production environment. It can be changed to `Restart=on-failure` if required. As we won't be using it as a service in development environment I didn't add that.

For records-keeping:

Praneet9:
I tested it with a fresh installation and the bug has been fixed.
I have added a script to `ExecStartPre` which resets the flags in config before starting chromium.
This script checks if the file exists and resets only if the config exists which solves the problem of starting chromium for the first time.

ethanjli (approved):

PR #184, opened 2020-10-06
Bug/backend rotary encoder fix

### LINKED ISSUE #176: Improve Rotary Encoder Driver
The rotary encoder sometimes has glitchy behavior. Most probably because of the encoder callback algorithm and debounce. This issue address this problem in the low-level driver of the rotary encoder.

Tanmay06:
This PR addresses the issue with rotary encoder. It uses `pigpio-encoder` based algorithm for reading rotary encoder inputs and accurately encoder ticks/steps. This PR closes #176 .


Tanmay06:
Since `pigpio` library requires `pigpio` daemon running in the background, I would suggest @Praneet9 to add `sudo pigpiod` in the bash scripts to start the daemon before the backend.

Tanmay06:
For records-keeping:
3. Does this work include, or is it based on, any third-party work which you did not create? **Yes**

Tanmay06:
Source of reference: https://github.com/vash3d/pigpio_encoder/blob/master/src/pigpio_encoder.py
License of the source:  https://github.com/vash3d/pigpio_encoder/blob/master/LICENCE.txt

Tanmay06:
@Praneet9 you'll also need to build `pigpiod` binaries during the installation process. 
```
wget https://github.com/joan2937/pigpio/archive/master.zip
unzip master.zip
cd pigpio-master
make
sudo make install
```
You can find the details [here](http://abyz.me.uk/rpi/pigpio/download.html).

Tanmay06:
In some fresh install cases, I'm facing a bug where the button is working but the knob is not. @ethanjli can you check whether you are facing the same issue ? Meanwhile, I'm looking into the cause of the problem.

ethanjli:
A few comments:

1. I only just realized that `pigpio_encoder` is GPL-licensed - so we aren't supposed to use code from it in our project, since our project is Apache-licensed (i.e. too permissive). If you only used `pigpio_encoder`'s algorithm as inspiration and the same algorithm is described somewhere else on the internet, I think that's okay - but in that case we would be clear that we have not used code from `pigpio_encoder` in our project. Instead, our project just brings in the Unlicense-licensed `pigpio` library as a dependency.
2. I tested this with a hardware-debounced rotary encoder, and the backend software driver works well: I don't get glitches from bouncing, and I don't get missed clicks (as far as I can tell).
3. I think it would be clearer if we followed the `A`/`B` pin naming convention everywhere, rather than `CLK`/`DT`, since our encoder also uses `A`/`B` pin names.
4. When the button is pressed, I am seeing a race condition between when the backend resets `steps` to 0 and when the frontend tries to commit the value based on `steps` into `ParametersRequest`. To avoid this race condition, we should have the frontend commit the value into `ParametersRequest` when the button is pressed down, and we should have the backend reset `steps` to 0 when the button is released - this way the value is always committed before `steps` is changed by the backend. Currently, the frontend is committing the value when the button is pressed down, but the backend is resetting `steps` to zero when the button is pressed down. I've confirmed in testing that I can fix this by moving the `self.counter = 0` line from the `button_fall` callback to the `button_rise` callback in the rotary encoder I/O endpoint. I think if you make this change in the PR, then the rotary encoder should behave correctly consistently.

Tanmay06:
Sounds good. 

1. I've just used the algorithm from `pigpio-encoder`, so I guess we just have to look for another source of that algorithm and add `pigpio` as our only dependency and then we are good to go. 
2. When I tested it without the hardware debouced driver it was working fine, apart from very rare cases of missed clicks. I guess that hardware debouncing will solve that issue.
3. Yes, I was about to rename it before giving it a final go. I had kept it just to keep things clear and avoid any confusion on my end, since in the reference they were using `clk/dt` convention.
4. In the previous driver with `RPi.GPIO`, I was resetting the the `counter` when the button is released. This time around I wanted to see what difference does it make when I tried to reset at  button press. I can see now what is causing the issue at the frontend. I will switch it back to the `button_rise`.
5. I've also discovered that `pigpiod` is available by default in RaspberryPi OS.

Tanmay06:
[Here](https://www.brainy-bits.com/arduino-rotary-encoder-ky-040/) I found a Arduino sketch implementation of a similar algorithm that records the input pattern to determine the direction of rotation. A detailed explanation of the algorithm could be found [here](https://www.best-microcontroller-projects.com/rotary-encoder.html). From this two references it seems like the algorithm is pretty generic and it won't cause much trouble if we use it in our project. However, I'll also ensure that no part of the code is derived from `pigpio-encoder`. 

If this is done then do I need to re-answer the records-keeping questions, especially the third one ? 

Apart from this, we can also use the information from [second link](https://www.best-microcontroller-projects.com/rotary-encoder.html) to write automated test for the rotary encoder driver.

ethanjli (approved):
I think this is good to merge, it works well in testing with our new debouncing circuit (which is just an RC filter with a capacitor of 100 nF and a resistor of 10 kOhm on A and on B), and the comment log in this PR is sufficient to address+record the questions about use of third-party works.

PR #186, opened 2020-10-06
Fix pylint check issues introduced in #165

ethanjli:
This PR fixes some pylint complaints about rotary encoder code. Not sure how they got missed before #165 was merged in.


Tanmay06:
Interesting, I'm curious why it was not caught in build checks after I pushed it.

PR #188, opened 2020-10-07
pigpio setup script

### LINKED ISSUE #187: `pgpiod` build and service autostart.
The new rotary encoder driver is built upon `pigpio`, which requires `pigpiod` daemon running in background. The deployment process needs to build the daemon during installation process and start the daemon before the backend. This issue is related to PR #184, and issue #176.

Praneet9:
This PR adds a script to build and make pigpio and adds it to `ExecStartPre` to backend service so that it is started before the backend starts.
It is added to setup.sh so that it is built in development environment as well.

### DIFF
--- a/deploy/configs/pufferfish_backend.service
+++ b/deploy/configs/pufferfish_backend.service
@@ -7,6 +7,7 @@ Before=lightdm.service
 [Service]
 Type=simple
 User=pi
+ExecStartPre=sudo pigpiod
 ExecStart=/home/pi/.pyenv/versions/ventserver/bin/python python_file_absolute_path
 Restart=always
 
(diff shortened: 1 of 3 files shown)


Tanmay06 (commented):
Looks good. I've tested it and it works as expected.

Tanmay06 (approved):

Praneet9:
For records keeping:

ethanjli (approved):
Looks good to me too. I only ran an earlier commit which tried to build pigpio from source, but I'm confident this will work fine now that it's been replaced with `sudo apt install pigpio`. You might want to replace the comment in line 9 of `deploy/setup.sh` from `# Build pigpio` to `# Install pigpio`.

PR #198, opened 2020-10-09
Frontend routing fixes & battery power value

### LINKED ISSUE #33: Power state indicator
We will need to implement power indication functionality in the frontend for displaying the status of the power supply: which power source is being used, and what is the charge level of the battery. The latter has been at least partially designed by the UI/UX team, while the former might not have been designed yet.

This will require defining a new Protocol Buffer message in `proto/mcu_pb.proto` for power. For now, this message just needs to include a PowerSource enum (currently with two values: Internal, External) and a uint8 for the battery charge level (0% to 100%). Please add comments identifying any additional design specifications which need to be provided by the UI/UX team's mockup.

Sudhir-dev:
This PR consists of

* Proto message for battery power & ventilation on/off
* Updated Screensaver routing to dashboard/quick start based on ventilation on/off
* Updated battery power value to load from redux
* Removed Quickstart navigation from sidebar


Sudhir-dev:

ethanjli:
Generally looks good, a few questions/comments:

1. Is activation of the screensaver after some idle time included in this PR, or will that be a separate item?
2. When I go to a screen from a button the sidebar and then go to the standby/dashboard screen from the Pufferfish button, the button for the previous screen is still highlighted as the active page.
3. It looks like in the new Adobe XD prototype the quickstart button will be added back into the sidebar.

We can merge this PR in as-is, if you think we'll need to make deeper changes anyways to any aspect of the sidebar in the future.

ethanjli (approved):

Sudhir-dev:
Sure, for sidebar we might anyway need to make changes. And for the screensaver, I will raise another PR for that

PR #200, opened 2020-10-10
TypeError checks for libraries that don't support type annotations. 

### LINKED ISSUE #125: Add type checks when using packages without type annotations
We will need to go through backend code and look for any packages not providing type annotations (indicated by a `# type: ignore` comment) and make sure we are doing appropriate `TypeError` handling, since mypy cannot catch type errors with that

Tanmay06:
This PR is follow up to #125. This PR includes type checks for modules using `trio_websocket` and `cobs`.  Type checks for modules using `serial` is not added yet.


Tanmay06:
For records-keeping:

ethanjli (approved):

PR #201, opened 2020-10-10
Prototype frontend kill.

### LINKED ISSUE #65: Web browser process supervision
We will need some way to start the web browser at startup of the RPi; it will need to start only *after* the HTTP server (which will serve the frontend website to the web browser) has started; and it will also need to automatically restart the web browser if it crashes (and it will need to remove any temporary files after a crash). Importantly, the web browser should start with regular user privileges, not with superuser/root privileges.

It is also possible that we will need to restart the entire web browser if the frontend's browser tab crashes or freezes. In this case, we might need the backend server to be able to (directly or indirectly) kill any web browser processes on the operating system if the backend server detects that it has lost its WebSocket connection to the frontend for more than ~1 second. And of course the web browser will need to be automatically restarted after it is killed.

Tanmay06:
This PR handles the situation where the frontend freezes and becomes irresponsive. If the backend detects that no message has been received from the frontend for more than 1s, it spawns a sub-process to kill all frozen frontend process. It also ensures that no kill process is spawned within 2s of the last one. This PR will resolve #65


backend/ventserver/protocols/server.py
```diff
@@ -137,9 +146,30 @@ def output(self) -> Optional[ReceiveOutputEvent]:
         if not any_updated:
             return None
 
+        # Kill frontend process if it stops responding.
+        # The frontend service will automatically restart the frontend process.
+        if (int(self.current_time - self.last_frontend_event) > 1) and\
+            (int(self.current_time - self.frontend_connection_time) > 2) and\
+            self.frontend_connected:
+            if self._kill_process is None or\
```
Tanmay06: Need suggestions on better way to do this.

backend/ventserver/integration/_trio.py
```diff
@@ -231,6 +232,11 @@ async def process_io_persistently(
     async with push_endpoint:
         while True:
             await io_endpoint.persistently_open(nursery=nursery)
+            if isinstance(io_endpoint, websocket_io.Driver):
+                protocol.receive.frontend_connected =\
```
ethanjli: For asynchronous execution correctness, we should only pass data to/from the protocol through `protocol.receive.input`/`protocol.receive.output`/`protocol.send.input`/`protocol.send.output`.

backend/ventserver/protocols/server.py
```diff
@@ -1,6 +1,8 @@
 """Sans-I/O ventilator backend server protocol."""
 
 from typing import Optional, Union, Tuple
+import subprocess
```
ethanjli: The `subprocess` module should not be used by any protocol. In fact, it is almost like an I/O operation. I wonder if it would be reasonable to make an I/O endpoint which sole responsibility is to kill a process (or just run an arbitrary shell command) when a method is called? Then the responsibility of the `Server` or `Backend` protocol (probably the receive filter) is to only keep track of the time since the last received websocket message and activate a flag in `ReceiveOutputEvent` once the websocket has been unresponsive for too long. And there would be some code in `integration._trio` which makes the I/O endpoint run the command to kill the process when the flag is activated.

backend/ventserver/protocols/server.py
ethanjli: Would it be enough to only check for `int(self.current_time - self.last_frontend_event) > 1)`? So the server protocol only cares about whether it's been too long since it received a websocket message, and the integration code and/or I/O endpoint for running a shell command only cares about running the command when the server tells it to (and the integration code also cares about ignoring any timeout signals which occur too soon after the last timeout signal).

This way we can also put a bit more of the behavior into unit tests, specifically the behavior for the server protocol (and, if we write a mock for the I/O endpoint, we could also test the behavior of the integration code).

backend/ventserver/protocols/server.py
ethanjli: Also, I believe Trio has its own functions for spawning subprocesses in an asynchronous manner: https://trio.readthedocs.io/en/stable/reference-io.html#spawning-subprocesses

ethanjli (commented):
Here's an initial (preliminary) review, which hopefully addresses your question from #65.

backend/ventserver/protocols/server.py
Tanmay06: No it's not enough. In the situations where the system has just started or the frontend has restarted, the `last_frontend_event` has it's default value or the old value. So, as soon as the frontend connects, the `current_time` is compared with, not so updated, `last_frontend_event`. And since the difference is usually more than a sec, the kill process is spawned. Furthermore, the `last_front_event` is then never updated; the system is stuck in a situation where whenever the frontend connects the backend kills it.

backend/ventserver/protocols/server.py
Tanmay06: Even I was hesitant about adding a `subprocess` in protocol.  No, I don't think that it would be reasonable to make an I/O endpoint just for spawning a kill process, since we don't need to interact with the process. I was thinking of something similar to `process_clock`, an async process whose sole job is to check frontend and spawn frontend kill process.

ethanjli: That sounds reasonable. Let's put that function in the `io` subpackage rather than the `integration` subpackage.

ethanjli:
I tried testing this by running `poetry run python3 -m ventserver.simulation` and opening a Chromium browser with the frontend, confirming that the frontend and backend are talking to each other by seeing the waveform plots update in real-time, and then opening a new Chromium tab and closing the one with the frontend; Chromium was not killed. I then repeated the test but used `sudo ~/.pyenv/versions/ventserver/bin/python -m ventserver.simulation` to start the backend, and again the Chromium browser didn't get killed. However, manually running `killall /usr/lib/chromium-browser/chromium-browser-v7` was able to kill the Chromium browser.

What test did you run to confirm that this PR was working? I'll try to reproduce that so that I can see what the difference is.

Tanmay06:
This PR is supposed to kill the frontend if the connected frontend freezes and stops responding. To recreate this, as you had suggested before, you can open the frontend and then use chromium dev-console by pressing `ctrl+shift+j` and then sources->pause. This will imitate a situation where the frontend is not responding, or else you can just open a lot of chromium tabs.
Note: The dev-console will only work with browser mode and not the kiosk mode.

ethanjli:
Got it, the way I previously tested it (i.e. closing the tab with the frontend) was actually closing the websocket connection, so I imagine that does not trigger the browser kill. I've tested it with pausing the frontend through the developer console (indeed, sometimes jut pressing Ctrl+Shift+J freezes the frontend long enough that that alone triggers a browser kill), and I think this PR is good to merge. For records-keeping:

Tanmay06:
For records-keeping:

ethanjli (approved):

PR #202, opened 2020-10-12
Fixed README for backend.

Tanmay06:
Moved auto-documentation details from `backend\README` to `backend\docs\README`. Removed `GPIO.*` , triggering E1101, from `.pylintrc`.


backend/ventserver/simulation.py
```diff
@@ -15,7 +15,7 @@
 try:
     from ventserver.io.trio import rotaryencoder
 except RuntimeError:
-    logging.getLogger().warning('Running without RPi.GPIO!')
+    logging.getLogger().warning('Running without pigpio!')
```
ethanjli: Now that we're using pigpio instead of RPi.GPIO, is this exception handling on import still necessary, or can we run the script without it?

backend/README.md
```diff
@@ -97,22 +97,4 @@ Afterwards, you can quickly re-run tests without running the full tox process:
 
 ```bash
 (ventserver) $ poetry run coverage run --parallel -m pytest --basetemp=.tox/py37/tmp --verbose tests
-```
-
-## Documentation
```
ethanjli, line 102: If the documentation for auto-docs-generation has now been moved into `backend/docs/README.md`, can you leave a section in `backend/README.md` which points people to `backend/docs/README.md` for instructions about auto-docs-generation?

backend/ventserver/simulation.py
Tanmay06: Yeah, we don't need that anymore. I'm removing it for now, but we'll have to test it first so that it won't cause chaos again.

ethanjli (approved):
Looks good.

PR #204, opened 2020-10-12
Test/backend test guidelines

### LINKED ISSUE #49: Write guidelines for writing automated backend tests
We will need to write guidelines & best practices for how to clearly write and document automated tests on the backend, using pytest (and maybe also using hypothesis, I've been using that too). These guidelines could go into Notion as a first step, though it could also make sense to eventually have them be compiled into our Sphinx documentation package. Anyways, we'll eventually need to incorporate them into our regulatory documentation too, so we can just put them on Notion for now.

Tanmay06:
Documented guidelines for writing unit tests. These should be modified and added to the notion documentation. Furthermore, also added sample tests demonstrating how to use BDD type documentation with our current unit tests infrastructure. This PR addresses #49


backend/tests/bdd_samples/guidelines.md
```diff
@@ -0,0 +1,61 @@
+This document describes guidelines to write unit tests for the ventserver modules. We are loosely adhering to BDD (Behaviour Driven Development) style development process, keeping the unit tests writing and description guidelines while leaving the tests first approach. This will help us ensure that we write tests for all the crucial components of our system while not hampering the speed and flow of the development. For further reading [here](https://dannorth.net/introducing-bdd/).
+
+There two reasons for using BDD:
```
ethanjli: It looks like Github messes up the formatting of numbered lists, I believe if you add a blank line before the start of each numbered list it will fix the formatting.

backend/tests/bdd_samples/guidelines.md
```diff
@@ -0,0 +1,61 @@
...
+
+@hp.given(...)
+@pt.parametrize(...)
+def test_this_functioanality(args):
+    """Scenario Sample unit test to show how to integrate tags with docstrings.
+       Given The understanding of the functionality.
+       And The knowledge of BDD.
+       When The unit tests are written.
+       And Conditions are met.
+       Then The test result should show pass.
```
ethanjli: Do you think it'd be feasible to also require adding inline comments (e.g. `# Given a CRCElement object`) in front of the corresponding lines of the actual test code (e,g, `crc_element = crcelements.CRCElement()`)? Or is there some test code which just won't work with that?

ethanjli (commented):
Here's some initial feedback about the tests and test guidelines

backend/tests/bdd_samples/guidelines.md
Tanmay06: Yes, I'll look into formatting stuff. Anyhow, we'll be moving it to notion so I don't think we have to worry much about it. Apart from this, content wise, what are your thoughts?

backend/tests/bdd_samples/guidelines.md
Tanmay06: That's a great point. I think in most cases we can do that, this will help the viewer to easily understand and relate the code with BDD steps. However, there might be some scenarios where it won't work, but there we can put it as a comment for some code block instead of the line.

backend/tests/bdd_samples/guidelines.md
ethanjli: The content looks great to me!

backend/tests/bdd_samples/test_crcelements_bdd_style.py
```diff
@@ -0,0 +1,129 @@
...
+    And: The CRC generated/parsed is in bytes.
+    When: The CRC is between \\x00\\x00\\x00\\x00 and \\xff\\xff\\xff\\xff'
+    And: It's length is 4.
+    Then: ValueError should not be raised.
+    """
+    if (b'\x00\x00\x00\x00' <= crc <= b'\xff\xff\xff\xff') and (len(crc) == 4):
+        crcelements.CRCElement(crc=crc)
+    else:
+        with pt.raises(ValueError):
+            crcelements.CRCElement(crc=crc)
```
ethanjli: Maybe this should actually be split up into two separate tests: one where the `if` case runs (i.e. Given a crc constructor value which is outside the allowed range for 32-bit CRCs, when a CRCElement is constructed using it, then ValueError should not be raised), and the other where the `else` case runs (i.e. Given a crc constructor value which is outside the allowed range for 32-bit CRCs, when a CRCElement is constructed using it, then ValueError should be raised)

backend/tests/bdd_samples/test_crcelements_bdd_style.py
```diff
@@ -0,0 +1,129 @@
...
+    # Given: We have CRCElement object.
+    crc_element = crcelements.CRCElement()
+
+    # And: The parsed CRC and payload, in bytes.
+    crc_element.crc = crc
+    crc_element.payload = payload
+    
+    try:
+        # When: The parsed CRC matches the computed CRC
+        crc_element.check_integrity()
```
ethanjli: I think lines 72 - 78 make more sense as just "When: the CRCElement object checks the integrity of a payload and expected CRC which are consistent with each other"

backend/tests/bdd_samples/test_crcelements_bdd_style.py
```diff
@@ -0,0 +1,129 @@
...
+    And: A CRC element.
+    When: The protobuf message has correct size
+    Then: We should get seperated CRC key and message body.
+    """
+    # Given: A protobuf message in bytes.
+    # And: A CRC element.
+    crc_element = crcelements.CRCElement()
+    
+    # When: The protobuf message has correct size
+    crc_element.parse(body)
```
ethanjli: I think line 98 is just "Given: a CRC element", and line 101 is "When: the CRCElement parses a body of a valid size"

backend/tests/bdd_samples/test_crcelements_bdd_style.py
```diff
@@ -0,0 +1,129 @@
...
+    
+    # Then: We should get seperated CRC key and message body.
```
ethanjli: This might be more descriptive as "Then: the CRC checksum is parsed correctly from the header" for line 104 and "And: the payload is parsed correctly" for line 105

backend/tests/bdd_samples/test_crcelements_bdd_style.py
```diff
@@ -0,0 +1,129 @@
...
+    Scenario: The system reads corrupted file or
+        processes corrupted message.
+    Given: A protobuf message in bytes.
+    And: A CRC message receiver.
+    When: The message bytes are corrupted
+    Then: Raise ProtocolDataError for that message.
+    """
+    # Given: A protobuf message in bytes.
+    # And: A CRC message receiver.
+    receiver = crcelements.CRCReceiver()
```
ethanjli, line 124: I think line 122 makes more sense as just "Given: a CRC message receiver", and line 125 makes more sense as just "When: the receiver is given a body with corrupted bytes", and line 128-129 might be more specific as "Then: processing the body raises a ProtocolDataError"

ethanjli (commented):
I've given some specific feedback on how we might restructure the GIVEN-WHEN-THENs so that they fit well as in-line comments. @chittti it'd be useful to get your input on these example tests, too

backend/tests/bdd_samples/test_crcelements_bdd_style.py
chittti: Yes, I would use your 'When' instead of what's present. At the very least, I would suggest removing the 'And:' on line 65. It is redundant since the CRC and payload that we're receiving is in the form of a CRCElement object (so 'Given' and 'And' are referring to the same thing). 

So, just the simple removal of line 65, and no other change, would be ok

backend/tests/bdd_samples/test_crcelements_bdd_style.py
chittti: Agreed

backend/tests/bdd_samples/test_crcelements_bdd_style.py
chittti: Agreed

backend/tests/bdd_samples/test_crcelements_bdd_style.py
chittti: Agreed

Tanmay06:
I've made the suggested changes and I've also updated the guidelines accordingly. If the guidelines look good we can format it and then move it to the notion page before merging this PR.

ethanjli:
I've copied the contents of guidelines.md into https://www.notion.so/pufferfish/Testing-13ce58d72a154b6883fbd57b71d37f76 , so we can remove it from this PR now. For records-keeping:

Tanmay06:
For records-keeping:

ethanjli (approved):

PR #205, opened 2020-10-13
Implement driver for FDO2 sensor

ethanjli:
This PR:

- Adds the stateless and stateful driver code for the FDO2 sensor which will be used to measure FiO2.
- Integrates the FDO2 sensor (which produces pO2 measurements) with the simulator (which simulates FiO2 measurements).
- Implements support for a subset of the commands made available by the FDO2 sensor.
- Performs a version check and starts the sensor in broadcast mode as the setup routine, but does it in a blocking manner (which may take up to 100 ms).
- Does not support CRC checks
- Does not do error-checking in some places.
- Really needs refactoring.


ethanjli:
This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed for me to make progress on the milestone 8 (v0.5), is of at least comparable (probably slightly lower) quality to the rest of the firmware codebase, and probably would not be able to find someone to review it very quickly. We will need to go back and review everything in this PR and refactor it. I will merge in this PR now.

For records-keeping:

PR #209, opened 2020-10-13
running pigpio as system service

### LINKED ISSUE #208: Run Pigpio as a service
The current execution of pigpio is in `ExecStartPre` of backend service. This uses the command `sudo pigpiod` to run the daemon.
In Production environment as we are disabling sudo priveleges of pi this may not work. A better way would be to run it as a service which is a dependency in `After` and `Requires` section of backend service.

Praneet9:
This PR fixes the issue of pigpio daemon running with the sudo command which is disabled for pi user in the Production environment. It also adds pigpio as a dependency for backend service to start.

### DIFF
--- a/deploy/configs/pufferfish_backend.service
+++ b/deploy/configs/pufferfish_backend.service
@@ -1,12 +1,11 @@
 [Unit]
 Description=Pufferfish Backend Service
-Requires=nginx.service media-pi-LOGS.mount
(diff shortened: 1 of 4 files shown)


ethanjli:
For records-keeping:

Praneet9:

Tanmay06 (approved):
Looks good and works as expected.

PR #211, opened 2020-10-14
Restrict scroll, Screen locking & bug fixes

### LINKED ISSUE #97: Screen locking
Currently the ventilator has a hardware button which toggles whether the screen is locked or unlocked. Locking the screen will allow the operator to lock the screen and then wipe it down with a wet cloth to clean it, without generating spurious touchscreen interactions. The state of this button will be reported from the backend to the frontend as a variable somewhere in Redux state. We will need the frontend to ignore all touch inputs when the screen is supposed to be locked.

Additionally, we will need to show a notification that the screen is locked if the operator tries to tap the screen while the screen is locked. The UI/UX team has not yet designed this so we don't need to implement this yet, but I want to bring this onto our radar.

### LINKED ISSUE #197: Ventilation mode text overflow causes scrollbars
On pages where there is the large low-contrast text at the bottom edge of the screen indicating ventilation mode (e.g. in the Quick Start screen), that text overflows and causes a scrollbar to show up on the screen. This should not happen; instead, the text should simply be cut off at the bottom edge of the screen, and it should not be possible to scroll the entire screen.

Sudhir-dev:
This PR includes fixes for

* Restrict Scroll on all pages & Fit content accordingly
* Sidebar navigation selection issue (Focus of sidebar menu based on page url)
* Modes page, mode selection hover color issue
* Added Overlay to restrict user input (Screen locking) & updated protobuf


Sudhir-dev:

ethanjli (approved):
I haven't been able to test the screen locking, but in general it looks good!

PR #203, opened 2020-10-12
Screen inactive timeout

Sudhir-dev:
* Screen inactive timeout - On timeout, redirects to screensaver page

PR #212, opened 2020-10-15
Fix new issue with typescript versioning

ethanjli:
A new issue appears to have popped up when I run `yarn install` as of today, where some kind of typescript syntax error appears when running `yarn start` - it appears to be the same issue as was reported in https://github.com/facebook/create-react-app/issues/8714 . As reported in that issue, it is resolved by upgrading Typescript to 3.8 or above, so I've done that version bump here. @Sudhir-dev Can you check about whether this TypeScript upgrade appears to have broken anything?

This also caused some additional complaints about possibly undefined items `xScale`, `x`, and, `y`, in `Axes.tsx` and `Loop.tsx` - I did a quick-and-dirty workaround by asserting (using the `!` operator) that they are never undefined, because I'm not sure yet what's the right way to actually address the issue

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -39,7 +39,7 @@
     "redux": "^4.0.5",
     "redux-saga": "^1.1.3",
     "reselect": "^4.0.0",
-    "typescript": "~3.7.2"
+    "typescript": "^3.9.7"
   },
   "scripts": {
     "start": "react-app-rewired start",
--- a/frontend/src/modules/dashboard/components/Loop.tsx
+++ b/frontend/src/modules/dashboard/components/Loop.tsx
@@ -73,13 +73,15 @@ export const Loop = ({
     return yScale;
   }
 
+  // TODO: we need to remove the exclamation points, which are claiming that the
+  // values are never undefined
   return (
     <svg width={width} height={height}>
       <Group>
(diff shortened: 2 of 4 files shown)


Sudhir-dev:
@ethanjli , Same typescript version issues was already solved in #203 , I will merge that PR. Can you verify if its solves the issue. If it does, we could discard this PR

ethanjli:
@Sudhir-dev your fix seems to have solved the issue for me. @Hiwot985 can you check whether you are able to run the frontend successfully after following the following steps:

1. Checkout the `develop` branch
2. Delete the `node_modules` folder
3. Run `yarn install` again
4. Try to run `yarn start` again

Sudhir-dev:
> @Sudhir-dev your fix seems to have solved the issue for me. @Hiwot985 can you check whether you are able to run the frontend successfully after following the following steps:
> 
> 1. Checkout the `develop` branch
> 2. Delete the `node_modules` folder
> 3. Run `yarn install` again
> 4. Try to run `yarn start` again

@ethanjli I have checked this on my system. Typescript issue still persists. This PR will need to be merged for package.json & yarn.lock is also updated here

Sudhir-dev (approved):

PR #214, opened 2020-10-16
Fix setup indicator LED to account for blocking I/O in setup functions

ethanjli:
This PR makes the indicator LED behave more correctly in the firmware during setup. Previously, the indicator LED was using software PWM (assuming non-blocking I/O in the setup functions), while the setup functions themselves were using blocking I/O and blocking delays. Thus, the indicator LED wouldn't actually flash or blink during failure or setup. Now, setup phase has a solid on indicator LED, while failure blocks for flashing for ~2 seconds. Also, now the "setup succeeded" indicator is a slow blink for 2 seconds rather than a solid on.


ethanjli:
This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed for to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, and probably would not be able to find someone to review it very quickly. We will probably go back and refactor the affected sections of code anyways. I will merge in this PR now.

For records-keeping:

PR #215, opened 2020-10-16
detect data tampering

### LINKED ISSUE #216: Detect tampering of OS configs or software sources
Disable backend and frontend services if the tampering check fails.

Praneet9:
Checks for Data Tampering before starting backend and frontend services.
Creates a popup if the tampering check fails and disables backend and frontend services.


Praneet9:

ethanjli:
Can you split up the security script into several scripts which can be run separately?

ethanjli:
I tried to run the new `deploy` script (but with the `security_network` and `security_user` scripts commented out) on an SD card image where I had already previously run the `deploy` script from the develop branch and got an error in `security_tampering`: `ERROR: The pufferfish.target file doesn't exist`. When I do `ls /etc/systemd/system/*.target`, I do indeed see the `pufferfish.target` file there, so something else is going on. Maybe there's an issue with the fact that I had previously already run the `deploy` script including `backend_service.sh`, so I will try again on a fresh SD card image.

Praneet9:
> The pufferfish.target file doesn't exist

This error pops up when there's no such file in the configs directory.

ethanjli:
I just did the clean install and got the same issue:

1. I imaged my SD card with pufferfish.img
2. I ran `git clone https://github.com/pez-globo/pufferfish-software`
3. I `cd`'ed into `pufferfish-software/deploy`
4. I ran `git checkout feature/deploy-detect-data-tampering`
5. I commented out the lines to run `scripts/security_network.sh` and `scripts/security_user.sh`
6. I ran `./deploy.sh`. In the "Setting up Tampering Security" section, I got the same "ERROR: The pufferfish.target file doesn't exist" message. However, I can confirm that the `tampering.service` file does exist inside `deploy/configs`.

If I uncomment the line to run `scripts/security_network.sh` from `deploy.sh` and re-run `deploy.sh` (without re-flashing the SD card), I still get the same error.

From inspecting `scripts/security_network.sh` and `scripts/security_tampering.sh`, I notice that `scripts/security_network.sh` has two lines which define the values of the `script_dir` and `config_dir` variables, while `scripts/security_tampering.sh` does not. If I add those two lines into `scripts/security_tampering.sh`, then I can successfully run `scripts/security_tampering.sh`, and the hash check does seem to work fine on my system. Can you add a fix for `scripts/security_tampering.sh` in this PR?

ethanjli:
I notice that the `deploy/configs/compare_hash.sh` script merely exits once the user presses the button to acknowledge the zenity error message, instead of initiating system shutdown. Would it be reasonably easy to initiate system shutdown, or would there be systemd issues with that?

ethanjli:
Also, would it be reasonably easy to move the pufferfish-related files in `/opt` into `/opt/pufferfish`, just to keep them in a pufferfish-specific subfolder?

Praneet9:
> I notice that the `deploy/configs/compare_hash.sh` script merely exits once the user presses the button to acknowledge the zenity error message, instead of initiating system shutdown. Would it be reasonably easy to initiate system shutdown, or would there be systemd issues with that?

Initiating shutdown is possible but will need some configuration changes. We don't have sudo priveleges so we need to grant poweroff permissions to the pi user. Is that a good thing to do?

Praneet9:
> Also, would it be reasonably easy to move the pufferfish-related files in `/opt` into `/opt/pufferfish`, just to keep them in a pufferfish-specific subfolder?

Yes that can be done. I'll add this in a new commit.

ethanjli:
Right now dirhash is installed in `security_network.sh` rather than `security_tampering.sh`, which means that the `hash_check.py` only works correctly after `security_network.sh` was run. Would it be okay to move the dirhash installation into the `security_tampering.sh` script? Besides this, everything else looks good in my preliminary testing.

ethanjli (approved):

PR #217, opened 2020-10-19
Avoid setup code in SFM3019::Sensor::output when setup was skipped

ethanjli:
This PR fixes a bug in the `SFM3019::Sensor` code in which the calling code is able to skip `setup` but call `output`. When this happened, `SFM3019::Sensor` would simply attempt to run the setup code, which is incorrect behavior. If the setup code was configured to do many retries, then it would prevent the firmware from doing any other work. Now I have removed setup code from the `output` method so that it can only be reached by calling the `setup` method.


ethanjli:
This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed for to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, probably would not be able to find someone to review it very quickly, and is a bugfix. I will merge in this PR now.

For records-keeping:

PR #218, opened 2020-10-21
Fix/backend file I/O directory

Tanmay06:
Fixes issues where there is no `statestore` directory in the location. Also fixes empty `Handler: ` exception logs.


ethanjli (approved):
From cursory testing, it looks good to me. I would suggest changing the file open/close messages from INFO level to DEBUG level for normal operation. However, it would be reasonable to have some sort of INFO-level logging message for the state initialization at startup to communicate what the File I/O stuff is doing (e.g. if it created a directory, or successfully loaded something, etc.) - that message would jut occur once, rather than being repeated.

For records-keeping:

PR #220, opened 2020-10-25
Refactor FDO2 Sensor for longer setup (up to 5 sec) & non-blocking I/O

ethanjli:
This PR:

- Refactors the FDO2 `Sensor` to use non-blocking I/O by decomposing management of states and state transitions into a `StateMachine`.
- Fixes problems observed in FDO2 sensor initialization when the Nucleo-H743ZI2 board was reset using the RESET button rather than by unplugging and plugging in the USB cable: in some cases, setup would fail. What I have observed in the logic analyzer is that sometimes multiple start-broadcast requests need to be sent before a response is received in a reasonable time. The fix is to increase the duration of time that the setup routine retries sending requests before giving up and reporting a failure. I have adjusted this to increase from 1 sec to 5 sec, though in testing I have always seen the sensor either complete setup quickly (<1 sec) or take ~3 sec to complete setup successfully.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/FDO2/Device.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/FDO2/Device.h
@@ -61,9 +61,6 @@ class Device {
 
   static const uint16_t broadcast_interval = 100;  // ms
 
-  static constexpr Responses::Bcst expected_bcst{broadcast_interval};
-  static constexpr Responses::Vers expected_vers{8, 1, 341, 15};
-
   explicit Device(volatile HAL::BufferedUART &uart) : uart_(uart) {}
 
   /**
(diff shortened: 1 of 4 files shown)


ethanjli:
@Tanmay06 Can you and the rest of your team test this PR to check whether it fully fixes the problem you reported? I expect that you should now see that the Nucleo board's LED1 sometimes starts with a brief (<1 sec) solid ON before transitioning into the setup success indicator pattern (2 sec of brief blinks before transitioning into solid OFF), while other times it will start with a longer (~3 sec) solid ON before transitioning into the setup success indicator pattern. It should never start with a very long (5 sec) solid ON before transitioning into the setup failure indicator pattern (very rapid flashing forever).

This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed for to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, and is a bugfix. I will merge in this PR after the Bharat Forge team tests this and confirms that it fixes the problem they have reported, but before a full code review is completed.

For records-keeping:

Tanmay06 (approved):
I've tested this and it works as expected.

PR #225, opened 2020-10-28
Fix bug where the error integral was always set to max value

ethanjli:
This PR:

- Fixes a bug from when I had copy-pasted bounds-checking code for the error integral term in the PI controller. In this bug, the error integral term was allowed to grow without limits, even though I had defined an upper limit; effectively it made the upper limit behave like a **lower** limit. This should probably fix the issues with overshooting after saturation observed by the BF team in testing today.
- Changes the PI controller output limits into `static const` parameters rather than leaving them hard-coded into the implementation of the `transform` method.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/Algorithms.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/Algorithms.h
@@ -18,6 +18,8 @@ class PI {
   void transform(float measurement, float setpoint, float &actuation);
 
  private:
+  static constexpr float out_max = 1;
+  static constexpr float out_min = 0;
   static constexpr float p_gain = 0.00001;
   static constexpr float i_gain = 0.0002;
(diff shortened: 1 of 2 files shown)


ethanjli:
@Tanmay06 Can you try that test where we tried to do with FiO2 setting of 21% and flow rate setting of 60 L/min and had wall air input pressure too low to be able to achieve 60 L/min and then made some other adjustment which caused an overshoot in the flow rate, and see if we still get the same problem in this PR?

This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, and is a bugfix. I will merge in this PR after the Bharat Forge team tests this and confirms that it fixes the problem we observed, but before a full code review is completed.

For records-keeping:

Tanmay06:
This PR works as expected. When I tried with FiO2 setting of 21% and flow rate setting of 60 L/min and the wall air input pressure was not enough for the 60 L/min and then increased the FiO2 to 60% / 100%, the flow rate didn't shoot up as it used to do before. I tested this even with the O2 input and it works fine.

PR #226, opened 2020-10-29
Read conversion factors from SFM3019

ethanjli:
This PR:
- Removes a (likely incorrect) fudge factor which had been multiplied with the default scale factor, and which is causing sensor measurements reported by the ventilator to be off by a factor of two, which likely would explain an off-by-a-factor-of-2 error the BF team had found between flow rates reported by the ventilator and actual flow rates in the output port.
- Adds functionality to write a command with a 2-byte argument, with a CRC.
- Fixes functionality for reading with CRC to use a single read command for arbitrarily long data, instead of one command per 2-byte word.
- Reads the conversion factors from SFM3019 sensors and checks to confirm that they are the expected values listed from the datasheet.


ethanjli:
@Tanmay06 Can you test out this PR with O2 + air supply at a few different FiO2 settings (21%, 60%, and 100%) and add some test data into this thread about measurement differences between the flow rate reported by the frontend and externally-measured flow rate from the output port?

This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed for to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, and includes a bugfix. I will merge in this PR after the Bharat Forge team tests this and obtains test data, but before a full code review is completed.

For records-keeping:

Tanmay06:
I've tested this PR and here are the test results for some scenarios that you mentioned on slack:  https://docs.google.com/spreadsheets/d/1s3AgTBJu_aPdPp0T3XgKbET2ecVFiSr4CUljBRrPrzY/edit#gid=1520171778

Tanmay06:
I've done some testing with O2 + air. This [sheet](https://docs.google.com/spreadsheets/d/1s3AgTBJu_aPdPp0T3XgKbET2ecVFiSr4CUljBRrPrzY/edit#gid=1002650474) has some flow rate readings at various FiO2 levels.

ethanjli:
Interesting - it looks like the actual output flow rate is always ~90% of the flow rate measurement reported by the ventilator, regardless of O2 concentration. From this data, I think we can say that the remaining measurement error is very unlikely to come from the composition of the gas. @hongquanli , do you have any ideas about what might be causing the remaining 10% error?

I will merge in this PR now because it sounds like the system has behaved reasonably during testing.

hongquanli:
what's status of the active safety valve and the pressure measurement port on the injection molded piece? Are they closed/sealed?

PR #227, opened 2020-10-29
New Designs - Dashboard & Sidebar

Sudhir-dev:
This PR includes 

* Sidebar with new Design
* Dashboard page design

This PR does does not include
* New Sidebar - Close & Lock Screen Icon (Need to be provided by UI/UX team)
* New Sidebar - Home button navigation (Not sure which page to be navigated)


ethanjli:
This generally looks good to me. A few things I've noticed which need to be changed in either this PR or a future PR:

- The "More Values" and "More Controls" buttons should be removed from the HFNC layout.
- The Controls toolbar is divided into four areas, two of which are FiO2 and Flow Rate, and two of which are blank areas. Instead, it should only be divided into two areas.
- The "Modes", "Set Alarms", "All Values", and "Settings" screens during active ventilation should have the same sidebar as the dashboard - currently they have the same sidebar as the "Quick Start" screen during standby. I believe the set of screens accessible from the sidebar during active ventilation will be changing anyways - so I'll let you decide whether to fix the sidebar for those screens now or later.
- The Home button in the sidebar from the dashboard should have the same effect as the "hide sidebar" button.
- The "More Waveforms" button from the dashboard appears to be layered on top of everything else, including the Lock Screen overlay and the sidebar overlay.

Sudhir-dev:
Have fixed issues mentioned except for 

"The Home button in the sidebar from the dashboard should have the same effect as the "hide sidebar" button."

Once the sidebar layout is changed according to Standby/Ventilating state, Home button could be used to navigate "Dashboard" page

Sudhir-dev:

Sumitzways:
Improvement Required:- 

- On set alarm page after clicking on Apply changes button we should have "Successfully change Message /Pop-up"  or else redirect it on home page.  Same for Settings page.  Settings -> Display.
 
-Disable the mode expect HFNC for now because by mistake if user select other mode then waveforms getting freeze.

ethanjli (approved):

ethanjli:
Another small last-minute change (either for this PR or a future PR): the "Main view" menu button should be removed from the top bar, as it is no longer needed.

PR #228, opened 2020-10-29
Ensure that valves are closed at initialization

### LINKED ISSUE #223: Idle flow at ventilator startup.
**Describe the bug**
When the ventilator is started we get around 2-3 l/min of idle flow until the frontend has been completely booted up. This starts when the RaspberryPi OS has booted up till the frontend is loaded. 

**To Reproduce**
1. Open the Air and O2 inputs.
2. Turn on the ventilator.

**Expected behavior**
There should not be any flow until instructed through the frontend.

### LINKED ISSUE #224: Add constraint on max flow rate.
Since the flow rate higher than 80 L/min is consider to be a health risk, there should a warning for the user and limiter in the software that won't allow the such setting. 
The frontend should warn the user and limit the the setting when the user tries to change the flow rate to anything higher than 80 L/min. Simultaneously, the firmware should also have a constraint that blocks any such high flow rate settings .

ethanjli:
This PR:

- Makes the firmware use the `ParametersRequest.ventilating` field to determine whether to run the simulator and control loop.
- Changes PI controller setpoints to 0 when the the `ventilating` field is set to `false`.
- Overrides the PI controller to close the valves when flow rate setpoint is 0 L/min.
- Makes the backend application turn off ventilation at startup, regardless of what the value of `ParametersRequest` loaded from file is. This is only a workaround and is undesirable behavior, we will need to fix this later (in fact we might just eventually want to use the value of `ParametersRequest` from file, since the `ventilation` flag should be set to off with a clean shutdown of the RPi or a clean close of the backend program).

### DIFF
--- a/backend/ventserver/application.py
+++ b/backend/ventserver/application.py
@@ -60,7 +60,7 @@ async def main() -> None:
         server.ReceiveOutputEvent
     ] = channels.TrioChannel()
 
-    # Initialize State
+    # Initialize state with defaults
     all_states = protocol.receive.backend.all_states
     for state in all_states:
         if state is mcu_pb.ParametersRequest:
@@ -71,6 +71,7 @@ async def main() -> None:
         else:
             all_states[state] = state()
 
+    # Load state from file
     states: List[Type[betterproto.Message]] = [
         mcu_pb.Parameters, mcu_pb.CycleMeasurements,
         mcu_pb.SensorMeasurements, mcu_pb.ParametersRequest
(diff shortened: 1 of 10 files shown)


ethanjli:
@Tanmay06 Can you try out this PR to see if you're able to reproduce the issue you had seen with non-zero flow rate between when the backend starts and when the frontend starts? I haven't been able to reproduce the issue, but my intuition is that it should not happen after the changes from this PR.

This PR is in the critical path of development, passes clang-tidy and clang-format checks, is needed to support hardware bring-up at Bharat Forge, is of at least comparable quality to the rest of the firmware codebase, and is a bugfix. I will merge in this PR after the Bharat Forge team tests this and confirms that it fixes the problem we observed, but before a full code review is completed.

For records-keeping:

Tanmay06:
I've tested this patch, it has fixed the initial idle flow issue. With this PR I'm not able to reproduce the issue we had seen with non-zero flow rate between when the backend starts and when the frontend starts. I've tested it out using a flow sensor at the output of the ventilator, and I didn't see any flow reading during the startup.

Tanmay06 (approved):

PR #236, opened 2020-11-06
Frontend is killed when the modal is operated or frontend is idle.

Tanmay06:
Increased duration for which the backend will wait for last frontend message. This PR needs to be tested before merging.


ethanjli (approved):
I tested this and it seems to work fine on the Raspberry Pi. @Tanmay06 If you can press the "squash and merge" button, please go ahead and do so; otherwise, I'll do it in on Friday PST.

PR #237, opened 2020-11-09
Set HFNC fiO2 & flow rate

### LINKED ISSUE #230: Set HFNC flow rate upper limit to 80 L/min
Right now the "Set Value" dialog for flow rate in HFNC mode allows us to choose flow rates from 0 L/min to 100 L/min. We should change this to only allow flow rates from 0 L/min to 80 L/min.

Sudhir-dev:
This PR includes

* Flow rate upper limit to 80 L/min 
* FiO2 input range between 21% to 100%.


Sudhir-dev:

ethanjli (approved):

PR #238, opened 2020-11-09
Toolbar & Screensaver styles

Sudhir-dev:
This PR consists of

* Toolbar new design (Event Alert)
* Screensaver page new design

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -13,3 +13,4 @@ body {
     user-select: none;
     touch-action: none;
 }
+.MuiDrawer-modal{z-index: 99;}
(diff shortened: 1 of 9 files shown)


Sudhir-dev:

ethanjli:
Basic testing of the new alarm display in the topbar generally works as I expect, except for one behavior: when an alarm condition becomes inactive on its own (i.e. the `activeLogEvents.id` array becomes empty), the alarm is still shown as an active event in the topbar, rather than going away.

Sudhir-dev:
> Basic testing of the new alarm display in the topbar generally works as I expect, except for one behavior: when an alarm condition becomes inactive on its own (i.e. the `activeLogEvents.id` array becomes empty), the alarm is still shown as an active event in the topbar, rather than going away.

 Have fixed this issue now

ethanjli (approved):

PR #240, opened 2020-11-12
Refactor CRC interfaces and implementations

ethanjli:
This PR:

- Creates a templated `CRCChecker` interface class, as well as `CRC8` and `CRC32` aliases which can be used as interfaces.
- Removes inclusion of the `Pufferfish/HAL/STM32/CRC.h` header from driver files so that they can compile without the STM32 libraries.
- Updates the CRC32C-based driver classes to take the `CRC32` interface in their constructors, rather than an STM32-specific `CRC32C` class.
- Generalizes the CRC8 software implementation to work for CRCs of any size in the templated `SoftCRC` class in `Pufferfish/HAL/CRCChecker.h`. This class subclasses `CRCChecker` and is usable both for any CRC8 and for CRC32C. It uses a table lookup-based algorithm rather than the naive bit-by-bit algorithm. Because it implements the `CRC32` interface, it should be possible to use the `SoftCRC32` as a drop-in replacement for the hardware-based `HALCRC32` in unit tests by dependency injection, though I haven't written a unit test confirming this.
- Defines a templated `CRCParameters` struct for initializing `SoftCRC` objects in the SFM3019 driver and other Sensirion drivers.
- Simplifies the interface of `SensirionDevice` by passing the CRC implementation in the constructor, rather than passing in CRC parameters with every method call.
- Creates Catch2 unit tests for CRC8 (with SDP3019 parameters) and CRC32C, matching the example inputs which were used in a unit test in the backend (though that test was deleted in #90 because we moved to a third-party library for CRC in python).
- Adds clang-tidy checks for the Catch2 unit tests to the Github Actions workflow.
- Fixes clang-tidy issues for the Catch2 Nonin OEM III unit tests written by Hemanth.


ethanjli:
As discussed with Renji on Slack, this PR appears to work, so I will merge this PR in first. For records-keeping:

PR #242, opened 2020-11-12
Screen lock alert & minor fixes

### LINKED ISSUE #232: Provide a sensible indication of FiO2 measurement when HFNC flow rate is 0
Currently, when HFNC flow rate is set to 0 L/min, the ventilator is still measuring and displaying the FiO2 based on the FDO2 sensor's measurement of the (stagnant) gas inside the breathing circuit. Thus it is possible, for example, to have a setting of target FiO2 = 100% and target flow rate = 0 L/min and obtain an FiO2 measurement of 63% (Sumit has seen this in testing). This may be an edge case because it seems unlikely that a flow rate of 0 L/min would be used in practice (as opposed to pressing the "Stop Ventilation" button), but perhaps it could arise in a clinical situation. We should have a way to deal with this situation based on what would make sense for clinicians.

One approach could be to display "N/A" as the FiO2 measurement when flow rate is set to 0 L/min, to indicate that the ventilator cannot provide a reasonable FiO2 measurement at that flow rate. However, there might be other ways to deal with this situation

Sudhir-dev:
This PR addresses

* Dashboard Control section - Alignment
* Screen lock - Alert popup only when tapped while locked
* Sidebar Auto close - on every navigation change
* Fio2 value is "--" when flow rate is 0
* Red Border for Active event
* Updated yarn.lock for rxjs package

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -39,6 +39,7 @@
     "redux": "^4.0.5",
     "redux-saga": "^1.1.3",
     "reselect": "^4.0.0",
+    "rxjs": "^6.6.3",
     "typescript": "^3.9.7"
   },
   "scripts": {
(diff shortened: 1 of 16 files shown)


Sudhir-dev:
Updated with following fixes

* Sidebar now closes immediately on navigation
* Home button closes sidebar

Sudhir-dev:

ethanjli (approved):
Looks good!

PR #244, opened 2020-11-17
Fix modernize-make-unique/shared IncludeStyle options

ethanjli:
This PR:
- Fixes the [`modernize-make-shared.IncludeStyle`](https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-make-shared.html) and [`modernize-make-unique.IncludeStyle`](https://releases.llvm.org/10.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/modernize-make-unique.html) options by changing them from `'0'` (which is not a valid value) to `llvm` (which is the default value). For some reason, clang-tidy 10 did not complain about this problem, but clang-tidy 11 does.
- Makes the `.clang-tidy` configuration compatible with clang-tidy 11
- Suppresses a few false positives from clang-tidy 11
- Reorganizes the declaration of variables in the firmware's `main.cpp` by moving the declaration of non-const global variables into the body of the `main` function wherever possible.

### DIFF
--- a/.github/workflows/firmware.yml
+++ b/.github/workflows/firmware.yml
@@ -117,30 +117,6 @@ jobs:
       run: |
         ./clang-tidy-all.sh Clang
 
-  clang-tidy-11:
-    runs-on: ubuntu-20.04
-
-    steps:
-    - uses: actions/checkout@v1
-
-    - name: Install dependencies
-      run: |
-        sudo apt-get install gcc-arm-none-eabi cmake
-        sudo apt-get remove -y libllvm10
-        wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-        sudo add-apt-repository -y -s 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main'
-        sudo apt-get update
-        sudo apt-get install clang-tidy-11
-
-    - name: Run CMake
-      run: |
-        ./cmake.sh Clang /usr/lib /usr/bin
-
-    - name: Run clang-tidy
-      run: |
-        alias clang-tidy="clang-tidy-11"
-        ./clang-tidy-all.sh Clang
-
   catch2-clang-tidy:
     runs-on: ubuntu-20.04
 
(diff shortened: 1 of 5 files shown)


ethanjli:
@akdhawan Can you try running your local installation of clang-tidy (version 12) on this branch and report what errors (if any) clang-tidy generates?

akdhawan:
@ethanjli After running on my local installation of clang-tidy(version 11) on this branch, I see the following output documented in the output.log
[output.log](https://github.com/pez-globo/pufferfish-software/files/5552470/output.log)

ethanjli:
The cppcoreguidelines-pro-type-vararg complaints should be resolvable by modifying the implementation to use array indices rather than deferencing pointers in an iterator style. With only approx. two exceptions (namely the `BufferedUART` objects which need to be globals so that they can be referenced in the STM32CubeMX-generated file for interrupt handlers), the cppcoreguidelines-avoid-non-const-global-variables complaints should be resolvable by instantiating everything inside a section in the main function (I'll think more about whether we want to do this tomorrow, but https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-global makes a reasonable argument about initialization orders).

ethanjli:
It turns out that actually cppcoreguidelines-pro-type-vararg complaints were a false positive, and we actually do need to work with pointers in an iterator style (because of the C++ std library functions we're using). I've suppressed those complaints. I've also moved non-const global variable declarations into the `main` function wherever possible. I ran a quick test uploading this new version of the firmware to my Nucleo board (with the FDO2 driver commented out, since my FDO2 sensor is broken) and it successfully completed setup of the other sensors and was able to send flow measurements to the backend.

@akdhawan Can you try running clang-format and clang-tidy again and report what it outputs? If there are no errors, please go ahead and do an approval through the Github PR's code review interface, and then I'll merge it in.

For records-keeping:

ethanjli:
I am going to close this PR for now - we can reopen it later as needed.

PR #245, opened 2020-11-18
Event Log Popup

Sudhir-dev:
This PR includes

* Event Alerts - Change to popup
* Add Active Alarms filter in popup
* Added AlarmMute request in proto & redux
* Alarm mute functionality in header toolbar


ethanjli:
Everything seems to work fine, except:

- When an alarm is active and I press the button for the alarm, nothing happens - currently the only way I can access the Active Alarms modal is by pressing the "Active Alarms" button from the Events Log modal.
- The backend does not recognize the new protobuf message of type code 11. I would suggest just commenting out the `{ time: 50, pbMessageType: AlarmMuteRequest },` line of `initialSendSchedule` in `frontend/src/store/controller/protocols/backend.ts` for now, and I'll uncomment that line when I work on the backend support.

Sudhir-dev:
> Everything seems to work fine, except:
> 
> * When an alarm is active and I press the button for the alarm, nothing happens - currently the only way I can access the Active Alarms modal is by pressing the "Active Alarms" button from the Events Log modal.
> * The backend does not recognize the new protobuf message of type code 11. I would suggest just commenting out the `{ time: 50, pbMessageType: AlarmMuteRequest },` line of `initialSendSchedule` in `frontend/src/store/controller/protocols/backend.ts` for now, and I'll uncomment that line when I work on the backend support.

@ethanjli Should it be allowed to mute & unmute for active alarms? 
Also, what should be other ways to access Active Alarms?

ethanjli:
I'm not sure I understand your first question - are you asking whether this PR should provide a way to do the global alarm mute/unmute from the Active Alarms modal?

For your second question: I believe there should only be two ways to access the Active Alarms: one is by tapping on the button showing the current prioritized active alarm, and the other is from the "Active Alarms" button in the Events Log modal. I don't think the UI/UX team has provided any other ways.

Sudhir-dev:
Sorry was not able to explain. I actually wanted more clarification on your point 

"When an alarm is active and I press the button for the alarm, nothing happens"

"Button for Alarm" here means the Volume Icon button or Bell Icon button or the button where alarm title is shown?
And when it's pressed what is the intended functionality?

ethanjli:
I see - when the bell icon button is pressed, it should cause the frontend to issue a request to globally silence all alarms for 2 minutes. When the button with the alarm text is pressed, it should show the Active Alarms modal. Both buttons should only be pressable when there is at least one active alarm. This appears to be the behavior from your most recent commit.

One more change which you could do in this PR or a future PR: when we toggle between the all-alarms-muted state and the all-alarms-audible state, the button with alarm text changes in length (because of the countdown timer). I think we should keep the button at the same length in both states.

For records-keeping:

Sudhir-dev:

ethanjli (approved):

PR #247, opened 2020-11-19
Trigger repo-wide checks when needed

ethanjli:
This PR:

- Makes Github Actions run checks for all domains on PR review-related events, so that we can catch potential check failures related to merging from develop
- Makes Github Actions run checks for a domain whenever its workflow configs are changed, so that we don't have to edit another file to re-run the checks (or to do force-pushes) when we only want to change the configs in a commit.
- Uploads the results of firmware builds as a build artifact.
- Adds a `yarn build` step to the frontend checks and uploads the result as a build artifact.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -4,6 +4,12 @@ on:
   push:
     paths:
       - 'backend/**'
+  pull_request:
+    types: [ready_for_review, review_requested]
+  pull_request_review:
+    types: [submitted]
+  milestone:
+    types: [opened, deleted]
   workflow_dispatch:
     inputs:
       git-ref:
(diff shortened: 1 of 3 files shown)


ethanjli:
@Sudhir-dev I've added a Github Actions step to upload the frontend build artifacts. However, currently the build fails because all those "unused import"/"unused definition" eslint warnings are treated as errors in the Github Actions build environment. For this PR I am allowing the build job to fail with the `continue-on-error: true` line in the `frontend.yml` config file; once you get those warnings resolved in a separate PR, then you can remove that line. Can you do an approval on this PR so I can double-check that I am able to merge in the PR even though the frontend build job is failing?

Sudhir-dev (approved):

PR #249, opened 2020-11-24
Added two test files for protocols unit tests

renjipanicker:
@ethanjli Please review the python test cases in this branch.


ethanjli:
This is a start, but I would suggest that instead of starting by writing code, it may be better to first write BDD test specifications (without code) for various tests/test cases as described on [our Notion](https://www.notion.so/pufferfish/Testing-13ce58d72a154b6883fbd57b71d37f76).

ethanjli:
This will be replaced by a new PR to be opened by @rohanpurohit

Issue #233, opened 2020-10-30
Frontend is killed while changing the set values.


Tanmay06:
**Describe the bug**
The backend kills the frontend, when the user is setting any value through modal. Since the backend waits for ~1s of delay of message from the frontend before spawning the kill process, it seems like no messages from the frontend are sent when the modal is being operated.

**To Reproduce**
1. Start the backend and the frontend in a linux machine.
2. Open any modal
3. Keep changing the value till the chromium browser is killed.

**Expected behavior**
The backend should not kill the frontend.

**Screenshots**

**Device (please complete the following information):**
 - Hardware Model: [Raspberry Pi 4B]

**Software (please complete the following information):**
 - Domain: [backend, frontend]
 - Version: [0c3dfc4]

ethanjli:
This issue is related to #124: for some reasoning, opening the modal or doing other React graphics-redraw-heavy things appears to momentarily freeze all processing of the frontend (and with freeze durations which vary across different trials of the same test); when no waveform data is received by the frontend for a graph for 2 sec, then the graph resets itself. Obviously the backend is still sending that data, it is just getting dropped or rejected by the frontend (presumably because the frontend is frozen). I am not sure how we can solve this problem - @Sudhir-dev , do you have ideas for where the performance bottleneck might be that is causing such a long freeze in the frontend?

ethanjli:
Out of curiosity, I had opened the Preact Dev Tools with redraw highlighting, and it looks like the frontend is wasting a lot redraws on things which aren't actually changing - this includes various pieces of the topbar, as well as almost everything in the dashboard. This is a big regression from dashboard implementation for the V1 UI/UX design, where only the curves themselves and the changed numbers triggered redraws. @Sudhir-dev How much of the dashboard consists of nested components, vs. containers which are passed in by dependency injection?

ethanjli:
We will implement a work-around for this by limiting the rate at which the modal dialog's increment and decrement buttons may be clicked, to a rate which is low enough that the frontend can do some processing with the communication protocol in between button clicks. We can start by trying a rate-limit of a minimum of 50 ms between button-presses. In between those 50 ms intervals, the button should be temporarily disabled.

This applies to all increment and decrement buttons.

PR #250, opened 2020-11-26
Re-rendering Optimization

Sudhir-dev:
This PR includes

* Re-rendering Optimization on Toolbar, Screensaver page and Dashboard page
* Chrome Error fixes related to Material UI Grid component

### DIFF
--- a/frontend/src/modules/app/EventAlerts.tsx
+++ b/frontend/src/modules/app/EventAlerts.tsx
@@ -1,6 +1,6 @@
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect } from 'react';
 import { Alert } from '@material-ui/lab';
-import { useDispatch, useSelector } from 'react-redux';
+import { shallowEqual, useDispatch, useSelector } from 'react-redux';
 import { Button, Grid, makeStyles, Theme, Typography } from '@material-ui/core';
 import VolumeOffIcon from '@material-ui/icons/VolumeOff';
 import VolumeUpIcon from '@material-ui/icons/VolumeUp';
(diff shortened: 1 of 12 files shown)


ethanjli:
I had been expecting that you would be instantiating a bunch of containers by connecting various components to various selectors, and then passing the containers into the props of parent components - an example of how I did this is in [`FlowGraphInfo.tsx`](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/dashboard/containers/FlowGraphInfo.tsx). However, it looks like what you did instead was pass in selectors as props of parent components, and then use it in a memoized component. Your approach is a bit harder for me to understand implementation details (e.g. memoized components, shallow equality comparison, and what are the implications of passing a selector in as a prop) but it makes a lot of sense, and I suspect your approach leads to simpler and more maintainable code than my approach of instantiating a bunch of containers everywhere, where each container is a single component connected to a different selector. Is the approach you've taken, of passing selectors in the props of components, a standard practice in the Redux world?

ethanjli:
I tested it on my computer and things work well. I've also identified some more areas for optimization to fix in either this PR or a future PR:
- When the Set New value modal is opened, it is constantly being redrawn all the time - I suspect this is where some of the input freezing related to #233 is coming from.
- When the Events Log is opened, it is also constantly being redrawn all the time
- When the Set Alarm limits modal is opened, it is not constantly being redrawn all the time. However, it is redrawing the entire modal whenever an alarm limit value is changed.
- The Display tab of the Settings screen is constantly being redrawn all the time
- When the red visual alarm border is added or removed, the entire screen is redrawn.

For records-keeping:

Sudhir-dev:

ethanjli (approved):

PR #259, opened 2020-12-14
Backend Connection lost - Audio Alarm in Frontend

Sudhir-dev:
This PR includes

* Heartbeat update on every message from backend
* Trigger Frontend based alarm on lost of connectivity for more than 2 seconds
* Audio Alarm


ethanjli:
This looks good to me! A few observations from testing, which are outside the scope of this PR:
1. If I start the frontend and press "start ventilation" before the backend server has started, I will get the "software connectivity lost" alarm, and it starts immediately rather than after a 2-second delay. I think it would be better to prevent the ventilator from entering "ventilating" mode if the backend is disconnected - this will require some design work for the backend and microcontroller, and it would probably involve using the `ventilating` field of the `Parameters` protobuf message received from the backend (i.e. the app only opens the dashboard screen once `Parameters.ventilating` becomes true). Or a simpler solution might just be to check whether the backend is connected and just show an error message if it's disconnected and the user presses "Start Ventilation".
2. If I pause ventilation while an alarm is still active, the red border still shows up in the Standby screen and there is no way to see what the alarm is. However, because there are some alarms that may fire in standby mode (e.g. low battery), we may need to add the alarms section of the topbar into standby mode as well.

For records-keeping:

Sudhir-dev:

ethanjli (approved):

PR #266, opened 2020-12-22
Implement alarms in backend simulator

ethanjli:
This PR:

- Decomposes the backend simulator module into several classes, which will also be the rough structure for the firmware alarms implementation.
- Adds reusable alarms functionality for parameters based on alarm limits requests.
- Simplifies the alarms-related message fields of the communication protocol by unifying min and max record types into a `Range` message type.
- **Renames the backend simulator module from `ventserver.simulation` to `ventserver.simulator`**, as `simulation` is now a subpackage

For records-keeping:

1. This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? **Yes**
2. Were any of these contributions also part of work you did for an employer or a client? **No**
3. Does this work include, or is it based on, any third-party work which you did not create? **No**

PR #267, opened 2020-12-22
Implement python unit test cases

rohanpurohit:
This PR:
Added Test cases for:

* protocols/backend.py
* protocols/server.py
* simulation.py


ethanjli:
The simulation module has been refactored in #266, which I've just merged develop, which I suspect means that the tests in this PR for `simulation.py` will need to be reorganized - can you

rohanpurohit:
updated the cases with proper method calls.
for setting simulator service initial time , doing this https://github.com/pez-globo/pufferfish-software/pull/267/commits/05128407d071c3b702f5a7df3414ff2dd9a80c3b#diff-02243302b9789b0287db9ab95d99f5a94e4ed6e51d0a5fcd2fad474977c22323R88
@ethanjli , is that ok ?

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
+Scenario: Backend receive event behaves properly
+Given: A ReceiveEvent object
+When: ReceiveEvent class is initialised without data
```
ethanjli:
```suggestion
When: the object is initialised without data
```

or

```suggestion
When: no data is given as input
```

Rationale: the things we initialize are objects, not classes

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Then: has_data function returns false   
+
+Scenario: Backend receive event behaves properly
+Given: A ReceiveEvent object
+When: time is passed as input
```
ethanjli:
```suggestion
When: time is given as input
```
Rationale: I'm being pedantic, but in these ReceiveEvent object tests I would suggest using `given` rather than `passed` - since "passing" evokes passing values into a function (e.g. "pass by value" and "pass by reference" in C++ terminology), and here we're testing a method which checks public data members.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Given: A ReceiveEvent object
+When: mcu event is passed as input
+Then: has_data function returns true
+
+Scenario: Backend receive event behaves properly
+Given: A ReceiveEvent object
+When: frontend data is passed as input
+Then: has_data function returns true
+
+Scenario: Backend receive event behaves properly
```
ethanjli:
```suggestion
Scenario: The backend's output event behaves properly
```
Rationale: this event is used for the backend receiver's output, the backend sender's output, and the backend sender's input.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
```
ethanjli:
```suggestion
Scenario: The backend receiver's input event behaves properly
```
Rationale: this wording for the scenario is more specific, and it is a bit more descriptive about the class we're testing, by associating it with the backend's receiver (which is implemented by the `ReceiveFilter` class).

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Given: A OutputEvent object
+When: ouptut event has no data
+Then: has_data function returns false
+
+Scenario: Backend receive event behaves properly
+Given: A OutputEvent object
+When: output event has data
+Then: has_data function returns true
+
+Scenario: Backend send event behaves properly
```
ethanjli:
```suggestion
Scenario: The backend's output event behaves properly
```

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Given: a announcement object
+When: backend recieves announcement strings as input
+Then: has_data function returns true
+
+#ReceiveFilter
+
+Scenario: Receive Filter behaves properly
+Given: receive event with no data
```
ethanjli:
```suggestion
Given: a `ReceiveFilter` object which has not yet received any events
```
Of course, we will also need to write tests for `ReceiveFilter` objects which have already received various sequences of events.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+When: empty receive event is passed as input to receive filter
```
ethanjli:
```suggestion
When: a receive event with no data is passed as input to the receive filter
```

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Then: output of receive filter None
```
ethanjli:
```suggestion
Then: the output of the receive filter is None
```

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+
+Scenario: Receive Filter behaves properly
+Given: mcu receive event
+When: mcu receive event is passed as input to receive filter
+Then: output of receive filter is not None
```
ethanjli, line 61: We will also need to write tests which make more specific assertions about what the contents of the output of the receive filter should look like, under different conditions. This will probably require writing multiple new tests.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+When: output event has data
+Then: has_data function returns true
+
+Scenario: Backend send event behaves properly
+Given: a announcement object
+When: backend recieves announcement strings as input
+Then: has_data function returns true
+
+#ReceiveFilter
+
```
ethanjli, line 52: We will also need to write a test for what we expect to happen when we pass `None` into the `input` method. And we will need to write a test for what we expect to be returned by `output` when we have not passed any data into the `input` method, and when we called `input` a few times but we've also already called `output` more times.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+When: announcement event is passed as input to send filter
+Then: output of send filter is not None
+
+Scenario : Send Filter behaves properly
+Given: An Event of type SendEvent with data
+When: Given event is passed as input to send filter
+Then: output of send filter is not None
+
+#Protocols
+
```
ethanjli, line 92: We will also need to write tests where `Then` is `the output of the function is None`.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Scenario: get_mcu_send function behaves properly
+Given: A OutputEvent from mcu
```
ethanjli:
```suggestion
Given: The get_mcu_send function
```
I think the `Given`s should be the functions.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+When: given event is passed to get_mcu_send function
```
ethanjli:
```suggestion
When: an OutputEvent object containing data in the mcu_send field is passed to the function
```
Maybe this is too specific to the names of the data members of the `OutputEvent` class, but we need to be at least somewhat specific.

ethanjli (commented):
I've done an initial review of `backend/tests/unit/protocols/backend.txt`.

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion, new tests will push in a different pull request

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Changed as per suggestion

backend/tests/unit/protocols/backend.txt
rohanpurohit: Will push new tests in a different pull request

backend/tests/unit/protocols/backend.txt
rohanpurohit: Will push new tests in a different pull request

backend/tests/unit/protocols/backend.txt
rohanpurohit: Will push new tests in a different pull request

backend/tests/unit/protocols/backend.txt
rohanpurohit: Resolved

backend/tests/unit/protocols/backend.txt
rohanpurohit: updated as per suggestion

backend/tests/unit/protocols/backend.txt
ethanjli: I still see this wording in the latest commit - have you pushed all your changes?

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
+Scenario: Backend receive event behaves properly
+Given: A ReceiveEvent object
+When: the object is initialised without data
+Then: has_data function returns false
```
ethanjli: Based on what we discussed today with the `CRCElement` class in the C++ code, I wonder if it might be more consistent with other tests to structure these tests differently, so that the Given corresponds to our instantiation of the object. For example:

Given: A ReceiveEvent object constructed with default parameters
When: the object is not modified
Then: the has_data method returns false

Given: A ReceiveEvent object constructed with default parameters
When: a new value is set for the time field of the object
Then: the has_data method returns true

Given: A ReceiveEvent object constructed with a value for the time parameter
When: the object is not modified
Then: the has_data method returns true

Given: A ReceiveEvent object constructed with a value for the time parameter
When: the time field of the object is set to None
Then: the has_data method returns false

and so on.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+
+Scenario: The backend's output event behaves properly
+Given: a announcement object
+When: backend recieves announcement strings as input
+Then: has_data function returns true
+
+#ReceiveFilter
+
+Scenario: Receive Filter behaves properly
+Given: a `ReceiveFilter` object which has not yet received any events
```
ethanjli:
```suggestion
Given: a ReceiveFilter object which has not yet received any events
```
Oops, I had accidentally used Markdown when I made the code change suggestion for this line.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+When: a receive event with no data is passed as input to the receive filter
```
ethanjli, line 55: We could also, for example, add a 'Then: the internal buffer is empty', which would correspond to checking the state of the `_buffer` "private" member - but that test may not be so useful here. In other cases where there's complicated logic (particularly in the `output` methods of the backend and server protocols, as well as in the `input` and `output` methods of application protocols such as for list synchronization), we need to write some tests on the state of private members before and after method calls, to ensure that assumptions, preconditions, and invariants are maintained by the methods. Anyways, you can keep those for a future PR.

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,91 @@
...
+Then: the output of the receive filter is None
+
+Scenario: Receive Filter behaves properly
+Given: mcu receive event
```
ethanjli: You will also need to update the Givens and Whens for this scenario and the next six scenarios, in a similar way to how I had previously suggested changes to the scenario above.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
+Scenario: check if server connects to frontend
```
ethanjli: We're only testing the `FrontendConnectionEvent` class here; this test needs to be rewritten like how we rewrote the test for the backend protocol's event classes. It's not clear to me what "event is passed at some time" means in the context of this class. Tthere is no frontend WebSocket connection to make assertions on. The way the "Then" is written here is does not reflect that.

backend/tests/unit/protocols/backend.txt
ethanjli: I've seen the wording you used for scenarios in your tests of event classes in `tests/unit/protocols/server.txt`, and I like it. Maybe we should change this to "Scenario: The backend receiver's input event correctly reports whether it has data" (and apply this pattern to all other relevant tests in this PR, too).

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
...
+Given: Frontendconnection event object
+When: event is passed at some time
+Then: frontend connection has data
+
+Scenario: check if receive event has data
+Given: A ReceiveRvent object
+When: test data is given to it
```
ethanjli: What kind of data is given to the ReceiveEventObject? How is it given? See also my first comment on lines 2-4 of `tests/unit/protocols/backend.txt`, that applies to these tests too.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
...
+
+Scenario: check if receive output event has data
+Given: A backend output event object
+When: backend output event is passed to ReceiveOutputEvent
+Then: server output has data
+
+Scenario: check if server output event has data
+Given: A SendOutputEvent object
+When: data is written to it
+Then: server output event has data
```
ethanjli: These tests need to be rewritten like how we rewrote `tests/unit/protocols/backend.txt`.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
...
+Scenario: Check if make serial receive gives receive event object
+Given: example data and times
+When: data is written to it
+Then: receive event is not None
+
+Scenario: websocket receive behaves properly
+Given: example data and times
+When: data is written to it
+Then: receive event is not None
+
```
ethanjli: We also need tests for `make_rotary_encoder_receive`.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
...
+Given: A backend output event object
+When: backend output event is passed to ReceiveOutputEvent
+Then: server output has data
+
+Scenario: check if server output event has data
+Given: A SendOutputEvent object
+When: data is written to it
+Then: server output event has data
+
+Scenario: Check if make serial receive gives receive event object
```
ethanjli: For all tests, let's write Scenarios as descriptions of the behavior of the code rather than as commands to the tester. Here is a description of the coder's behavior: `The make_serial_receive function creates a ReceiveEvent object correctly`. Here is a command to the tester: `Check if the make_serial_receive_function creates a ReceiveEvent object correctly`.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,39 @@
...
+Given: example data and times
```
ethanjli: Let's see if things are okay when we standardize all unit tests for functions using `Given: the <function_name> function, When: it is called with <the input arguments to the function> is passed to the function`, rather than `Given: <the input arguments to the function>, When: the inputs are used`.

backend/tests/unit/simulation/alarm_limits.txt
```diff
@@ -0,0 +1,34 @@
+Scenario: Alarm Limit services behave properly
```
ethanjli: Let's have different scenarios for the PCAC class vs. the HFNC class vs. the `Serivces` class.

backend/tests/unit/simulation/alarm_limits.txt
```diff
@@ -0,0 +1,34 @@
...
+Given: A PCAC Alarm Limit service
+When: parameter service is not equal to pc_ac or HFNC
```
ethanjli: I don't think the `PCAC` alarm limits service class checks any service parameter, rather the `Services` class does those checks. If you're actually testing the `Services` class and not the `PCAC` alarm limits service class, you'll need to modify your Given to be consistent with that. This comment also applies to other tests in this file and in the `alarms.txt` and `parameters.txt` and `simulators.txt` files.

backend/tests/unit/simulation/alarm_limits.txt
```diff
@@ -0,0 +1,34 @@
...
+Then: Alarm Limits are updated as per Alarm Limit requests
+
+Scenario: Alarm Limit services behave properly
+Given: A PCAC Alarm Limit service
+When: Requested alarm limit lower value is lesser than defined lower value
+Then: Alarm Limits are not updated as per Alarm Limit requests
+
+Scenario: Alarm Limit services behave properly
+Given: A PCAC Alarm Limit service
+When: Requested alarm limit upper value is greater than defined upper value
```
ethanjli: Instead of "the defined upper value", I think it would be clearer to say "the maximum allowed upper value for the alarm limit". This comment also applies to other tests in this file.
```suggestion
When: The requested upper value for the SpO2 alarm limits range is greater than the maximum allowed upper value for the range.
```

backend/tests/unit/simulation/alarms.txt
```diff
@@ -0,0 +1,34 @@
+Scenario: alarm services behaving properly
+Given: a pc_ac alarm service object 
+When: parameter mode is not equal to pc_ac
+Then: alarms are not updated for parameters
+
+Scenario: alarm services behaving properly
+Given: a pc_ac alarm service object 
+When: parameter ventilating mode is false
+Then: alarms are not updated for parameters
```
ethanjli: My understanding is that the alarms are actually updated - all active alarms are deactivated.

backend/tests/unit/simulation/alarms.txt
```diff
@@ -0,0 +1,34 @@
...
+
+Scenario: alarm services behaving properly
+Given: a pc_ac alarm service object 
+When: sensor measurments spo2 is less than lower alarm_limit for spo2
+Then: alarm for spo2 too low is triggered
```
ethanjli: The alarm should only be triggered if there isn't already an active alarm on SpO2 too low, otherwise, the previous alarm for that same condition continues to remain active. This comment also applies to the FiO2 alarms.

backend/tests/unit/simulation/alarms.txt
```diff
@@ -0,0 +1,34 @@
+Scenario: alarm services behaving properly
+Given: a pc_ac alarm service object 
+When: parameter mode is not equal to pc_ac
+Then: alarms are not updated for parameters
+
+Scenario: alarm services behaving properly
+Given: a pc_ac alarm service object 
+When: parameter ventilating mode is false
+Then: alarms are not updated for parameters
+
```
ethanjli: Presumably in a future (lower-priority) PR we will also test what happens when the value is within the limits: if no alarm is active on that value, nothing changes; if an alarm was previously active on that value, then the alarm gets deactivated.

backend/tests/unit/simulation/parameters.txt
```diff
@@ -0,0 +1,29 @@
...
+
+Scenario: parameters services behaves properly
+Given: A PCAC parameter service
+When: parameter mode is equal to pc_ac
+Then: parameters are updated as per the parameter requests
+
+Scenario: parameters services behaves properly
+Given: A PCAC parameter service
+When: parameter request fio2 is less than fio2_min
+Then: parameters are not updated as per the parameter requests
```
ethanjli, line 38: We should make a stronger assertion: the parameter remains at the previous value it had before the request was passed in. An alternative strategy would have been to update the parameter differently than requested, e.g. if the requested value was lower than the min allowed value, we would've just updated to the min allowed value instead (which is not as per the request). I made a decision here to instead ignore the requested value entirely, and we need a test to express this explicitly so that my design decision can be reviewed by other people.

backend/tests/unit/simulation/simulators.txt
```diff
@@ -0,0 +1,41 @@
...
+Then: sensor measurements time is unchanged
+
+Scenario: PCAC simulator behaves properly
+Given: A PCAC simulator service
+When: parameter ventilating value is false
+Then: sensor measurements time is unchanged
+
+Scenario: PCAC simulator behaves properly
+Given: A PCAC simulator service
+When: time_step function returns 0
```
ethanjli: What is the `time_step` function?

ethanjli (commented):
I did a round of review on all the test description files (the `txt` files). In a few places I pointed out places where we'll need more tests in the future, but I also skipped many of those places since we are leaving those tests for future PRs. Several of my comments are for standardizing the kinds of changes we made in `backend.txt` over to the other test description files.

ethanjli:
It looks like Github Actions failed its check of `test_pcac_sensors_time_step` in  d354eac with the `isclose(20.005124912516756, 20, abs_tol=0.002)` assertion. This is probably happening because I use a random number generator to add some noise into my simulators. You should probably reinitialize the random number generator's seed for your tests involving the simulator classes, so that the tests run reproducibly. You may also need to increase your `abs_tol` for numbers which I've intentionally added noise to.

backend/tests/unit/protocols/backend.txt
rohanpurohit: i think the comments were written for the particular commit-id, therefore it is not visible with new commits? comments written in the main pr page seem to be visble for all new commits when pressed on "view changes"

backend/tests/unit/protocols/backend.txt
rohanpurohit: I have structured the descriptions as per the suggestions here : 

https://github.com/pez-globo/pufferfish-software/blob/feature/backend-test-protocols2/backend/tests/unit/protocols/backend.txt
please review.

backend/tests/unit/protocols/backend.txt
rohanpurohit: Updated. check here https://github.com/pez-globo/pufferfish-software/blob/feature/backend-test-protocols2/backend/tests/unit/protocols/backend.txt

backend/tests/unit/protocols/backend.txt
rohanpurohit: Noted, I intentionally avoided checking private members at most places because of lint checks, I'll keep in mind when writing future tests. thanks

backend/tests/unit/protocols/backend.txt
rohanpurohit: Updated in the latest commit for the file.

backend/tests/unit/protocols/server.txt
rohanpurohit: I'll be pushing the changes for this file today.
The FrontendConnectionEvent class requires a last_connection_time parameter as the last known time when the frontend was connected, 
and for the "then" statement should i then pass parameterised values for time and see if it returns same ?

backend/tests/unit/simulation/alarm_limits.txt
rohanpurohit: I have updated them here https://github.com/pez-globo/pufferfish-software/blob/feature/backend-test-protocols2/backend/tests/unit/simulation/alarm_limits.txt. 
please review

backend/tests/unit/simulation/alarm_limits.txt
rohanpurohit: Yes. I'm testing the Services class , but from the active_service parameter which is obtained from parameters class, mode value in all_states parameter, i passed in PCAC, therefore the transform method for PCAC service was called, 
so i have structured the tests in that manner, 
so Should I then only do checks on parameters for the "Services" class tests or current structure is fine?

backend/tests/unit/simulation/alarm_limits.txt
rohanpurohit: updated as suggested.

backend/tests/unit/simulation/alarms.txt
rohanpurohit: Yes! will update it.

backend/tests/unit/simulation/alarm_limits.txt
rohanpurohit: i have updated the given-when-then according to previous suggestions here https://github.com/pez-globo/pufferfish-software/blob/feature/backend-test-protocols2/backend/tests/unit/simulation/alarm_limits.txt
will update similarly for alarms.txt, parameters.txt and simulators.txt after review.

backend/tests/unit/protocols/backend.txt
ethanjli: For whatever reason, it had still shown up under the "Files Changed" tab which usually only shows comments which still apply to the latest changes. But it's gone away, so it's not an issue anymore.

backend/tests/unit/protocols/backend.txt
ethanjli: Got it. Yes, for tests it's always fine to check the values of private members and add a manual suppression for pylint

backend/tests/unit/simulation/alarm_limits.txt
ethanjli: Looks good, we will change the wording of scenarios as discussed.

backend/tests/unit/simulation/alarm_limits.txt
ethanjli: Looks good.

backend/tests/unit/protocols/backend.txt
ethanjli: Looks good.

backend/tests/unit/protocols/backend.txt
ethanjli: Looks good.

backend/tests/unit/protocols/backend.txt
ethanjli: I'll add a comment on the changed version of the descriptions.

backend/tests/unit/protocols/server.txt
ethanjli: I will add a comment on the updated version of the test description.

backend/tests/unit/protocols/server.txt
ethanjli: I will add a comment on the updated version of the test description.

backend/tests/unit/protocols/server.txt
ethanjli: I will add a comment on the updated version of the test descriptions.

backend/tests/unit/simulation/alarms.txt
ethanjli: Perhaps the different conditions of whether an alarm is already active on the out-of-range parameter should be expressed as part of the "Given" statement. I can think of several cases:

1. No alarm was already active on the parameter, but now it's out of range. Then we trigger a new alarm.
2. An alarm is already active on the out-of-range parameter, and the parameter is still beyond the limit which activated the alarm (e.g. spo2 was previously below the lower limit, and it still is so). Then we just leave the previous alarm active; we don't trigger a new alarm.
3. An alarm is already active on the out-of-range parameter, but now the parameter is beyond the limit which hadn't activated the alarm (e.g. spo2 was previously below the lower limit, and now it's above the upper limit). Then we deactivate the previous alarm and activate a new one.
4. An alarm was previously active on the parameter for being out-of-range, but now it's within range. Then we deactivate the previous alarm.
5. No alarm was already active on the parameter, and it's still within range. Then nothing changes.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,50 @@
...
+
+Scenario: The make_websocket_receive function creates a ReceiveEvent object correctly
+Given: The make_websocket_receive function
+When: The data received from websocket is given to the function
+Then: A event of type ReceiveEvent is obtained with data received from websocket
+
+Scenario: The make_rotary_encoder_receive function creates a ReceiveEvent object correctly
+Given: The make_rotary_encoder_receive function
+When: The data received from rotary encoder is given to the function
+Then: A event of type ReceiveEvent is obtained with data received from rotary encoder
```
ethanjli: In these Whens and Thens, if we can use terminology related to functions ("passing" and "returning"), we should - to make things clearer.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,50 @@
...
+Scenario: The server receives output event and reports correctly whether event has data
+Given: A ReceiveOutputEvent object is constructed with default parameters
+When: The object is not modified
+Then: has_data function returns false
+And_When: The server receives output event with data
+Then: has_data function returns true
+
+Scenario: The server sends output event properly and reports whether event has data
+Given: A SendOutputEvent object is constructed with default parameters
+When: The data is sent via serial
```
ethanjli: This object doesn't have the responsibility of pushing the bytes over a Serial connection, so the When needs to be something else.

backend/tests/unit/protocols/server.txt
```diff
@@ -0,0 +1,50 @@
...
+Then: has_data function returns true
+
+Scenario: The make_serial_receive function creates a ReceiveEvent object correctly
+Given: The make_serial_receive function
+When: The data received from serial is given to the function
```
ethanjli: These `make_*` functions should make zero assumptions about where the data is coming from, so we should remove those assumptions from the Whens. It would be clearer and more specific to describe exactly what assumptions are made about the input parameters (e.g does the parameter need to be non-`None`? Does it need to be a bytestring?)

backend/tests/unit/protocols/backend.txt
```diff
@@ -0,0 +1,106 @@
...
+
+#ReceiveFilter
+
+Scenario: The receive filter's output correctly reports whether it has data
+Given: a ReceiveFilter object which has not yet received any events
+When: a receive event with no data is passed as input to the receive filter
+Then: the output of the receive filter is None
+
+Scenario: The receive filter's output correctly reports whether it has data
+Given: A ReceiveEvent object constructed with values for the time and mcu parameter
```
ethanjli: On these tests, we need to be given a receive filter object. If you are constructing a different ReceiveEvent object for each test, it might make more sense to define it in the When. So, for example:

Given: A ReceiveFilter object constructed with ___ and which has an empty internal input events buffer
(Then: The output method of the ReceiveFilter object returns None each time it is called)
When: A ReceiveEvent object with values for ___ is given as input to the ReceiveFilter object
Then: The output method of the ReceiveFilter object returns None each time it is called

ethanjli (commented):
I did a review of the items which were updated based on prior feedback.

PR #270, opened 2020-12-24
Add measurement smoothing and debouncing with timeout hysteresis

### LINKED ISSUE #264: Signal smoothing
**Is your feature request related to a problem? Please describe.**
We need a way to smooth the displayed SpO2 values, along with FiO2 and flow rate values.

We will add functionality in the frontend to perform some kind of signal smoothing, as a reducer producing derived values (the architecture will be similar to how the waveform processes input data points to produce a derived waveform). For now we can use EWMA because it is fast and easy, and if smoothing or responsiveness are unacceptable we can look at other possible algorithms.

**Describe alternatives you've considered**
We could instead implement this in the backend, but we'll need to work out additional architectural support in the backend protocol for signal processing, and we might want to shift any complexity over to the frontend instead.

ethanjli:
This PR resolves #264 by adding a custom signal smoothing algorithm as a parameterized reusable reducer in the frontend. It involves two stages: first, the raw signal (e.g. SpO2) is slightly smoothed by an exponentially weighted moving average with a weak filtering coefficient; then, the updated average is compared against the previous average to check whether it has been "close enough" (under a threshold) for at least a minimum amount of time. If so, then it is considered to have converged, and the smoothed value will be locked at the value upon convergence, until the moving average is "far enough" (beyond a threshold) from the smoothed value for at least a minimum amount of time, in which case then the smoothed value will start following the moving average as it is considered to have started transitioning to a different level.

The second stage of this algorithm, which is kind of like "button debouncing" but for continuous variables, is necessary for us to avoid allowing a value to flicker quickly between (for example) 3.49 and 3.51, which causes a rounded integer to flicker between 3 and 4. The updated algorithm instead just locks it in at (for example) 3.49 or 3.51, depending on which value it had when it first passed the convergence check.

Timeout hysteresis is introduced by using different timeout thresholds for the convergence check vs. for the transitioning check. This allows us to treat a measurement as converging quickly and wait longer before deciding that it has started transitioning (which is useful for e.g. SpO2); or to wait longer before deciding a measurement has converged and respond sooner to value transitions (which is useful for e.g. FiO2 and flow rate).

For records-keeping:

### DIFF
--- a/frontend/src/modules/dashboard/views/HFNCMainView.tsx
+++ b/frontend/src/modules/dashboard/views/HFNCMainView.tsx
@@ -6,9 +6,9 @@ import {
   getParametersFiO2,
   getParametersFlow,
   getROXIndex,
-  getSensorMeasurementsFiO2Value,
-  getSensorMeasurementsFlow,
-  getSensorMeasurementsSpO2,
+  getSmoothedFiO2Value,
+  getSmoothedFlow,
+  getSmoothedSpO2,
 } from '../../../store/controller/selectors';
 import { a11yProps, TabPanel } from '../../controllers/TabPanel';
 import { BMIN, LMIN, PERCENT } from '../../info/units';
@@ -136,7 +136,7 @@ const HFNCMainView = (): JSX.Element => {
           <Grid container item justify="center" alignItems="stretch">
             <ValueInfo
               mainContainer={{
-                selector: getSensorMeasurementsSpO2,
+                selector: getSmoothedSpO2,
                 label: 'SpO2',
                 stateKey: 'spo2',
                 units: PERCENT,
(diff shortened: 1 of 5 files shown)


ethanjli:
@Sudhir-dev Can you do a quick sanity check on the changes I've made (reviewing code changes to everything outside `frontend/src/store/controller/reducers/derived.ts`, and also trying to run it on your computer) to see if there are any big questionable things I should consider improving in this PR before I merge?

Sudhir-dev:
> @Sudhir-dev Can you do a quick sanity check on the changes I've made (reviewing code changes to everything outside `frontend/src/store/controller/reducers/derived.ts`, and also trying to run it on your computer) to see if there are any big questionable things I should consider improving in this PR before I merge?

Yes I have verified, it works, Sensor measurements are being updated. Also Code changes on other files beyond`frontend/src/store/controller/reducers/derived.ts` is also fine. Should we consider using new smooth based selectors for all other measurement values too or later when other ventilation modes are enabled?

Sudhir-dev (approved):

ethanjli:
In future ventilation modes, I think it would make sense to evaluate each additional `SensorMeasurements` field on a case-by-case basis to see whether it needs smoothing.

PR #271, opened 2020-12-28
Dashboard Control/Setvalue Multistep popup

Sudhir-dev:
This PR includes

* Multistep popup feature
* Replacement for existing individual popup in dashboard Set/Control value section
* Temporarily disabled Audio alarm (Because of bug which will be addressed in different PR)


Sudhir-dev:

ethanjli:
Looks good! Here are some observations from testing:

- If I go from the dashboard's "Flow Rate" box to the "Set New" modal, change the value, and submit, no confirmation message modal shows up, and the modal does not close on its own.
- If I go from the dashboard's "Flow Rate" box to the "Set New" modal, change the value, and go to the "HFNC Control" screen from the breadcrumb so that I can also change the FiO2 before submitting, then the value I had entered for Flow Rate gets lost. The "dirty changes" made in the multi-step modal need to persist until the modal closes, whether by a "Cancel" action or a "Submit" action.

I'm not sure how much of this is supposed to be within the scope of this PR; we can merge it in as-is if they'll be addressed separately.

Sudhir-dev:
I have updated Confirm/Cancel Dialogs for Dirty values and persist changes across values/Controls.

ethanjli (approved):

PR #273, opened 2020-12-30
Make selectors which round values return NaN, not 0, for missing values

ethanjli:
This PR makes it possible to display "--%" instead of "0%" for SpO2 when the SpO2 sensor is not producing data (e.g. because the sensor was taken off the finger). The React components for value display can already do this; this fix makes it possible for the selectors give correct inputs with those components when data is missing.


ethanjli:
@Sudhir-dev From basic testing nothing seems to have broken for me; do you know if this change could cause any breakage? Also, can you do some independent testing to see if it breaks anything for you?

Sudhir-dev (approved):

PR #274, opened 2020-12-30
Fix firmware bug and remove RR from HFNC

ethanjli:
This PR fixes a bug introduced by #266, which updated the set of allowed messages between the firmware and the Python backend. Specifically, it removed any message of type 1, but it did not correctly update the array of nanopb protobuf message descriptors: it just deleted the message descriptor in position 1 instead of replacing it with a placeholder. This meant that when the firmware tried to use message descriptor 7, it tried to index past the end of the array and caused a segfault which made the firmware completely stop.

This PR also removes respiratory rate from the HFNC simulations, as respiratory rate is not part of HFNC mode. Note to @Sudhir-dev: can you replace the RR box with HR (for heart rate, in units of BPM) in the frontend? In a future PR I will add a `hr` field to `SensorMeasurements` and update the firmware to use the heart rate measurement reported by our SpO2 sensor, as well as add heart rate into our simulators.

This PR also unifies some code between the backend's `application` and `simulator` top-level modules to resolve pylint duplicate-code complaints.

For records-keeping:

### DIFF
--- a/backend/ventserver/application.py
+++ b/backend/ventserver/application.py
@@ -67,7 +67,7 @@ async def main() -> None:
         if state is mcu_pb.ParametersRequest:
             all_states[state] = mcu_pb.ParametersRequest(
                 mode=mcu_pb.VentilationMode.hfnc, ventilating=False,
-                rr=30, fio2=60, flow=6
+                fio2=60, flow=6
             )
         else:
             all_states[state] = state()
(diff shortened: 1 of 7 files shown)


ethanjli:
@rohanpurohit Can you do a Github review where you just comment on what impacts, if any, this PR will have on the unit tests?

rohanpurohit:
thanks for this 3b8146249f8f2cd9d82b43f192fd0b80898c86f, i added a test case for it and it works, other than that there is no impact on other unit tests !

rohanpurohit (approved):

PR #277, opened 2021-02-08
Battery Icon, Version number, RR to HR

### LINKED ISSUE #233: Frontend is killed while changing the set values.

### LINKED ISSUE #257: Display Heart Rate instead of Respiratory Rate in frontend
Currently the frontend in HFNC mode is displaying and providing alarm limits for Respiratory Rate; however, we will not be able to measure this. Instead, we should show Heart Rate (which we can measure from a sensor on the microcontroller and report through the backend). This will have units of "bpm" or "beats/min" instead of "breaths/min".

### LINKED ISSUE #258: Display of version number/hash in frontend
We need a way to display the version of software in the frontend, so that bug reports can easily be associated with a version number.

At a minimum (and as our strongest source of truth) we should use git commit hashes; the first 6-8 chars should be sufficient. The deployment's `install.sh` script should query git for the hash of the current commit and insert it into an environment variable, which the frontend will use when building the frontend. Then the frontend should display that in the settings screen. If a git tag with a semantic version number (e.g. "v1.1.0") is associated with the current commit, the deployment's `install.sh` script should use that as the displayed version number instead of the git commit hash; however, it may still be useful to display both the semantic version number and the git commit hash, at least during development/testing phase.

Ideally we would have a place to show this on both the landing page and the settings screen; the settings screen should be detailed (e.g. if there are multiple version numbers on different components), while the landing page should be general (only one version number)

Sudhir-dev:
This PR includes

* Battery icons for different battery levels & charging level
* Software version number in settings page via environment variable "REACT_APP_VERSION"
* Replacement of Respiratory Rate to Heart Rate
* Removed sliders from Alarm Modal
* Added Rate limiter (100ms) to Increment/Decrement value clicker

### DIFF
--- a/frontend/src/modules/alarms/AlarmsPage.tsx
+++ b/frontend/src/modules/alarms/AlarmsPage.tsx
@@ -119,7 +119,7 @@ const alarmConfiguration = (ventilationMode: VentilationMode): Array<AlarmConfig
       return [
         { label: 'FiO2', stateKey: 'fio2' },
         { label: 'SpO2', stateKey: 'spo2' },
-        { label: 'RR', stateKey: 'rr' },
+        { label: 'HR', stateKey: 'hr', max: 200 },
       ];
   }
 };
--- a/frontend/src/modules/app/ToolBar.tsx
+++ b/frontend/src/modules/app/ToolBar.tsx
@@ -1,6 +1,6 @@
 import { AppBar, Button, Grid } from '@material-ui/core';
 import { makeStyles, Theme } from '@material-ui/core/styles';
-import React, { useCallback, useEffect } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
 import { Link, useLocation } from 'react-router-dom';
 import { getClockTime } from '../../store/app/selectors';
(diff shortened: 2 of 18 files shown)


ethanjli:
Looks good! Comments from testing, to be addressed in a future PR:

- After the Alarms/Set Value modals, if I press the "Cancel" button, the confirmation dialog pops up (this is good). However, if I press "Cancel" on that confirmation dialog, I should be sent back to the Alarms/Set Value modal, instead of just getting sent directly back to the dashboard.

For records-keeping:

Sudhir-dev:

ethanjli (approved):

PR #281, opened 2021-02-20
Implement unit tests for Protocols/CRCElements

rohanpurohit:
This PR includes:
- Unit test cases for Protocols/CRCElements

Changes from #269 PR :
- Resolved clang-tidy checks
- added Util.h under clang-tidy checks & resolved them
- Updated test case descriptions

### DIFF
--- a/firmware/ventilator-controller-stm32/CMakeLists.txt
+++ b/firmware/ventilator-controller-stm32/CMakeLists.txt
@@ -38,15 +38,45 @@ include_directories(
 )
 
 if ("${CMAKE_BUILD_TYPE}" STREQUAL "TestCatch2")
+    set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/)
+    include(CodeCoverage)
+    append_coverage_compiler_flags()
+
+    # disable all optimization
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0")
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -O0")
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fno-inline -fno-inline-small-functions -fno-default-inline")
+
+    # Some ARM header files give conversion errors from 32-bit to 64-bit
+    # TODO: Need t figure out how to enable this for a small section of code
+    # instead of for the whole project
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fpermissive")
+
+    set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DSTM32H743xx=1")
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -DSTM32H743xx=1")
+
+    setup_target_for_coverage_lcov(
+        NAME ${CMAKE_BUILD_TYPE}_coverage
+        EXECUTABLE ${CMAKE_BUILD_TYPE}
+        EXCLUDE "/usr/include/*" "Core/Inc/catch2/*" "Core/Test/*" "Core/Src/nanopb/*"
+    )
+
     file(
         GLOB_RECURSE LIBRARY_SOURCES
         "Core/Src/Pufferfish/Driver/Indicators/PulseGenerator.cpp"
         "Core/Src/Pufferfish/Driver/Serial/*.cpp"
+        "Core/Src/Pufferfish/HAL/CRC.cpp"
+        "Core/Src/Pufferfish/Application/mcu_pb.c"
+        "Core/Src/nanopb/pb_encode.c"
+        "Core/Src/nanopb/pb_decode.c"
+        "Core/Src/nanopb/pb_common.c"
+        "Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_crc.c"
+        "Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_crc_ex.c"
     )
     add_library(Pufferfish ${LIBRARY_SOURCES})
     file(GLOB_RECURSE EXECUTABLE_SOURCES "Core/Test/*.*")
     add_executable(${CMAKE_BUILD_TYPE} ${EXECUTABLE_SOURCES})
-    target_link_libraries(${CMAKE_BUILD_TYPE} Pufferfish)
+    target_link_libraries(${CMAKE_BUILD_TYPE} Pufferfish gcov)
 else ()
     add_definitions(-DUSE_HAL_DRIVER -DSTM32H743xx -DDEBUG)
 
--- a/firmware/ventilator-controller-stm32/CMakeLists.txt
+++ b/firmware/ventilator-controller-stm32/CMakeLists.txt
@@ -64,9 +64,11 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "TestCatch2")
     file(
         GLOB_RECURSE LIBRARY_SOURCES
         "Core/Src/Pufferfish/Driver/Indicators/PulseGenerator.cpp"
-        "Core/Src/Pufferfish/Driver/Serial/*.cpp"
+        "Core/Src/Pufferfish/Driver/Serial/*.*"
+        "Core/Src/Pufferfish/Application/*.*"
+        "Core/Src/Pufferfish/Util/*.*"
         "Core/Src/Pufferfish/HAL/CRC.cpp"
-        "Core/Src/Pufferfish/Application/mcu_pb.c"
+        "Core/Src/Pufferfish/HAL/Mock/MockTime.cpp"
         "Core/Src/nanopb/pb_encode.c"
         "Core/Src/nanopb/pb_decode.c"
         "Core/Src/nanopb/pb_common.c"
--- a/firmware/ventilator-controller-stm32/CMakeLists.txt
+++ b/firmware/ventilator-controller-stm32/CMakeLists.txt
@@ -47,14 +47,6 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "TestCatch2")
     set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -O0")
     set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fno-inline -fno-inline-small-functions -fno-default-inline")
 
-    # Some ARM header files give conversion errors from 32-bit to 64-bit
-    # TODO: Need t figure out how to enable this for a small section of code
-    # instead of for the whole project
-    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fpermissive")
-
-    set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DSTM32H743xx=1")
-    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -DSTM32H743xx=1")
-
     setup_target_for_coverage_lcov(
         NAME ${CMAKE_BUILD_TYPE}_coverage
         EXECUTABLE ${CMAKE_BUILD_TYPE}
@@ -68,16 +60,16 @@ if ("${CMAKE_BUILD_TYPE}" STREQUAL "TestCatch2")
         "Core/Src/Pufferfish/Application/*.*"
         "Core/Src/Pufferfish/Util/*.*"
         "Core/Src/Pufferfish/HAL/CRC.cpp"
-        "Core/Src/Pufferfish/HAL/Mock/MockTime.cpp"
-        "Core/Src/nanopb/pb_encode.c"
-        "Core/Src/nanopb/pb_decode.c"
-        "Core/Src/nanopb/pb_common.c"
-        "Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_crc.c"
-        "Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_crc_ex.c"
+        "Core/Src/Pufferfish/HAL/Mock/*.cpp"
+        "Core/Src/nanopb/*.c"
     )
     add_library(Pufferfish ${LIBRARY_SOURCES})
+
     file(GLOB_RECURSE EXECUTABLE_SOURCES "Core/Test/*.*")
+
     add_executable(${CMAKE_BUILD_TYPE} ${EXECUTABLE_SOURCES})
+    include_directories("Core/Inc")
+    include_directories("Core/Test/Inc")
     target_link_libraries(${CMAKE_BUILD_TYPE} Pufferfish gcov)
 else ()
     add_definitions(-DUSE_HAL_DRIVER -DSTM32H743xx -DDEBUG)
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.h
@@ -17,14 +17,17 @@ namespace Pufferfish::Driver::Serial::Backend {
 
 struct FrameProps {
   static const size_t payload_max_size = 254;
-  static const size_t chunk_max_size = payload_max_size + 2;  // including delimiter
+  static const size_t chunk_max_size = payload_max_size + 2;    // including delimiter
+  static const size_t encoded_max_size = payload_max_size + 1;  // including cobs
   using ChunkBuffer = Util::ByteVector<chunk_max_size>;
+  using EncodedBuffer = Util::ByteVector<encoded_max_size>;
+  using PayloadBuffer = Util::ByteVector<payload_max_size>;
 
   using InputStatus = Protocols::ChunkInputStatus;
   using OutputStatus = Protocols::ChunkOutputStatus;
 };
 
-using FrameChunkSplitter = Protocols::ChunkSplitter<FrameProps::chunk_max_size>;
+using FrameChunkSplitter = Protocols::ChunkSplitter<FrameProps::encoded_max_size>;
 
 // Decodes frames (length up to 255 bytes, excluding frame delimiter) with COBS
 class COBSDecoder {
@@ -54,7 +57,7 @@ class FrameReceiver {
 
   // Call this until it returns available, then call output
   FrameProps::InputStatus input(uint8_t new_byte);
-  FrameProps::OutputStatus output(FrameProps::ChunkBuffer &output_buffer);
+  FrameProps::OutputStatus output(FrameProps::PayloadBuffer &output_buffer);
 
  private:
   FrameChunkSplitter chunk_splitter_;
@@ -66,7 +69,7 @@ class FrameSender {
   FrameSender() = default;
 
   FrameProps::OutputStatus transform(
-      const FrameProps::ChunkBuffer &input_buffer, FrameProps::ChunkBuffer &output_buffer) const;
+      const FrameProps::PayloadBuffer &input_buffer, FrameProps::ChunkBuffer &output_buffer) const;
 
  private:
   const COBSEncoder cobs_encoder = COBSEncoder();
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
@@ -77,6 +77,10 @@ template <size_t input_size>
 typename CRCElementReceiver<body_max_size>::Status CRCElementReceiver<body_max_size>::transform(
     const Util::ByteVector<input_size> &input_buffer,
     ParsedCRCElement<body_max_size> &output_crcelement) {
+  if (input_buffer.size() > body_max_size) {
+    return Status::invalid_parse;
+  }
+
   if (output_crcelement.parse(input_buffer) != IndexStatus::ok) {
     return Status::invalid_parse;
   }
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
@@ -50,7 +50,8 @@ IndexStatus CRCElement<PayloadBuffer>::parse(const Util::ByteVector<input_size> 
       !std::is_const<PayloadBuffer>::value,
       "Parse method unavailable for CRCElements with const PayloadBuffer type");
 
-  if (input_buffer.size() < CRCElementHeaderProps::header_size) {
+  if (input_buffer.size() < CRCElementHeaderProps::header_size ||
+      input_buffer.size() > (payload_.max_size() + CRCElementHeaderProps::header_size)) {
     return IndexStatus::out_of_bounds;
   }
   Util::read_ntoh(input_buffer.buffer(), crc_);
@@ -77,10 +78,6 @@ template <size_t input_size>
 typename CRCElementReceiver<body_max_size>::Status CRCElementReceiver<body_max_size>::transform(
     const Util::ByteVector<input_size> &input_buffer,
     ParsedCRCElement<body_max_size> &output_crcelement) {
-  if (input_buffer.size() > body_max_size) {
-    return Status::invalid_parse;
-  }
-
   if (output_crcelement.parse(input_buffer) != IndexStatus::ok) {
     return Status::invalid_parse;
   }
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
@@ -49,9 +49,12 @@ IndexStatus CRCElement<PayloadBuffer>::parse(const Util::ByteVector<input_size> 
   static_assert(
       !std::is_const<PayloadBuffer>::value,
       "Parse method unavailable for CRCElements with const PayloadBuffer type");
+  static_assert(
+      Util::ByteVector<input_size>::max_size() <=
+          (PayloadBuffer::max_size() + CRCElementHeaderProps::header_size),
+      "Parse method unavailable as the input buffer size is too large");
 
-  if (input_buffer.size() < CRCElementHeaderProps::header_size ||
-      input_buffer.size() > (payload_.max_size() + CRCElementHeaderProps::header_size)) {
+  if (input_buffer.size() < CRCElementHeaderProps::header_size) {
     return IndexStatus::out_of_bounds;
   }
   Util::read_ntoh(input_buffer.buffer(), crc_);
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
@@ -76,13 +76,13 @@ template <size_t body_max_size>
 template <size_t input_size>
 typename CRCElementReceiver<body_max_size>::Status CRCElementReceiver<body_max_size>::transform(
     const Util::ByteVector<input_size> &input_buffer,
-    ParsedCRCElement<body_max_size> &output_datagram) {
-  if (output_datagram.parse(input_buffer) != IndexStatus::ok) {
+    ParsedCRCElement<body_max_size> &output_crcelement) {
+  if (output_crcelement.parse(input_buffer) != IndexStatus::ok) {
     return Status::invalid_parse;
   }
 
   if (ParsedCRCElement<body_max_size>::compute_body_crc(input_buffer, crc32c_) !=
-      output_datagram.crc()) {
+      output_crcelement.crc()) {
     return Status::invalid_crc;
   }
 
@@ -96,8 +96,8 @@ template <size_t output_size>
 typename CRCElementSender<body_max_size>::Status CRCElementSender<body_max_size>::transform(
     const typename Props::PayloadBuffer &input_payload,
     Util::ByteVector<output_size> &output_buffer) {
-  ConstructedCRCElement<body_max_size> datagram(input_payload);
-  if (datagram.write(output_buffer, crc32c_) != IndexStatus::ok) {
+  ConstructedCRCElement<body_max_size> crcelement(input_payload);
+  if (crcelement.write(output_buffer, crc32c_) != IndexStatus::ok) {
     return Status::invalid_length;
   }
 
--- a/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
+++ b/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
@@ -1,0 +1,100 @@
+/*
+ * Util.h
+ *
+ *  Created on: Nov 17, 2020
+ *      Author: Renji Panicker
+ *
+ *  A set of helper functins for writing and debugging test cases
+ */
+
+#pragma once
+
+#include <string>
+
+#include "Pufferfish/Util/Vector.h"
+
+namespace Pufferfish::Util {
+
+template <size_t payload_size>
+inline bool operator==(
+    const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs) {
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+  for (size_t i = 0; i < lhs.size(); ++i) {
+    if ((uint8_t)lhs[i] != (uint8_t)rhs.at(i)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+template <size_t payload_size>
+inline bool operator!=(
+    const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs) {
+  return !(lhs == rhs);
+}
+
+template <size_t payload_size>
+inline bool operator==(
+    const Pufferfish::Util::ByteVector<payload_size>& lhs,
+    const Pufferfish::Util::ByteVector<payload_size>& rhs) {
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+  for (size_t i = 0; i < lhs.size(); ++i) {
+    if ((uint8_t)lhs[i] != (uint8_t)rhs[i]) {
+      return false;
+    }
+  }
+  return true;
+}
+
+template <size_t payload_size>
+inline bool operator!=(
+    const Pufferfish::Util::ByteVector<payload_size>& lhs,
+    const Pufferfish::Util::ByteVector<payload_size>& rhs) {
+  return !(lhs == rhs);
+}
+
+template <size_t payload_size>
+inline bool convertStringToByteVector(
+    const std::string& input_string, Pufferfish::Util::ByteVector<payload_size>& output_buffer) {
+  if (input_string.size() >= payload_size) {
+    return false;
+  }
+  for (auto& ch : input_string) {
+    output_buffer.push_back(ch);
+  }
+  return true;
+}
+
+template <size_t payload_size>
+inline std::string convertByteVectorToHexString(
+    const Pufferfish::Util::ByteVector<payload_size>& input_buffer) {
+  std::string output_string;
+  for (size_t i = 0; i < input_buffer.size(); ++i) {
+    auto& ch = input_buffer[i];
+    output_string += "\\x";
+    int c1 = ((ch & 0xf0) >> 4);
+    if ((c1 >= 0) && (c1 <= 9)) {
+      output_string += (c1 + '0');
+    } else if ((c1 >= 10) && (c1 <= 15)) {
+      output_string += ((c1 - 10) + 'A');
+    } else {
+      output_string += ('?');
+    }
+
+    c1 = (ch & 0x0f);
+    if ((c1 >= 0) && (c1 <= 9)) {
+      output_string += (c1 + '0');
+    } else if ((c1 >= 10) && (c1 <= 15)) {
+      output_string += ((c1 - 10) + 'A');
+    } else {
+      output_string += ('?');
+    }
+  }
+  return output_string;
+}
+
+}  // namespace Pufferfish::Util
--- a/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
+++ b/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
@@ -4,7 +4,7 @@
  *  Created on: Nov 17, 2020
  *      Author: Renji Panicker
  *
- *  A set of helper functins for writing and debugging test cases
+ *  A set of helper functions for writing and debugging test cases
  */
 
 #pragma once
@@ -15,86 +15,36 @@
 
 namespace Pufferfish::Util {
 
+// operator to match ByteVector and HexString
 template <size_t payload_size>
-inline bool operator==(
-    const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs) {
-  if (lhs.size() != rhs.size()) {
-    return false;
-  }
-  for (size_t i = 0; i < lhs.size(); ++i) {
-    if (static_cast<uint8_t>(lhs[i]) != static_cast<uint8_t>(rhs.at(i))) {
-      return false;
-    }
-  }
-  return true;
-}
+bool operator==(const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs);
 
+// operator to match ByteVector and HexString
 template <size_t payload_size>
-inline bool operator!=(
-    const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs) {
-  return !(lhs == rhs);
-}
+bool operator!=(const Pufferfish::Util::ByteVector<payload_size>& lhs, const std::string& rhs);
 
+// operator to match two ByteVector's
 template <size_t payload_size>
-inline bool operator==(
+bool operator==(
     const Pufferfish::Util::ByteVector<payload_size>& lhs,
-    const Pufferfish::Util::ByteVector<payload_size>& rhs) {
-  if (lhs.size() != rhs.size()) {
-    return false;
-  }
-  for (size_t i = 0; i < lhs.size(); ++i) {
-    if (static_cast<uint8_t>(lhs[i]) != static_cast<uint8_t>(rhs[i])) {
-      return false;
-    }
-  }
-  return true;
-}
+    const Pufferfish::Util::ByteVector<payload_size>& rhs);
 
+// operator to match two ByteVector's
 template <size_t payload_size>
-inline bool operator!=(
+bool operator!=(
     const Pufferfish::Util::ByteVector<payload_size>& lhs,
-    const Pufferfish::Util::ByteVector<payload_size>& rhs) {
-  return !(lhs == rhs);
-}
+    const Pufferfish::Util::ByteVector<payload_size>& rhs);
 
+// converts HexString to ByteVector
 template <size_t payload_size>
-inline bool convert_string_to_byte_vector(
-    const std::string& input_string, Pufferfish::Util::ByteVector<payload_size>& output_buffer) {
-  if (input_string.size() >= payload_size) {
-    return false;
-  }
-  for (const auto& ch : input_string) {
-    output_buffer.push_back(ch);
-  }
-  return true;
-}
+bool convert_string_to_byte_vector(
+    const std::string& input_string, Pufferfish::Util::ByteVector<payload_size>& output_buffer);
 
+// converts ByteVector to HexString
 template <size_t payload_size>
-inline std::string convert_byte_vector_to_hex_string(
-    const Pufferfish::Util::ByteVector<payload_size>& input_buffer) {
-  std::string output_string;
-  for (size_t i = 0; i < input_buffer.size(); ++i) {
-    auto& ch = input_buffer[i];
-    output_string += "\\x";
-    int c1 = ((ch & 0xf0) >> 4);
-    if ((c1 >= 0) && (c1 <= 9)) {
-      output_string += std::to_string(c1);
-    } else if ((c1 >= 10) && (c1 <= 15)) {
-      output_string += static_cast<char>(c1 - 10 + 'A');
-    } else {
-      output_string += ('?');
-    }
-
-    c1 = (ch & 0x0f);
-    if ((c1 >= 0) && (c1 <= 9)) {
-      output_string += std::to_string(c1);
-    } else if ((c1 >= 10) && (c1 <= 15)) {
-      output_string += static_cast<char>(c1 - 10 + 'A');
-    } else {
-      output_string += ('?');
-    }
-  }
-  return output_string;
-}
+std::string convert_byte_vector_to_hex_string(
+    const Pufferfish::Util::ByteVector<payload_size>& input_buffer);
 
 }  // namespace Pufferfish::Util
+
+#include "Util.tpp"
--- a/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
+++ b/firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
@@ -22,7 +22,7 @@ inline bool operator==(
     return false;
   }
   for (size_t i = 0; i < lhs.size(); ++i) {
-    if ((uint8_t)lhs[i] != (uint8_t)rhs.at(i)) {
+    if (static_cast<uint8_t>(lhs[i]) != static_cast<uint8_t>(rhs.at(i))) {
       return false;
     }
   }
@@ -43,7 +43,7 @@ inline bool operator==(
     return false;
   }
   for (size_t i = 0; i < lhs.size(); ++i) {
-    if ((uint8_t)lhs[i] != (uint8_t)rhs[i]) {
+    if (static_cast<uint8_t>(lhs[i]) != static_cast<uint8_t>(rhs[i])) {
       return false;
     }
   }
@@ -58,19 +58,19 @@ inline bool operator!=(
 }
 
 template <size_t payload_size>
-inline bool convertStringToByteVector(
+inline bool convert_string_to_byte_vector(
     const std::string& input_string, Pufferfish::Util::ByteVector<payload_size>& output_buffer) {
   if (input_string.size() >= payload_size) {
     return false;
   }
-  for (auto& ch : input_string) {
+  for (const auto& ch : input_string) {
     output_buffer.push_back(ch);
   }
   return true;
 }
 
 template <size_t payload_size>
-inline std::string convertByteVectorToHexString(
+inline std::string convert_byte_vector_to_hex_string(
     const Pufferfish::Util::ByteVector<payload_size>& input_buffer) {
   std::string output_string;
   for (size_t i = 0; i < input_buffer.size(); ++i) {
@@ -78,18 +78,18 @@ inline std::string convertByteVectorToHexString(
     output_string += "\\x";
     int c1 = ((ch & 0xf0) >> 4);
     if ((c1 >= 0) && (c1 <= 9)) {
-      output_string += (c1 + '0');
+      output_string += std::to_string(c1);
     } else if ((c1 >= 10) && (c1 <= 15)) {
-      output_string += ((c1 - 10) + 'A');
+      output_string += static_cast<char>(c1 - 10 + 'A');
     } else {
       output_string += ('?');
     }
 
     c1 = (ch & 0x0f);
     if ((c1 >= 0) && (c1 <= 9)) {
-      output_string += (c1 + '0');
+      output_string += std::to_string(c1);
     } else if ((c1 >= 10) && (c1 <= 15)) {
-      output_string += ((c1 - 10) + 'A');
+      output_string += static_cast<char>(c1 - 10 + 'A');
     } else {
       output_string += ('?');
     }
--- a/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
+++ b/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
@@ -22,103 +22,85 @@ namespace PF = Pufferfish;
 constexpr size_t buffer_size = 254UL;
 using TestCRCElementProps = PF::Protocols::CRCElementProps<buffer_size>;
 using TestCRCElement = PF::Protocols::CRCElement<TestCRCElementProps::PayloadBuffer>;
+using TestConstructedCRCElement = PF::Protocols::CRCElement<TestCRCElementProps::PayloadBuffer>;
 using TestCRCElementHeaderProps = PF::Protocols::CRCElementHeaderProps;
 
 SCENARIO(
-    "Protocols::CRCElement: The compute_body_crc function correctly computes crc from body",
+    "Protocols::CRCElement: The TestCRCElement::compute_body_crc function correctly computes crc "
+    "from body",
     "[CRCElement]") {
   PF::HAL::SoftCRC32 crc32c{PF::HAL::crc32c_params};
   GIVEN("A CRC element initialized with an empty payload") {
-    TestCRCElementProps::PayloadBuffer input_payload;
+    WHEN("the crc is computed on an input_buffer with empty payload") {
+      PF::Util::ByteVector<buffer_size> buffer;
+      TestCRCElementProps::PayloadBuffer input_payload;
 
-    TestCRCElement crc_element{input_payload};
-
-    WHEN("the crc is computed from the payload") {
-      // Calculated using the Sunshine Online CRC Calculator
-      uint32_t expected_crc = 0x00000000;
-
-      TestCRCElementProps::PayloadBuffer output_payload;
-      output_payload.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
-      output_payload.copy_from(
+      buffer.copy_from(
           input_payload.buffer(), input_payload.size(), TestCRCElementHeaderProps::payload_offset);
 
-      auto computed_crc = TestCRCElement::compute_body_crc(output_payload, crc32c);
+      auto computed_crc = TestCRCElement::compute_body_crc(buffer, crc32c);
 
-      THEN("the computed crc is equal to the expected crc") {
+      THEN("the computed CRC is '0x00000000") {
+        // Calculated using the Sunshine Online CRC Calculator
+        uint32_t expected_crc = 0x00000000;
         REQUIRE(computed_crc == expected_crc);
       }
     }
-  }
 
-  GIVEN("A CRC Element initialized with the payload '123456789'") {
-    TestCRCElementProps::PayloadBuffer input_payload;
-    auto data = std::string("123456789");
-    for (auto& ch : data) {
-      input_payload.push_back(ch);
-    }
+    WHEN("the crc is computed on an input_buffer with payload '123456789'") {
+      TestCRCElementProps::PayloadBuffer input_payload;
+      auto data = std::string("123456789");
+      for (auto& ch : data) {
+        input_payload.push_back(ch);
+      }
 
-    TestCRCElement crc_element{input_payload};
-
-    WHEN("the crc is computed from the payload") {
-      // Calculated using the Sunshine Online CRC Calculator
-      uint32_t expected_crc = 0xe3069283;
-
-      TestCRCElementProps::PayloadBuffer output_payload;
-      output_payload.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
-      output_payload.copy_from(
+      PF::Util::ByteVector<buffer_size> buffer;
+      buffer.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
+      buffer.copy_from(
           input_payload.buffer(), input_payload.size(), TestCRCElementHeaderProps::payload_offset);
 
-      auto computed_crc = TestCRCElement::compute_body_crc(output_payload, crc32c);
+      auto computed_crc = TestCRCElement::compute_body_crc(buffer, crc32c);
 
-      THEN("the computed crc is equal to the expected crc") {
+      THEN("the computed crc is '0xe3069283'") {
+        // Calculated using the Sunshine Online CRC Calculator
+        uint32_t expected_crc = 0xe3069283;
         REQUIRE(computed_crc == expected_crc);
       }
     }
-  }
 
-  GIVEN("A crc element is initalised with payload 0x00") {
-    TestCRCElementProps::PayloadBuffer input_payload;
-    input_payload.push_back(0x00);
+    WHEN("the CRC is computed on an input_buffer with payload '0x00'") {
+      TestCRCElementProps::PayloadBuffer input_payload;
+      input_payload.push_back(0x00);
 
-    TestCRCElement crc_element{input_payload};
-
-    WHEN("the crc is computed from the payload") {
-      // Calculated using the Sunshine Online CRC Calculator
-      uint32_t expected_crc = 0x527d5351;
-
-      TestCRCElementProps::PayloadBuffer output_payload;
-
-      output_payload.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
-      output_payload.copy_from(
+      PF::Util::ByteVector<buffer_size> buffer;
+      buffer.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
+      buffer.copy_from(
           input_payload.buffer(), input_payload.size(), TestCRCElementHeaderProps::payload_offset);
 
-      auto computed_crc = TestCRCElement::compute_body_crc(output_payload, crc32c);
+      auto computed_crc = TestCRCElement::compute_body_crc(buffer, crc32c);
 
-      THEN("the computed crc is equal to the expected crc") {
+      THEN("the computed crc is '0x527d5351'") {
+        // Calculated using the Sunshine Online CRC Calculator
+        uint32_t expected_crc = 0x527d5351;
         REQUIRE(computed_crc == expected_crc);
       }
     }
-  }
 
-  GIVEN("A crc element is initalised with payload 0x01") {
-    TestCRCElementProps::PayloadBuffer input_payload;
-    input_payload.push_back(0x01);
+    WHEN("the crc is computed on an input_buffer with payload '0x01'") {
+      TestCRCElementProps::PayloadBuffer input_payload;
+      input_payload.push_back(0x01);
 
-    TestCRCElement crc_element{input_payload};
+      PF::Util::ByteVector<buffer_size> buffer;
 
-    WHEN("the crc is computed from the payload") {
-      // Calculated using the Sunshine Online CRC Calculator
-      uint32_t expected_crc = 0xa016d052;
-
-      TestCRCElementProps::PayloadBuffer output_payload;
-
-      output_payload.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
-      output_payload.copy_from(
+      buffer.resize(TestCRCElementHeaderProps::header_size + input_payload.size());
+      buffer.copy_from(
           input_payload.buffer(), input_payload.size(), TestCRCElementHeaderProps::payload_offset);
 
-      auto computed_crc = TestCRCElement::compute_body_crc(output_payload, crc32c);
+      auto computed_crc = TestCRCElement::compute_body_crc(buffer, crc32c);
 
-      THEN("the computed crc is equal to the expected crc") {
+      THEN("the computed crc is '0xa016d052") {
+        // Calculated using the Sunshine Online CRC Calculator
+        uint32_t expected_crc = 0xa016d052;
         REQUIRE(computed_crc == expected_crc);
       }
     }
@@ -137,10 +119,13 @@ SCENARIO(
       input_payload.push_back(bytes);
     }
 
-    TestCRCElement crc_element{input_payload};
+    TestConstructedCRCElement crc_element{input_payload};
     WHEN("The crc and payload are written to the output buffer") {
       TestCRCElementProps::PayloadBuffer output_buffer;
-      auto expected = std::string("\xD7\x91\x15\xF6\x01\x02\x05");
+
+      THEN("the crc accessor method returns 0 before the write method is called") {
+        REQUIRE(crc_element.crc() == 0x00000000);
+      }
 
       auto write_status = crc_element.write(output_buffer, crc32c);
       auto crc_compute = crc32c.compute(input_payload.buffer(), input_payload.size());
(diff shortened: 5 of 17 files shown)


firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
+/* compute_body_crc
+Scenario: CRCElement: The compute_body_crc function correctly computes crc from body
```
ethanjli:
```suggestion
Scenario: CRCElement: The compute_body_crc method correctly computes the CRC of a payload from a CRCElement body given as input
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+  GIVEN("A CRC element initialized with an empty payload")
+    WHEN("the crc is computed from the payload")
+      THEN("the computed crc is equal to the expected crc")
```
ethanjli:
```suggestion
      THEN("the computed CRC is 0x00000000")
```
In all THENs in this scenario, we should explicitly list the expected CRC.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
+/* compute_body_crc
+Scenario: CRCElement: The compute_body_crc function correctly computes crc from body
+  GIVEN("A CRC element initialized with an empty payload")
+    WHEN("the crc is computed from the payload")
```
ethanjli:
```suggestion
    WHEN("the CRC is computed on a body with an empty payload")
```
`compute_body_crc` ignores the payload which the CRCElement is initialized with. All WHENs in this scenario should be reworded in a similar way, to be specific that the payload which compute_body_crc cares about is in the input buffer, not in a CRCElement object.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the computed crc is equal to the expected crc")
+
+  GIVEN("A CRC Element initialized with the payload '123456789'")
```
ethanjli: All tests should work on CRCElements initialized with an empty payload, so all GIVENs in this scenario should be GIVEN("A CRC element initialized with an empty payload").

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
```diff
@@ -0,0 +1,537 @@
...
+using TestCRCElementHeaderProps = PF::Protocols::CRCElementHeaderProps;
+
+SCENARIO(
+    "Protocols::CRCElement: The compute_body_crc function correctly computes crc from body",
+    "[CRCElement]") {
+  PF::HAL::SoftCRC32 crc32c{PF::HAL::crc32c_params};
+  GIVEN("A CRC element initialized with an empty payload") {
+    TestCRCElementProps::PayloadBuffer input_payload;
+
+    TestCRCElement crc_element{input_payload};
```
ethanjli: Since `compute_body_crc` is a static method, perhaps these tests should be done without instantiating `CRCElement` objects, and rather just calling `CRCElement::compute_body_crc` directly. This way, we won't need to add THENs asserting that the constructor's payload is unchanged after calling `compute_body_crc`.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+  GIVEN("A crc element is initalised with payload 0x00")
+    WHEN("the crc is computed from the payload")
+      THEN("the computed crc is equal to the expected crc")
+
+  GIVEN("A crc element is initalised with payload 0x01")
+    WHEN("the crc is computed from the payload")
+      THEN("the computed crc is equal to the expected crc")
+
+/* write function
+Scenario: CRCElement: The write function correctly generates body from computed crc and payload
```
ethanjli:
```suggestion
Scenario: ConstructedCRCElement: The write method correctly generates a body from the payload given in the constructor
```
Part of generating the body is computing a CRC on the payload, so we don't need to state it explicitly in the scenario description.
Should we test the `write` method using a `ConstructedCRCElement`? This way we wouldn't need to add THENs asserting that the payload given in the constructor remains unchanged after a `write`, because the compiler enforces constness.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+  GIVEN("A crc element initalised with payload=[0x01, 0x02, 0x05]")
+    WHEN("The crc and payload are written to the output buffer")
+      THEN("crc computed in write method is same as crcchecker compute")
```
ethanjli:
```suggestion
      THEN("After the write method is called, the crc accessor method returns a value equal to the result of directly computing a CRC on the payload.")
```
This also applies to other tests in this scenario.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The output buffer is as expected")
```
ethanjli, line 29: Add a test above this asserting that the header's CRC field has the correct CRC in the correct position
THEN("The CRC field of the body's header matches the value returned by the crc accessor method.")

Also add a test above this asserting that the body has the correct payload in the correct position:
THEN("The body's payload section correctly stores the payload as 0x01 0x02 0x05.")

This also applies to other tests in this scenario.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the computed crc is equal to the expected crc")
+
+  GIVEN("A crc element is initalised with payload 0x01")
+    WHEN("the crc is computed from the payload")
+      THEN("the computed crc is equal to the expected crc")
+
+/* write function
+Scenario: CRCElement: The write function correctly generates body from computed crc and payload
+  GIVEN("A crc element initalised with payload=[0x01, 0x02, 0x05]")
+    WHEN("The crc and payload are written to the output buffer")
```
ethanjli, line 24: Add: THEN("the crc accessor method returns 0 before the write method is called")

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("crc computed in write method is same as crcchecker compute")
+      THEN("The output buffer is as expected")
+
+  GIVEN("A crc element is initalised with an empty payload")
+    WHEN("The crc and payload are written to the output buffer")
+      THEN("The CRC returned by the crc() method is equal to 0")
+      THEN("The output buffer is as expected")
+
+/* parse function
+Scenario: CRCElement: The parse function correctly updates internal crc and payload fields from the input buffer
```
ethanjli:
```suggestion
Scenario: CRCElement: The parse method correctly updates its internal crc value and the constructor's payload buffer from the input buffer
```
The way we retrieve the parsed payload is by accessing the payload buffer we gave to the constructor, so this suggested description doesn't imply that that buffer is internal.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+
+  GIVEN("A CRC Element initialized with an empty payload buffer with a capacity of 254 bytes")
```
ethanjli:
```suggestion
  GIVEN("A CRCElement constructed with an empty payload buffer with a capacity of 254 bytes")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+
+    WHEN("A body with crc more than 4 bytes and equal to the crc of the payload is parsed")
```
ethanjli:
```suggestion
    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is parsed")
```
I don't understand what you're saying, the suggested change is my best interpretation but I might've misinterpreted this.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the parse method reports ok status")
+      THEN("crc returned from the crc() method is same as the crc of the payload in input body")
+      THEN("payload returned from the payload() method is equal to the payload from the body")
+
+    WHEN("A body of less than 4 bytes is parsed")
+      THEN("the parse method reports out of bounds status")
+
+    WHEN("A body with crc and payload is given, where the crc is not equal to the actual crc of the payload")
```
ethanjli:
```suggestion
    WHEN("A body with a complete 4-byte header, and a payload inconsistent with the CRC field of the header, is parsed")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+Scenario: CRCElement: The parse function correctly updates internal crc and payload fields from the input buffer
+
+  GIVEN("A CRC Element initialized with an empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body with crc more than 4 bytes and equal to the crc of the payload is parsed")
+      THEN("the parse method reports ok status")
+      THEN("crc returned from the crc() method is same as the crc of the payload in input body")
+      THEN("payload returned from the payload() method is equal to the payload from the body")
+
+    WHEN("A body of less than 4 bytes is parsed")
```
ethanjli:
```suggestion
    WHEN("A body without a complete 4-byte header is parsed")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+
+/* parse function
+Scenario: CRCElement: The parse function correctly updates internal crc and payload fields from the input buffer
+
+  GIVEN("A CRC Element initialized with an empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body with crc more than 4 bytes and equal to the crc of the payload is parsed")
+      THEN("the parse method reports ok status")
+      THEN("crc returned from the crc() method is same as the crc of the payload in input body")
+      THEN("payload returned from the payload() method is equal to the payload from the body")
```
ethanjli:
```suggestion
      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
```
Add: THEN("the payload given in the CRCElement constructor is independent of the input body"), i.e. when we edit the data inside the input buffer used for the parse method, the data inside the payload buffer does not change.

Add: THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor") - i.e. when we edit the data in the payload buffer given to the constructor, then the data will also be edited in whatever the `payload()` accessor method returns

This also applies to line 48.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The output buffer is as expected")
+
+/* parse function
+Scenario: CRCElement: The parse function correctly updates internal crc and payload fields from the input buffer
+
+  GIVEN("A CRC Element initialized with an empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body with crc more than 4 bytes and equal to the crc of the payload is parsed")
+      THEN("the parse method reports ok status")
+      THEN("crc returned from the crc() method is same as the crc of the payload in input body")
```
ethanjli:
```suggestion
      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
```
This assertion should always be true, whether or not the payload is consistent with the CRC. So we should assert it more directly in this WHEN as well as the other WHENs, rather than relying on the fact that they're all equal in this WHEN.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("payload returned from the payload() method is equal to the payload from the body")
+
+    WHEN("A body of less than 4 bytes is parsed")
+      THEN("the parse method reports out of bounds status")
+
+    WHEN("A body with crc and payload is given, where the crc is not equal to the actual crc of the payload")
+      THEN("the parse method reports ok status")
+      THEN("The crc returned from the crc() method is equal to the crc in the body/input_buffer")
```
ethanjli:
```suggestion
      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
```
This suggested THEN was copied directly from my suggestion on line 38.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The crc returned from the crc() method is not equal to the actual crc for the given payload")
```
ethanjli:
```suggestion
      THEN("the value returned by the crc accessor method is inconsistent with the payload")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The payload returned from the payload() method is equal to the payload from the body")
+
+/* write parse roundtrip
+Scenario: The crc element correctly writes and parses to a buffer
+
+  GIVEN("A CRC Element initialized with a non empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body with crc and payload is generated, parsed and generated again")
+      # write  
+      THEN("The status of write function returns ok")
```
ethanjli:
```suggestion
      THEN("The first write method call returns ok status")
```
We should specify first vs. second for the THENs asserting things about the write method calls.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the parse method reports out of bounds status")
+
+    WHEN("A body with crc and payload is given, where the crc is not equal to the actual crc of the payload")
+      THEN("the parse method reports ok status")
+      THEN("The crc returned from the crc() method is equal to the crc in the body/input_buffer")
+      THEN("The crc returned from the crc() method is not equal to the actual crc for the given payload")
+      THEN("The payload returned from the payload() method is equal to the payload from the body")
+
+/* write parse roundtrip
+Scenario: The crc element correctly writes and parses to a buffer
```
ethanjli:
```suggestion
Scenario: CRCElement correctly preserves data in roundtrip writing and parsing
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The payload of the crc element is as expected")
+      THEN("The output buffer is as expected")
+      # parse
+      THEN("The status of parse function returns ok")
+      THEN("The crc from the CRCElement is equal to the crc in the body")
+      THEN("payload from the CRCElement is equal to the expected payload from the body")
+      # write  
+      THEN("The status of write function returns ok")
+      THEN("The crc from the crc element is equal to expected")
+      THEN("The payload of the crc element is as expected")
```
ethanjli: Comments and suggestions made about THENs for the individual write and parse scenarios above also apply here. In general these THENs should be changed to look as similar as possible to those previous THENs.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("The output buffer is as expected")
+
+
+/* CRCElementReceiver
+Scenario: CRCElementReceiver: The crc element receiver correctly parses datagrams into payloads
```
ethanjli:
```suggestion
Scenario: CRCElementReceiver correctly parses CRCElement bodies and performs consistency checking on them
```

Would you also mind fixing https://github.com/pez-globo/pufferfish-software/blob/eb742d07167caf2d1700923b2b94f5fedc3f71d2/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.h#L85 by renaming the `output_datagram` parameter to `output_crcelement`? It looks like I had also missed a spot when I was copy-pasting things :)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the crc in output crc_element is equal to the given crc in input_buffer")
+
+    WHEN("body with less than 4 bytes is given to the crc element receiver")
+      // body of length 1,2,3 bytes is passed, passing 4 bytes will give a invalid_crc error instead
+      THEN("the transform status is equal to invalid parse") // for each time
+      THEN("The crc returned from the crc() method is equal to 0") // for each time
+
+    WHEN("The CRC stored in the input body is inconsistent with the payload stored in the input body")
+      THEN("the transform method reports invalid crc status")
+      THEN("the crc returned from the crc() method is equal to the invalid crc given in input body")
```
ethanjli: These WHENs and THENs should be updated according to the suggestions made on the write scenario WHENs and THENs above.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+
+/* CRCElementSender
+Scenario: CRCElementSender: The crc element receiver correctly generates datagrams from payload
```
ethanjli:
```suggestion
Scenario: CRCElementSender correctly generates CRCElement bodies from payloads
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+
+  GIVEN("A CRC element sender of capacity 254 bytes")
+    WHEN("A payload is given along with a sufficiently large output buffer")
+      THEN("the transform status is ok")
+      THEN("output buffer is as expected, equal to (crc + input_buffer)")
+
+  GIVEN("A CRC element sender of buffer size 20 bytes")
+    WHEN("The output buffer is not large enough for the input payload")
+      THEN("the transform status is equal to invalid_length")
```
ethanjli, line 218: Is it possible to write a similar test for `CRCElementReceiver`? i.e. where the input buffer is too large for body_max_size

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
...
+      THEN("the transform method reports invalid crc status")
+      THEN("the crc returned from the crc() method is equal to the invalid crc given in input body")
+
+/* CRCElementSender
+Scenario: CRCElementSender: The crc element receiver correctly generates datagrams from payload
+
+  GIVEN("A CRC element sender of capacity 254 bytes")
+    WHEN("A payload is given along with a sufficiently large output buffer")
+      THEN("the transform status is ok")
+      THEN("output buffer is as expected, equal to (crc + input_buffer)")
```
ethanjli, line 211: It would be good to add that the CRC and payload are consistent in either this THEN or a new THEN

firmware/ventilator-controller-stm32/Core/Test/Inc/Pufferfish/Test/Util.h
```diff
@@ -0,0 +1,100 @@
+/*
+ * Util.h
+ *
+ *  Created on: Nov 17, 2020
+ *      Author: Renji Panicker
+ *
+ *  A set of helper functins for writing and debugging test cases
+ */
```
ethanjli, line 8: Do we need all these functions to be marked `inline` and defined in `Util.h` rather than a `Util.cpp`?

ethanjli (commented):
I did a complete review of CRCElements.txt - I won't review CRCElements.cpp for this PR, that will need to be done before we merge any CRC element-related code from the develop branch into the quality (i.e. passed through all quality control processes) branch. I also added comments on some of the suggested changes indicating other changes which will need to be made, so please review them before accepting the suggested changes.

After all these comments are addressed, this PR will be informally approved for merging - you'll need to answer the following questions for records-keeping before formal Github approval:

Your answer to the 3rd question should be yes, because you've added lcov and CodeCoverage.cmake, and your response to that question should list those dependencies/third-party works and any others included in this PR. From preliminary review there shouldn't be any problems with lcov and CodeCoverage.cmake, but I'll need to add licensing information into our internal spreadsheet about anything you list.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
rohanpurohit: I'm indeed using `CRCElement::compute_body_crc` directly as clang-tidy wouldn't pass without that. 
So in these tests what i did was, initialize `CRCElement`  with a certain payload, and copy that payload to the `input_buffer` as `compute_body_crc` requires the `data` to be in the `payload_offset`. 

I will update by removing `CRCElement `objects and without giving payload to the constructor.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,104 @@
+/* compute_body_crc
+Scenario: CRCElement: The compute_body_crc function correctly computes crc from body
+  GIVEN("A CRC element initialized with an empty payload")
+    WHEN("the crc is computed from the payload")
+      THEN("the computed crc is equal to the expected crc")
+
+  GIVEN("A CRC Element initialized with the payload '123456789'")
+    WHEN("the crc is computed from the payload")
```
rohanpurohit:
```suggestion
    WHEN("the crc is computed on an input_buffer with payload '123456789'")
```
Is this better?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: ya my bad , I guess I meant to write a 4 byte CRC , you have rightly interpreted it.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: yep!
also 
https://github.com/pez-globo/pufferfish-software/blob/eb742d07167caf2d1700923b2b94f5fedc3f71d2/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp#L79

and

https://github.com/pez-globo/pufferfish-software/blob/eb742d07167caf2d1700923b2b94f5fedc3f71d2/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp#L99

maybe datagram to crcelement ?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Yes I've written a test for this, so should I then keep it in this PR and fix along with issue 282?

rohanpurohit: Added a temporary change  for this issue in `CRCElementReceiver ` where I'm just returning an `invalid_parse ` 
if `input_buffer ` is greater than `body_max_size `, 

And for the write method I think `CRCElement::write_protected` does a resize instead to handle this issue.

Then maybe we should follow the same approach for `parse ` as well? 
If it's better to change `Vector::copy_from ` to return error codes then should it be in this PR?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: using this https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r581883306
will change if necessary.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: > Should we test the write method using a ConstructedCRCElement? This way we wouldn't need to add THENs asserting that the payload given in the constructor remains unchanged after a write, because the compiler enforces constness.

Yes, updated the tests for  `write` by using ConstructedCRCElement.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: current change: 7504e9a1da4104363ef9eb80fa03677a6af17afc

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added  THEN("The CRC field of the output buffer's header is consistent with the input_payload.")

rohanpurohit:
Added all the suggested changes. for issue 282 added comments here https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r580799264

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/CRCElements.cpp
ethanjli: Let's change the GIVEN to "The CRCElement::compute_body_crc static method"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: We'll include fixes for CRCElement and Datagram in this PR, and we'll reserve changes to `Vector` for a future PR

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: We'll say "an input_buffer body", but it looks good

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+
+    WHEN("the CRC is computed on an input_buffer with payload '0x00' ")
+      THEN("the computed crc is '0x527d5351'")
+
+    WHEN("the crc is computed on an input_buffer with payload '0x01' ")
+      THEN("the computed crc is '0xa016d052'")
+
+/* write function
+Scenario: ConstructedCRCElement: The write method correctly generates a body from the payload given in the constructor
+  GIVEN("A crc element initalised with payload=[0x01, 0x02, 0x05]")
```
ethanjli: I mentioned this in a previous comment but forgot to check whether it applied to other tests, but it does: let's use "constructed" instead of "initialized" when we refer to inputs given into the constructor. There might rarely be some tests of stateful classes where we use both, e.g. "GIVEN a StateSynchronizer constructed with ____ schedules/etc. and initialized to be on the 1st schedule entry"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+    WHEN("The crc and payload are written to the output buffer")
+      THEN("the crc accessor method returns 0 before the write method is called")
+      THEN("After the write method is called, the crc accessor method returns a value equal to the result of directly computing a CRC on the payload.")
+      THEN("The CRC field of the body's header matches the value returned by the crc accessor method.")
+      THEN("The body's payload section correctly stores the payload as 0x01 0x02 0x05")
+      THEN("The output buffer is as expected")
+
+  GIVEN("A crc element is initalised with an empty payload")
+    WHEN("The crc and payload are written to the output buffer")
+      THEN("the crc accessor method returns 0 before the write method is called")
```
ethanjli: I think we need a `THEN("After the write method is called, the crc accessor method returns...")` - I think it returns 0?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+      THEN("The CRC field of the body's header matches the value returned by the crc accessor method.")
+      THEN("The body's payload section is empty")
+      THEN("The output buffer is as expected")
+
+/* parse function
+Scenario: CRCElement: The parse method correctly updates its internal crc value and the constructor's payload buffer from the input buffer
+  GIVEN("A CRCElement constructed with an empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body without a complete 4-byte header is parsed")
+      THEN("the parse method reports out of bounds status")
```
ethanjli, line 52: I think we should also add a `THEN("the constructor's payload buffer remains unchanged")`

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload given in the CRCElement constructor is independent of the input body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor")
+
+  GIVEN("A CRCElement constructed with a non-empty payload buffer of capacity 254 bytes")
+    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is parsed")
+      THEN("the crc accessor method returns 0 before the parse method is called")
+      THEN("the payload returned from the payload accessor method is equal to the payload buffer given in the CRCElement constructor")
+      THEN("the parse method reports ok status")
+      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
```
ethanjli: This THEN should be specified as applying after the parse method is called.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload given in the CRCElement constructor is independent of the input body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor")
+
+/* write parse roundtrip
+Scenario: CRCElement correctly preserves data in roundtrip writing and parsing
+
+  GIVEN("A CRC Element initialized with a non empty payload buffer with a capacity of 254 bytes")
+
+    WHEN("A body with crc and payload is generated, parsed and generated again")
```
ethanjli: I thought about this a bit more and I think the test would be stronger if we didn't reuse the same CRCElement object (and the same payload buffer) for all three write/parse/write method calls, but instead instantiated a new ParsedCRCElement/ConstructedCRCElement object (and a new payload buffer, for ParsedCRCElement objects) for each method call. This way we can ensure that roundtripping works correctly even if no internal state is preserved across those method calls, which is closer to how the objects would be instantiated and used in the application.

Then we could update the WHEN as: WHEN A body with CRC and payload is generated, parsed into a new ParsedCRCElement object, and generated again with a new ConstructedCRCElement object

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is given to the receiver")
+      THEN("the transform method reports ok status")
+      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
+      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload buffer updated in the CRCElement is independent of the input body")
+
+    WHEN("A body without a complete 4-byte header and empty payload is given to the crc element receiver")
+      THEN("the transform status is equal to invalid parse")
+      THEN("the value returned by the crc accessor method is equal to 0x00000000r")
+      THEN("The body's payload section is empty")
```
ethanjli: This doesn't seem correct to me - the body is exactly the contents of input_buffer, which is const. So the payload section of the body is preserved at what it was. Separately, the value returned by the payload accessor method of output_crcelement will be empty if (and only if) output_crcelement had an empty payload before the CRCElementReceiver::transform was called.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+
+    WHEN("body with less than 4 bytes is given to the crc element receiver")
+      // body of length 1,2,3 bytes is passed, passing 4 bytes will give a invalid_crc error instead
+      THEN("the transform status is equal to invalid parse") // for each time
+      THEN("the value returned by the crc accessor method is equal to 0x00000000") // for each time
```
ethanjli: I thought about this a bit more and this is not strictly correct (also applies to the same THEN in the previous WHEN). Actually the values returned by the crc and payload methods should just be whatever they were before. For example, if we passed in an output_crcelement with a nonzero crc and a payload with some bytes inside, those should be preserved after transform returns invalid_parse. We should cover this in a test, and I think the WHEN will need to specify what (if anything) was the state of output_crcelement before calling transform.

This test should also specify that the crc accessor method is called on the output_crcelement, not on the CRCElementReceiver object.
This test should also have a THEN checking the value returned by the payload accessor method of output_Crcelement.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+Scenario: CRCElementReceiver correctly parses CRCElement bodies and performs consistency checking on them
+
+  GIVEN("A CRC element receiver of capacity 254 bytes")
+    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is given to the receiver")
+      THEN("the transform method reports ok status")
+      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
+      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload buffer updated in the CRCElement is independent of the input body")
+
+    WHEN("A body without a complete 4-byte header and empty payload is given to the crc element receiver")
```
ethanjli: I don't understand how this WHEN is different from the next WHEN - I think one of these was the original one and the other was added in response to a suggested revision. I would go with something like "WHEN a body with an incomplete header and an empty payload is..." or "WHEN A body without a complete 4-byte header is..." (the latter is a direct copy from tests above, but maybe the tests above should be changed to be more like the former). The current wording, "WHEN A body without a complete 4-byte header and empty payload is...", is a bit ambiguous and could be misinterpreted as meaning "WHEN A body with an incomplete 4-byte header and a non-empty payload is..." (this is the `!A && B` vs. `!(A && B)` problem)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+      THEN("The body's payload section is empty")
+
+    WHEN("body with less than 4 bytes is given to the crc element receiver")
+      // body of length 1,2,3 bytes is passed, passing 4 bytes will give a invalid_crc error instead
+      THEN("the transform status is equal to invalid parse") // for each time
+      THEN("the value returned by the crc accessor method is equal to 0x00000000") // for each time
+
+    WHEN("A body with a complete 4-byte header, and a payload inconsistent with the CRC field of the header, is given to the receiver")
+      THEN("the transform method reports invalid crc status")
+      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
```
ethanjli: This test should also have a THEN checking the value returned by the payload accessor method of output_crcelement.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,131 @@
...
+
+  GIVEN("A CRC element receiver of capacity 10 bytes") {
+
+    WHEN("A body of size 20 bytes with a complete 4-byte header, And a payload of capacity 16 bytes completely filled with data is given to the receiver")
+      THEN("the transform method reports invalid parse status")
```
ethanjli, line 192: This test should also have THENs checking the values returned by the crc and payload accessor methods of output_crcelement.

ethanjli (commented):
I did another pass of review on the changed test definitions and found some small details to fix. Besides the remaining changes, I think everything else is ready for merging.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Sorry about the ambiguity, so in this test case I'm basically using an empty `input_buffer`, in the latter test we are using a body with crc header less than 4 bytes, so 1,2 and 3 respectively so there it's the case of `incomplete header`, here there's nothing.
so maybe the WHEN should be something like WHEN("An empty input_buffer is given to the crc element receiver")

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Yes i'm not sure how this got here, I'm actually not checking what payload accessor returns for this case as it's empty also the `input_buffer` is empty to begin with.

Secondly in all CRCElementReceiver test cases I'm indeed using an empty payload buffer for the output_crcelement. 
So should I get rid of this THEN completely or should I do something like
THEN("The payload buffer returned by the payload accessor method is empty") ?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: yes returns 0, added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Fixed across test cases. also updated for receiver as `After the transform method is called`

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Updated using ParsedCRCElement & ConstructedCRCElement, using different payload buffers for constructing the elements and different buffers for input.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Correct! so to give some context, for all the tests in CRCElementReceiver I'm using a `output_crcelement` with empty payload given in the constructor, I did that mainly as we are using it as an output parameter and if parse is succesful it will get updated.

> This test should also specify that the crc accessor method is called on the output_crcelement,

will update this for all cases.

So I was thinking for these I'll divide into two GIVEN's , the ones that are already there with
  GIVEN("A CRC element receiver of capacity 254 bytes and output_crcelement constructed with empty payload")
and
  GIVEN("A CRC element receiver of capacity 254 bytes and output_crcelement constructed with payload '0x12 0x13 0x14 0x15 0x16'")

> . For example, if we passed in an output_crcelement with a nonzero crc and a payload with some bytes inside,

Also even with a output_crcelement with some bytes inside, `crc_ ` member parameter is only updated when `write` method is called, so if we want a test that has crcelement with some crc then we must do that.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r584449591

So when the GIVEN is a empty payload in output_crcelement, I'll say "returns empty payload buffer"
and for a GIVEN with payload in output_crcelement "returns 'payload'"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: My preference would be something like a combination of the following:
- THEN("after the transform method is called, the payload given in the ParsedCRCElement constructor remains unchanged") // i.e. if we inserted some dummy data into the buffer before giving it to the constructor, the data stays there; and if we give make the buffer empty before giving it to the constructor, the buffer stays empty
- THEN("the payload returned from the payload accessor method is equal to the payload from the body")

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: Yes, I agree with the approach of into two GIVENs - I would just specify that the output_crcelement is constructed with an empty payload _buffer_, since the way we're using it in this test is as a place to dump data rather than as the input data itself. In both GIVENs, the THEN could just say that output_crcelement's payload buffer remains unchanged, and the only difference is in the test code assertions.

For the GIVEN with a nonzero CRC, yes part of the test setup code would involve calling `output_crcelement.write` and throwing away the output buffer, in order to make the `crc_` private member become nonzero, in order to set up the internal state of that object for the test.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: Yes, I like that - good idea to test an empty input as distinct from a non-empty too-small input

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: I think this will be clearest if we split it into two THENs, as in https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r585177976 - so here I think we'd assert that the contents of the payload buffer which had been given to the constructor are set to the bytes from the payload section of the input body, and then we'd assert that the payload accessor method returns a reference to that payload buffer in the constructor

firmware/ventilator-controller-stm32/CMakeLists.txt
```diff
@@ -38,15 +38,39 @@ include_directories(
...
 
 if ("${CMAKE_BUILD_TYPE}" STREQUAL "TestCatch2")
+    set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/)
+    include(CodeCoverage)
+    append_coverage_compiler_flags()
+
+    # disable all optimization
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0")
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -O0")
+    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -fno-inline -fno-inline-small-functions -fno-default-inline")
```
ethanjli, line 48: How important are these flags? Are they needed for code coverage? If they're only needed for code coverage, how easy would it be to make a separate build target for running tests vs. for checking test coverage?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
```diff
@@ -50,7 +50,8 @@ IndexStatus CRCElement<PayloadBuffer>::parse(const Util::ByteVector<input_size>
       !std::is_const<PayloadBuffer>::value,
       "Parse method unavailable for CRCElements with const PayloadBuffer type");
 
-  if (input_buffer.size() < CRCElementHeaderProps::header_size) {
+  if (input_buffer.size() < CRCElementHeaderProps::header_size ||
+      input_buffer.size() > (payload_.max_size() + CRCElementHeaderProps::header_size)) {
```
ethanjli: Let's see if we can do a static_assert on input_buffer.max_size() vs. payload_.max_size() + CRCElementHeaderProps::header_size for checking whether the input_buffer is small enough. This way we don't need to define a test checking the sizes of the buffers, we just need to describe in the documentation for the method.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: If we can establish the following in the CRCElement tests:

THEN the bytes in the constructor's payload buffer look like "..."
THEN the payload accessor method returns the constructor's payload buffer

is equivalent to:

THEN the CRCElement contains the payload "..."

then in the CRCElementReceiver tests we can just assert THEN the CRCElement contains the payload "..."

firmware/ventilator-controller-stm32/CMakeLists.txt
ethanjli: Tracked in #284

ethanjli (commented):
Here are the comments discussed during the call today.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
rohanpurohit: For the current condition , we need to use `payload_.max_size()` , So if we need to access `payload_` that means we are accessing `this` which is not a compile time constant and static_assert will fail,
we could make max_size() method static to solve aforementioned issue. what do you suggest we do @ethanjli ?

places where max_size() method is used: 
- https://github.com/pez-globo/pufferfish-software/blob/8c73b599d3da4b3df112f22a7685385bec57327c/firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/FDO2/Commands.cpp#L327 (This fails clang-tidy check, if we make it static)
- https://github.com/pez-globo/pufferfish-software/blob/8c73b599d3da4b3df112f22a7685385bec57327c/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp#L26

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added in CRCElement following https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r585243812
https://github.com/pez-globo/pufferfish-software/pull/281/files#diff-10ffb7b9ec39ab80c47393699e09e302b3ba3f871bf74fb4f834bf350cc97ef0R65

in Receiver https://github.com/pez-globo/pufferfish-software/pull/281/files#diff-10ffb7b9ec39ab80c47393699e09e302b3ba3f871bf74fb4f834bf350cc97ef0R132

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Added.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
ethanjli: Ah, good catch. If we make `Vector::max_size` be static, can we use `PayloadBuffer::max_size()` instead of `payload_.max_size()` and `input_size` or `Util::ByteVector<input_size>::max_size()` instead of `input_buffer.max_size()`? If so, let's do that.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: Pattern to use for all CRCElementReceiver tests (and CRCElement tests in a similar pattern):

```
GIVEN ...
  WHEN ...
    THEN the transform method returns ... status code
    THEN the ParsedCRCElement output parameter has the following properties:
      - Its crc accessor method returns ...
      - The payload given to its constructor is ...
      - Its payload accessor method returns a reference to the payload given in its constructor
      - The data in its payload is independent of the data in input_buffer
```

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
ethanjli: Changes to make:
FrameProps: using encoded_max_size = payload_max_size + 1
FrameProps: using EncodedBuffer = Util::ByteVector<encoded_max_size>
FrameProps: using PayloadBuffer = Util::ByteVector<payload_max_size>
Frames.h: using FrameChunkSplitter = Protocols::ChunkSplitter<FrameProps::encoded_max_size>;
FrameReceiver::output: FrameProps::EncodedBuffer temp_buffer;
FrameReceiver::output: FrameProps::PayloadBuffer &output_buffer
FrameSender::transform: const FrameProps::PayloadBuffer &input_buffer, FrameProps::ChunkBuffer &output_buffer
BackendReceiver::output: FrameProps::PayloadBuffer temp_buffer1;

ethanjli: With these changes, I'll test on the Nucleo board before final approval.

rohanpurohit: > using FrameChunkSplitter = Protocols::ChunkSplitterFrameProps::encoded_max_size;

I was wondering by not using chunk_max_size are we enforcing the `delimiter `to be in the `254 bytes` itself and not `254 non-delimiter bytes+ delimiter byte + ecnoded byte`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.h
```diff
@@ -17,14 +17,17 @@ namespace Pufferfish::Driver::Serial::Backend {
 
 struct FrameProps {
   static const size_t payload_max_size = 254;
-  static const size_t chunk_max_size = payload_max_size + 2;  // including delimiter
+  static const size_t chunk_max_size = payload_max_size + 2;    // including delimiter
+  static const size_t encoded_max_size = payload_max_size + 1;  // including cobs
   using ChunkBuffer = Util::ByteVector<chunk_max_size>;
+  using EncodedBuffer = Util::ByteVector<encoded_max_size>;
```
rohanpurohit, line 23: maybe we should be reusing `ChunkBuffer ` here instead of using a different size for `EncodedBuffer`, I maybe wrong but I think we are missing the case where include_delimiter is true, https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r588073273

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
rohanpurohit: Done with this change https://github.com/pez-globo/pufferfish-software/pull/281/commits/f947b80f5794f159f569382d81704313f06492e3

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
rohanpurohit: commit for the changes discussed: https://github.com/pez-globo/pufferfish-software/pull/281/commits/d8ffd6b5d066fe8fbe8b278cfa3c8958eba843cc

rohanpurohit:
- Added static assert
- Made changes to frames, backend to fix static assert
- Updated test cases for CRCElementReceiver as suggested.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: It looks like this change got reverted somehow?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the value returned by the crc accessor method is inconsistent with the payload")
+      THEN("the payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor")
+      THEN("the payload given in the CRCElement constructor is independent of the input buffer body")
+
+  GIVEN("A CRCElement constructed with a non-empty payload buffer of capacity 254 bytes")
+
+    WHEN("A body without a complete 4-byte header is parsed")
+      THEN("the parse method reports out of bounds status")
+      THEN("the value returned by the crc accessor method is equal to the crc field of the input body's header")
```
ethanjli, line 67: Actually, the CRC remains at whatever value it was given before parse was called.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("The payload buffer given to it's constructor is '0x12 0x13 0x14 0x15 0x16' ")
```
ethanjli, line 68: I think it would be clearer if this were `THEN("The payload buffer given to its constructor remains unchanged")`

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
ethanjli: I think the CRC remains at whatever value it had before parse was called.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor")
+      THEN("the payload given in the CRCElement constructor is independent of the input buffer body")      
+  
+    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is parsed")
+      THEN("the crc accessor method returns 0 before the parse method is called")
+      THEN("The bytes in the constructor's payload buffer are '0x12 0x13 0x14 0x15 0x16' ")
+      THEN("The payload accesor method returns the constructor's payload buffer")
```
ethanjli: I think either these three THENs should also be in the previous WHEN for consistency, or they should be omitted because they can just be REQUIREs under the WHEN, but not explicitly THENs.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("The output buffer is as expected")
+
+/* CRCElementReceiver
+Scenario: CRCElementReceiver correctly parses CRCElement bodies and performs consistency checking on them
+
+  GIVEN("A CRC element receiver of capacity 254 bytes and output_crcelement constructed with an empty payload buffer")
+
+    WHEN("A body with a complete 4-byte header, and a payload consistent with the CRC field of the header, is given to the receiver")
+      THEN("the transform method reports ok status")
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
```
ethanjli: I didn't mean for this to literally be its own THEN - after all, there can't be any REQUIREs associated specifically with it. Rather, it was meant so that the following THENs would substitute "its" with "output_crcelement's":
```
      THEN("the value returned by output_crcelement's crc accessor method is equal to the crc field of the input body's header")
      THEN("the payload given to output_crcelement's constructor is empty")
      THEN("output_crcelement's payload accessor method returns a reference to the payload given in its constructor")
      THEN("The data in output_crcelement's payload is independent of the data in input_buffer")
```
(this also applies to the other tests in this file)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the value returned by it's crc accessor method is equal to the crc field of the input body's header")
+      THEN("the payload given to it's constructor is empty")
```
ethanjli: The CRCElementReceiver successfully the payload from the body. Is the payload supposed to still be empty?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("Its payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in its payload is independent of the data in input_buffer")
+
+    WHEN("An empty input_buffer is given to the crc element receiver")
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("the value returned by it's crc accessor method is equal to the crc field of the input body's header")
```
ethanjli: The input body is empty, so this THEN doesn't make sense to me

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the payload given to it's constructor is empty")
+      THEN("Its payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in its payload is independent of the data in input_buffer")
+
+    WHEN("body with less than 4 bytes is given to the crc element receiver")
+      // body of length 1,2,3 bytes is passed, passing 4 bytes will give a invalid_crc error instead
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("the value returned by it's crc accessor method is equal to the crc field of the input body's header")
```
ethanjli: I'm pretty sure the CRC accessor method returns the same value as whatever it would've returned before calling the transform method, because the parse returned early with out_of_bounds.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the payload given to it's constructor is empty")
+      THEN("Its payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in its payload is independent of the data in input_buffer")
+
+    WHEN("A body with a complete 4-byte header, and a payload inconsistent with the CRC field of the header, is given to the receiver")
+      THEN("the transform method reports invalid crc status")
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("the value returned by it's crc accessor method is equal to the crc field of the input body's header")
+      THEN("the payload given to it's constructor is empty")
```
ethanjli: The CRCElementReceiver successfully the payload from the body. Is the payload supposed to still be empty?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("the value returned by it's crc accessor method is equal to the crc field of the input body's header")
+      THEN("the payload given to it's constructor is '0x01 0x05 0x12 0x13 0x14 0x15 0x16'")
+      THEN("Its payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in its payload is independent of the data in input_buffer")
+  
+    WHEN("An empty input_buffer is given to the crc element receiver")
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("it's crc accessor method returns 0")
```
ethanjli: This THEN would be clearer if you specified in the WHEN what the CRC member was before it was given to the CRCElementReceiver (you do this in the next WHEN)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("the payload given to it's constructor is '0x01 0x05 0x12 0x13 0x14 0x15 0x16'")
```
ethanjli: I think it would be clearer This THEN would be clearer if you specified in the WHEN what the payload was before it was given to the CRCElementReceiver (I would suggest doing this in the next WHEN too)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("Its payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in its payload is independent of the data in input_buffer")
+
+    WHEN("An empty input_buffer is given to the crc element receiver and output_crcelement with a non-zero crc member")
```
ethanjli: Did you mean: WHEN("An empty input_buffer and an output_crcelement with a non-zero crc member are given to the crc element receiver")

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("The write method of output_crcelement reports ok status")
+      THEN("The value returned by the crc accessor method of output_crcelement is '0x40D06D79'")
+      THEN("the transform status is equal to invalid parse")
```
ethanjli: My comment on lines 73-75 applies here, too

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,186 @@
...
+      THEN("After the transform method is called, The ParsedCRCElement output parameter has the following properties : ")
+      THEN("the value returned by it's crc accessor method is equal to '0x40D06D79'")
+      THEN("the payload given to it's constructor is '0x01 0x05 0x12 0x13 0x14 0x15 0x16'")
```
ethanjli: Are these values coming from the input_buffer after a successful parse, or are they the data which was already in output_crcelement and which was left unchanged after an unsuccessful parse? The CRC value appears to be the same before and after transform, while the payload is left unspecified before transform.

I think it'll be clearer and more maintainable if you use the following structure for THENs in this file (don't copy these literally, adapt the language and assertions to make sense for each specific test):
- THEN the value returned by output_crcelement's crc accessor method remains the same as before the transform method call
- THEN the value returned by output_crcelement's crc accessor method becomes equal to the crc field of input_buffer's header
- THEN the value returned by output_crcelement's payload accessor method remains the same as before the transform method call
- THEN the value returned by output_crcelement's payload accessor method becomes equal to the payload of input_buffer

ethanjli (commented):
I went through and identified several regressions, errors, and/or ambiguities introduced with the latest round of changes after my last comments.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/CRCElements.tpp
ethanjli: It seems to run fine on the Nucleo board!

rohanpurohit:
- Updated Test descriptions as suggested.
- added REQUIRE'S in GIVEN'S wherever needed

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+  GIVEN("A CRCElement constructed with a non-empty payload buffer of capacity 254 bytes")
+    // assertions carried out in GIVEN:
+    // The crc accessor method returns 0
+    // The payload buffer returned by the payload accessor method is '0x12 0x13 0x14 0x15 0x16'
+
+    WHEN("A body without a complete 4-byte header is parsed")
+      THEN("the parse method reports out of bounds status")
+      THEN("After the parse method is called, the value returned by the crc accessor method remains unchanged")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the CRCElement constructor")
+      THEN("the payload given in the CRCElement constructor is independent of the input buffer body")
```
ethanjli, line 80: What should the contents of the payload buffer be after parse method is called? (I think the contents should be unchanged)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+
+  GIVEN("A CRC element receiver of capacity 254 bytes and an output_crcelement constructed with an empty payload buffer")
+    // assertions carried out in GIVEN:
+    // The output_crcelement's crc accessor method returns 0
+    // The payload buffer returned by the output_crcelement's payload accessor method is empty
+
+    WHEN("An empty input_buffer is given to the crc element receiver")
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, the value returned by the output_crcelement's crc accessor method remains unchanged")
+      THEN("The output_crcelement's payload accessor method returns a reference to the payload given in its constructor")
```
ethanjli, line 143: What should the contents of the payload buffer be after parse method is called? (I think the contents should be unchanged)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+      THEN("The data in the output_crcelement's payload buffer is independent of the data in input_buffer")
+
+    WHEN("A body with less than 4 bytes is given to the crc element receiver")
+      // body of length 1,2,3 bytes is passed, passing 4 bytes will give a invalid_crc error instead
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, the value returned by the output_crcelement's crc accessor method remains unchanged")
+      THEN("The output_crcelement's payload accessor method returns a reference to the payload given in its constructor")
```
ethanjli, line 151: What should the contents of the payload buffer be after parse method is called? (I think the contents should be unchanged)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+
+  GIVEN("A CRC element receiver of capacity 254 bytes and an output_crcelement constructed with an non-empty payload buffer")
+    // assertions carried out in GIVEN:
+    // The output_crcelement's crc accessor method returns 0
+    // The payload buffer returned by the output_crcelement's payload accessor method is '0x1e 0xe2 0x9a 0xf8 0x4f'
+
+    WHEN("An empty input_buffer is given to the crc element receiver")
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, the value returned by the output_crcelement's crc accessor method remains unchanged")
+      THEN("The output_crcelement's payload accessor method returns a reference to the payload given in its constructor")
```
ethanjli, line 176: What should the contents of the payload buffer be after parse method is called? (I think the contents should be unchanged)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+
+  GIVEN("A CRC element receiver of capacity 254 bytes, and output_crcelement constructed with an non-empty payload buffer and non-zero crc member")
+    // assertions carried out in GIVEN:
+    // The value returned by the output_crcelement's crc accessor method is '0x40D06D79'
+    // The payload buffer returned by the output_crcelement's payload accessor method is '0x23 0x84 0xf5 0x6a 0xac 0xae'
+
+    WHEN("An empty input_buffer is given to the crc element receiver")
+      THEN("the transform method reports invalid parse status")
+      THEN("After the transform method is called, the value returned by the output_crcelement's crc accessor method remains unchanged")
+      THEN("The output_crcelement's payload accessor method returns a reference to the payload given in its constructor")
```
ethanjli, line 194: What should the contents of the payload buffer be after parse method is called? (I think the contents should be unchanged)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/crcelements.txt
```diff
@@ -0,0 +1,212 @@
...
+
+    WHEN("A payload is given along with a sufficiently large output buffer")
+      THEN("the transform method reports ok status")
+      THEN("output buffer is as expected, equal to (crc + input_buffer)")
+      THEN("The CRC field of the output buffer's header is consistent with the input_payload.")
+
+  GIVEN("A CRC element sender of buffer size 20 bytes")
+
+    WHEN("The output buffer is not large enough for the input payload")
+      THEN("the transform status is equal to invalid_length")
```
ethanjli, line 218: What should the contents of output_buffer be after the transform method is called? (I think the contents should be unchanged, whether output_buffer is empty or not)

ethanjli (commented):
Everything looks good, the only suggested changes remaining are to add some missing THENs (need to assert that the contents of the payload buffer remain unchanged when a parse fails), e.g. `THEN("the constructor's payload buffer remains unchanged")` or something similar (this example was taken verbatim from line 54/line 97)

rohanpurohit:
Yes,
- codecoverage.cmake
- lcov

ethanjli (approved):

Issue #124, opened 2020-09-16
Waveform data gets discarded during frontend freezes


ethanjli:
When the frontend freezes, it misses some data and that data gets discarded - in earlier versions, this would cause missing data in the waveforms, but now we reset the waveforms as a workaround. It would be better if we didn't discard the data when the frontend freezes.

PR #283, opened 2021-02-24
Add Rotary encoder support in Clicker  & PR comments resolved

### LINKED ISSUE #69: Buggy behavior in the Set Alarm Limits modal dialog
**Describe the bug**
When I open one of the set alarm modal dialogs and leave the upper limit at 100 and start moving the lower slider from 0 to 5, I notice that the order of the spinners (where numbers and the arrow buttons are) switches. I only get this behavior when I drag the slider, not when I press the arrow buttons in the spinners.

**To Reproduce**
1. Start the frontend
2. Go to the dashboard
3. Click on the "Alarm" button of the PIP box. The dialog box currently starts with "0" on top and "100" on bottom (though it should actually start with "100" on top and "0" on bottom)
4. Drag the lower limit slider from 0 to 2. Now the alarm limits have switched so that "100" is on top and "2" is on bottom.
5. Press the up arrow button for the "2" value. This changes the upper number to 3 and the lower number to 4, and it also moves the slider's limits.

**Expected behavior**
The number on top should always correspond to the upper limit, and its up and down buttons should always adjust that value. The number on bottom should always correspond to the lower limit, and its up and down buttons should always adjust that value. These invariants should hold true no matter how the slider is dragged.

The slider might be more intuitive (and it might reduce the complexity of the code involved in tracking which number is the upper limit vs. which is the lower limit) if the upper limit is not allowed to be dragged to the left of the lower limit, and if the lower limit is not allowed to be dragged to the right of the upper limit. This might also make things simpler on the "Set Alarms" screen after we need to add increment/decrement buttons. However, I am not sure how easy it would be to enforce this constraint using Material UI's slider component(s).

Sudhir-dev:
This PR covers following points

* Added Rotary encoder support for all Clicker functions in Alarms page, Alarm Modal, Settings page & Quickstart page
* Added border around clicker to identify which value clicker is active
* Event Log settings popup updated to Multistep popup
* Added HR alarms in protobuf & Events log
* Start/Pause Ventilation is disabled when backend is disconnected
* Multistep popup on Cancel of Confirm/Cancel confirmation popup, reopens multistep popup with previous state
* Added Clicker in Alarms page & updated Slider CSS for disabled mode

### DIFF
--- a/frontend/src/modules/alarms/AlarmsPage.tsx
+++ b/frontend/src/modules/alarms/AlarmsPage.tsx
@@ -1,7 +1,7 @@
 import { Button, Grid, Typography } from '@material-ui/core';
-import { makeStyles, Theme } from '@material-ui/core/styles';
+import { makeStyles, Theme, useTheme } from '@material-ui/core/styles';
 import Pagination from '@material-ui/lab/Pagination';
-import React, { useEffect, useState } from 'react';
+import React, { RefObject, useEffect, useRef, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { updateCommittedState } from '../../store/controller/actions';
 import { AlarmLimitsRequest, VentilationMode, Range } from '../../store/controller/proto/mcu_pb';
@@ -10,8 +10,11 @@ import {
   getParametersRequestMode,
 } from '../../store/controller/selectors';
 import { ALARM_LIMITS, ALARM_LIMITS_STANDBY } from '../../store/controller/types';
+import { setActiveRotaryReference } from '../app/Service';
+import { ValueClicker } from '../controllers';
 import ValueSlider from '../controllers/ValueSlider';
 import ModeBanner from '../displays/ModeBanner';
+import useRotaryReference from '../utils/useRotaryReference';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -56,6 +59,11 @@ interface AlarmProps {
   setAlarmLimits(alarmLimits: Partial<AlarmLimitsRequest>): void;
 }
 
+enum SliderType {
+  LOWER,
+  UPPER,
+}
+
 const Alarm = ({
   label,
   min,
@@ -65,22 +73,63 @@ const Alarm = ({
   alarmLimits,
   setAlarmLimits,
 }: AlarmProps): JSX.Element => {
+  const theme = useTheme();
+  const { initRefListener } = useRotaryReference(theme);
   const rangeValues: number[] = [alarmLimits[stateKey].lower, alarmLimits[stateKey].upper];
+  const [refs] = React.useState<Record<string, RefObject<HTMLDivElement>>>({
+    [`${stateKey}_LOWER`]: useRef(null),
+    [`${stateKey}_HIGHER`]: useRef(null),
+  });
   const setRangevalue = (range: number[]) => {
     setAlarmLimits({ [stateKey]: { lower: range[0], upper: range[1] } });
   };
+  const onClick = (value: number, type: SliderType) => {
+    setActiveRotaryReference(
+      type === SliderType.LOWER ? `${stateKey}_LOWER` : `${stateKey}_HIGHER`,
+    );
+    setAlarmLimits({
+      [stateKey]: {
+        lower: type === SliderType.LOWER ? value : rangeValues[0],
+        upper: type === SliderType.UPPER ? value : rangeValues[1],
+      },
+    });
+  };
+
+  useEffect(() => {
+    initRefListener(refs);
+  }, [initRefListener, refs]);
+
   return (
     <Grid container>
       <Grid item xs={12} style={{ paddingBottom: 20 }}>
         <Typography variant="h5">{label}</Typography>
       </Grid>
-      <Grid item xs={12}>
+      <Grid item xs={2} ref={refs[`${stateKey}_LOWER`]}>
+        <ValueClicker
+          referenceKey={`${stateKey}_LOWER`}
+          value={rangeValues[0]}
+          onClick={(value: number) => onClick(value, SliderType.LOWER)}
+          min={min}
+          max={max}
+        />
+      </Grid>
+      <Grid item xs={8}>
         <ValueSlider
+          disabled={true}
           min={min}
           max={max}
           step={step}
           onChange={setRangevalue}
           rangeValues={rangeValues}
+        />
+      </Grid>
+      <Grid item xs={2} ref={refs[`${stateKey}_HIGHER`]}>
+        <ValueClicker
+          referenceKey={`${stateKey}_HIGHER`}
+          value={rangeValues[1]}
+          onClick={(value: number) => onClick(value, SliderType.UPPER)}
+          min={min}
+          max={max}
         />
       </Grid>
     </Grid>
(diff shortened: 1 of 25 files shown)


Sudhir-dev:

ethanjli:
Observations from testing:
- It's hard for me to consistently reproduce, but sometimes when I close the Set Alarm Limits or Set Value modals without interacting with any increment/decrement buttons I still get a "Keep Previous Values?" confirmation dialog which is annoying. Sometimes I get this extraneous confirmation, while other times I don't.
- The mechanism of showing the active value being changed (e.g. in the SpO2 alarm limits) looks fine. For the visual styling of the highlight, I think it would be better if the entire region, bounded by the dotted lines, were highlighted. This could look like changing the color of the dotted line border, or it could look like changing the background color of the region within the border, or it could look like changing the colors of the UI elements (the number and buttons).
- The color contrast of the numbers in the disabled alarm limits slider is too low: white on light gray. We will need to choose some other color for this, either in this PR or a future PR.
- Currently the Set Alarm Limits modal is not the same size as the Event Log modal - so when I click the "Settings" button for an alarm from the Event Log, I clearly see that the Alarms modal is a new modal opened on top of the Event Log. To reduce visual clutter, either we should make the Set Alarm Limits modal be the same size as the Events Log modal (i.e. full-size) and have lots of whitespace around the main content; or we should close the Events Log when we click a "Settings" button from it and reopen the Events Log in the same state (including pagination state) when we press the "Confirm" button from the confirmation dialog which closes the Set Alarm Limits modal. 
- From performance testing on the RPi, the main source of UI choppiness and data dropping (which causes the waveform plots to reset themselves, related to #124 ) is not in having modals layered on top of each other but in the animations for opening and closing modals. So I think we definitely should simplify the animations (maybe even removing opening and closing animations entirely if a simple fade in/out still taxes the CPU too much); and we should also consider the layered-modal approach (where all modals are the same size) as that reduces the number of time we'd need to add and remove DOM elements or render animations. Right now the bottleneck in system performance is the CPU; RAM usage with just the dashboard open is ~950 Mb (out of 1.82 Gb available on the 2 GB RPi model, but I believe the production version will use the RPi 4 model with 4 GB RAM). When I open both the event log and the Set Alarm Limits modal, RAM usage temporarily increases to 980 Mb but drops back down to 950 Mb. I get the same RAM usage spike when I close the modals, which suggests that opening and closing modals is more taxing on the system's RAM than just keeping multiple modals around.
- I like the left vs. right layout of the decrement/increment buttons on the Alarms screen. I am worried that the numbers inside the disabled slider may be too small to read for some users - if we can also add the big number displays like what we have in the Set Alarm Limits modal, that could improve readability. We might also want to add horizontal dotted lines between the FiO2, SpO2, and HR sections.
- For the HFNC-only release of the frontend, we should remove the pagination buttons from the Alarms screen (these are the `<`, `1`, and `>` buttons above the "Apply Changes" button)

I haven't had a chance to test with the hardware rotary encoder, but otherwise this looks good. In today's meeting let's discuss which revisions to include in this PR, which ones to save for a future PR for the HFNC-only release, and which ones we need to collect more data/observations/testing on before we make a decision about what to implement.

Sudhir-dev:
@ethanjli

Have updated PR with following fixes

* Removed Animations across app
* Resized modal to be consistent across app
* Multistep popups layered modal one over the other
* Disabled color in slider made consistent with disabled button color
* Bug fixed on "Keep Previous Values?" popup reopening inconsistently without user making any changes

Sudhir-dev:
Updated with changes related to Event log

* Added enum for event log types in protobuf
* Added alarms for Settings changed, fio2 & flow
* Event log Alarm details updated to show high/low alarm range number

ethanjli (approved):
Observations from testing on the Raspberry Pi:

- The hardware rotary dial works correctly with the frontend. Currently the pushbutton of the dial doesn't seem to do anything (I'm not sure if this is because I've incorrectly configured the wiring between the rotary dial and the Raspberry Pi, or if it's because the frontend ignores pushbutton events), which I think is good enough for the HFNC-only release.
- When I'm on the dashboard and I open the Set New modal (e.g. for Flow Rate), the modal which pops up has the "back" button in the upper-right corner instead of the close button. I personally would've expected a close button to be in that place which would behave like the Cancel button, but the current design makes sense to me too and it is what the UI/UX team designed. The alternative would be to only have close buttons and not have return buttons, or to only show return buttons but in most/all cases they behave the way a close button would (since we have overlapping modals which give the illusion of a single modal changing its contents). Let's discuss to figure out what makes sense and whether we need to change anything for a future PR.
- I've discovered a bug or missing feature related to the Set New modal: I am on the dashboard, I open Set New (e.g. for FiO2), make a change without saving it, and then press the back button. This takes me to the HFNC Control screen, and the back button changes into a close button (this is good). However, when I press the close button it exits without bringing up the "Keep Previous Values?" confirmation dialog. The close button action should instead behave the same the Cancel button. Let's fix this in a future PR.
- I wonder if the "Submit" button should be colored green, the same way the Confirm button is? I also wonder if it should only say Submit when there are unsaved changes; e.g. when there are no unsaved changes it just says "Close" - but maybe it would be confusing or unintuitive to for the button text to change. How about disabling the submit button when there are no unsaved changes, so that only "Cancel" is available? Let's discuss and decide whether to change anything in a future PR.
- Now that the animations associated with the modals (and maybe also the sidebar?) have been removed, there is a lower (but still nonzero) rate at which the frontend freezes long enough to prevent the communication protocol from being serviced frequently enough. I think we should keep this change.
- On the dashboard, the right edge of the "Flow Rate" box isn't the right edge of a rounded rectangle (with rounded corners), but rather a thicker dashed line border with sharp corners. We'll need to fix that in a future PR. I can also see this issue on my own laptop, not just on the Raspberry Pi.
- As we had anticipated, resizing all modals to be large makes things look a bit strange. I think we can improve the appearance of the Events Log/Active Alarms modal by increasing the size of the table and improving the positioning of the header area ("Events Log", mute button, "Active Alarms", close button). I think the confirmation dialog is acceptable; it may or may not help to add a dotted line border (with rounded corners) around the content, might be worth trying out. The Set New modal is the least obvious one to improve: one option is to make the header area (the title, the back button, and the breadcrumbs) visually consistent and aligned with the header area of the Events Log modal and then to increase the size of the numbers and buttons being displayed, to make it more similar to the Events Log modal; another option is to reduce the width of the content and make it more similar to the confirmation dialog modal. Finally, I think making all of these modals have rounded corners of the same radius as the rounded corners on the panels of the dashboard would give a small improvement. We could fix this in either a future PR (so that we can merge in this PR now) or in this PR.
- The All Alarms screen has not been changed. How about we leave that for a future PR so that we can merge this PR in sooner?

Before merging, I'd like to go over the pattern you've set up for adding alarm/event types so that I understand how we'll do it in the future.

ethanjli:
Changes for future PRs, as discussed in the meeting:
- Let's move the "back" button in the breadcrumbs modal to be at the left of the breadcrumbs, and a close button will be added in its place
- Let's make the Submit button in the breadcrumbs modal be green (like the Confirm button of the confirmation dialog modal), and let's make it be disabled when there are no unsaved changes and enabled when there is at least one unsaved change to any of the ventilation controls
- Let's have the headers at the top, the breadcrumbs at the top just under the title in the headers, Cancel/Confirm buttons in the lower right corner, and we'll try to avoid unnaturally wide containers (e.g. dotted line border containers) and instead just have things at natural widths, centered in the modal

PR #285, opened 2021-03-02
Implement unit test cases for Protocols/Datagrams

rohanpurohit:
This PR includes:
- Initial draft for unit test cases of datagrams


rohanpurohit:
This PR does not contain `static assert fixes`, therefore the old cases pertaining to that fix still remain.
issue: https://github.com/pez-globo/pufferfish-software/pull/281#discussion_r585620496

rohanpurohit:
Changes made: 
- Rebased on develop
- added static assert for input buffer size
- updated comments for constructed and parsed datagram

@ethanjli  please review datagrams.txt

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+      THEN("The length field of the body's header matches the value returned by the length accessor method")
+      THEN("The body's payload section correctly stores the paylaod as '0xb7 0xe1 0x8d 0xaa 0x4d 0x02'")
+      THEN("The output buffer is as expected '0x00 0x06 0x01 0x02 0x03 0x04 0x05 0x02' ")
+
+    WHEN("The output buffer is too short to hold the sequence, length and the paylaod fields")
+      THEN("The write method reports out of bounds status")
+      THEN("After the write method is called, The value returned by the seq accessor method remains unchanged")
+      THEN("the value returned by the length accessor method remains unchanged")
+      THEN("The payload returned by the paylaod accessor method remains unchanged")
+      THEN("The output buffer is as expected 'empty' ")
```
ethanjli: I think the output buffer should actually remain unchanged, and here it's only empty because it was empty in the WHEN

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, The value returned by the seq accessor method remains unchanged")
+      THEN("the value returned by the length accessor method remains unchanged")
+      THEN("The payload returned by the paylaod accessor method remains unchanged")
+      THEN("The seq field of the body's header matches the value returned by the seq accessor method")
+      THEN("The length field of the body's header matches the value returned by the length accessor method")
+      THEN("The body's payload section correctly stores the paylaod as '0xc2 0x22 0x10 0xa6 0x3a 0xa5'")
+      THEN("The output buffer is as expected '0x0a 0x06 0xc2 0x22 0x10 0xa6 0x3a 0xa5' ")
+
+  GIVEN("A Datagram constructed with a non-empty paylaod buffer and a sequence ranging from 0 to 0xff")
```
ethanjli: When you say "a sequence number ranging from", do you mean "any sequence number in the range"?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, The value returned by the seq accessor method remains unchanged")
+      THEN("the value returned by the length accessor method remains unchanged")
+      THEN("The payload returned by the paylaod accessor method remains unchanged")
+      THEN("The seq field of the body's header matches the value returned by the seq accessor method")
+      THEN("The length field of the body's header matches the value returned by the length accessor method")
+      THEN("The body's payload section correctly stores the paylaod as '0xb7 0xe1 0x8d 0xaa 0x4d'")
+      THEN("The output buffer is as expected '0x00 0x05 0xb7 0xe1 0x8d 0xaa 0x4d' ")
+
+    WHEN("The paylaod given to the constructor is altered ('0xb7 0xe1 0x8d 0xaa 0x4d 0x02') before write method is called")
```
ethanjli, line 36: In this case, we should probably assert (with either just a REQUIRE or a full THEN) that the length method is inconsistent with the payload before the `write` method is called (similar to the relationship between the `crc` method and the payload in `CRCElement`)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+    WHEN("A body with a complete 2-byte header, and a non-empty paylaod buffer consistent with the length field of the header is parsed")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor")
+      THEN("the seq accessor method returns a reference to the sequence given in the Datagram constructor")
+      THEN("the payload given in the Datagram constructor is independent of the input buffer body")
+
+  GIVEN("A Datagram constructed with paylaod a non-empty paylaod buffer and sequence equal to 0")
```
ethanjli: Looks like a typo here

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+      THEN("The paylaod returned by the paylaod accessor method remains unchanged")
+      THEN("The constructor's payload buffer remains unchanged")
+
+    WHEN("A body with a complete 2-byte header, and a non-empty paylaod buffer consistent with the length field of the header is parsed")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor")
+      THEN("the seq accessor method returns a reference to the sequence given in the Datagram constructor")
```
ethanjli: The seq accessor method should never return a reference, it should always just return a value

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the input_buffer body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor") 
+      THEN("the seq accessor method returns a reference to the sequence given in the Datagram constructor")
+      THEN("the payload given in the Datagram constructor is independent of the input buffer body")
+
+    WHEN("A body with a complete 2-byte header, and a non empty payload buffer consistent with the length field of the header, is altered after it's parsed")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
```
ethanjli: Of the original or the altered input_buffer body's header?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,474 @@
...
+    WHEN("A body with a complete 2-byte header, and paylaod buffer as '0x00' consistent with the length field of the header is parsed")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the input_buffer body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor") 
+      THEN("the seq accessor method returns a reference to the sequence given in the Datagram constructor")
+      THEN("the payload given in the Datagram constructor is independent of the input buffer body")
+
+    WHEN("A body with a complete 2-byte header, and a non-empty payload buffer where the length field of the header is inconsistent with the actual length of the payload buffer")
```
ethanjli: This WHEN and the next WHEN need to specify that the parse method gets called. It might be awkward to put it at the end of the sentence for this WHEN, so you could change the order: WHEN the parse method is called on a body with ...

ethanjli (commented):
Here are my comments on the scenarios for Datagram::parse and Datagram::write.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the input_buffer body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor") 
+      THEN("the payload given in the Datagram constructor is independent of the input buffer body")
+
+
+Scenario: Datagram correctly preserves data in roundtrip writing and parsing
+  GIVEN("A datagram constructed with a non-empty payload buffer and sequence number equal to 0")
```
ethanjli: I'd prefer initializing with a non-zero sequence number, since 0 is the default value and what we'd expect to get at the end if a Datagram lost the sequence number somewhere in the roundtrip path.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+    WHEN("A body with seq, length and payload is generated, parsed into a new ParsedDatagram object, and generated again with a ConstructedDatagram object")
```
ethanjli, line 191: In this test, each section has a THEN which expects a different payload. I think the payload is supposed to be the same on each one?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+      THEN("After the transform method is called, the value returned by the output_datagram's seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The paylaod returned by the output_datagram's paylaod accessor method is equal to the paylaod from the input_buffer body")
+      THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer")
+
+Scenario: Datagram Receiver correctly reports incrementing, rollover, and resetting of expected sequence number
+
+  GIVEN("A Datagram receiver of buffer size 254 bytes and expected sequence number equal to 1")      
+  // assertions carried out in GIVEN:
```
ethanjli, line 326: These assertions need to be given more context in order to make sense, because output_datagram isn't declared in the GIVEN statement. If these are just describing the first input_buffer you call transform on, and you want to describe what's happening in setup, maybe it's enough to just say something like the following:
```
// Setup:
// Call transform with an input_buffer where the header's seq field has value 0
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+  // The output_datagram's seq accessor method returns 1
+  // The value returned by the output_datagram's length accessor method is 9
+  // The payload buffer returned by the output_datagram's payload accessor method is '0x00 0x07 0x48 0x17 0xb8 0x67 0x1c 0x45 0x28'  
+
+    WHEN("transform is called on two parseable datagram bodies, whose headers have sequence numbers as 1 & 2 respectively")
```
ethanjli: The THENs refer to "the input_buffer", I think they should refer to either "the first input_buffer" or "the second input_buffer". I would suggest introducing "input_buffer" in your WHEN, e.g. WHEN transform is called on two parseable datagram bodies as input_buffers, ...

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+
+    WHEN("transform is called on three parseable datagram bodies, whose headers have sequence numbers as 1, 10 & 11 respectively")
+      THEN("The first transform method reports ok status")
+      THEN("After the first transform method is called, the value returned by the output_datagram's seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The paylaod returned by the output_datagram's paylaod accessor method is equal to the paylaod from the input_buffer body")
+      THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer")
+
+      THEN("The second transform method reports ok status")
```
ethanjli: I'm surprised this returns ok, I would've expected it to return invalid_sequence

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+
+    WHEN("A body with a complete 2-byte header, a non-empty payload buffer consistent with the length field of the header is given to the receiver")
+      THEN("The transform method reports ok status")
+      THEN("After the transform method is called, the value returned by the output_datagram's seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The paylaod returned by the output_datagram's paylaod accessor method is equal to the paylaod from the input_buffer body")
+      THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer")
+
+Scenario: Datagram Receiver correctly reports incrementing, rollover, and resetting of expected sequence number
```
ethanjli, line 311: For certain THENs which are already fully covered by other scenarios, it may be fine to remove those THENs from this scenario if they're irrelevant to the test, e.g. THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer") and THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor") and THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header"). Removing some or all of these THENs could make it easier to understand what these tests care about and what their intent is, right now it might be a bit too distracting to keep all of them in this scenario. I think this is a judgement call for you to make.

ethanjli: From the test guidelines: "THENs should achieve full coverage of all relevant outputs, state changes, and side-effects for each WHEN." - so it comes down to deciding which of these things are relevant to each scenario. For scenarios covering basic functionality (which I think was all of the scenarios in CRCElement), we might want to cover everything; and for scenarios testing specific behaviors, we should cover some smaller subset of consequences.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+      THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer")
+
+  GIVEN("A Datagram receiver of buffer size 254 bytes and expected sequence number equal to 0xff")
+  // assertions carried out in GIVEN:
+  // The output_datagram's seq accessor method returns 0xff
+  // The value returned by the output_datagram's length accessor method is 8
+  // The payload buffer returned by the output_datagram's payload accessor method is '0xff 0x06 0xcd 0x6a 0xc2 0x7f 0xa1 0x5b'  
+
+    WHEN("transform is called on two parseable datagram bodies, whose headers have sequence numbers as 0 & 1 respectively")
```
ethanjli: The DatagramReceiver is expecting its next datagram to have sequence number 0xff. In this WHEN, do you mean that the headers will have sequence numbers 0xff and 0x00?

I'd also suggest adding a WHEN for a header with sequence 0xff and any non-0x00 value, so we can guarantee that 0xff isn't a special case where any sequence number is accepted for the next datagram.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: For this scenario, you could consider using Catch2 data generators for higher coverage of possible values of sequence numbers; up to you.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+    WHEN("A non empty payload buffer is given to the datagram sender")
+      THEN("The transform method reports ok status")
+      THEN("The seq field of the body's header is equal to the next sequence")
+      THEN("The length field of the body's header is equal to the size of the payload buffer given")
+      THEN("The body's payload section correctly stores the paylaod as '0xf9 0x23 0x4a 0xd4 0xe0'")
+      THEN("The output datagram is as expected '0x00 0x05 0xf9 0x23 0x4a 0xd4 0xe0'")
+
+    WHEN("The output datagram cannot hold enough data")
+      THEN("The transform method reports to invalid length")
+      THEN("The output buffer is as expected")
```
ethanjli: And what's the sequence number of the next datagram it will generate? (I think it should still be 0). Also, what is the output buffer supposed to be? (I think it should remain unchanged from whatever it had before transform was called, whether it's empty or has data)

This applies to the next WHEN, too

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+
+    WHEN("The input paylaod is too large for the output datagram to hold")
```
ethanjli: Is this WHEN different from the previous WHEN?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
+/* Write function
+
+Scenario: ConstructedDatagram: The write method correctly generates a body from the paylaod and sequence given in the constructor
```
ethanjli, line 3: For completeness, I'd suggest having a test for every possible payload length from 0 to 255 (in a 256-byte payload buffer), using Catch2 generators or just regular for loops.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+    
+    WHEN("2 non-empty payload buffers of capacity 254 bytes is given to the sender")
+      THEN("The first transform method reports ok status")      
+      THEN("The seq field of the body's header is equal to the next sequence")
+      THEN("The length field of the body's header is equal to the size of the payload given")        
+      THEN("The body's payload section correctly stores the paylaod as '0xc0 0x18 0x65 0xd1 0x03 0x5c'")
+      THEN("The output datagram is as expected '0x00 0x06 0xc0 0x18 0x65 0xd1 0x03 0x5c' ")
+
+      THEN("The second transform method reports ok status")      
+      THEN("The seq field of the body's header is equal to the next sequence")
```
ethanjli: GIVEN defines the next sequence as 0. This "next sequence" is different from the "next sequence" on line 422, so we should specify what it actually is.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,446 @@
...
+    
+    WHEN("2 non-empty payload buffers of capacity 254 bytes is given to the sender")
+      THEN("The first transform method reports ok status")      
+      THEN("The seq field of the body's header is equal to the next sequence")
+      THEN("The length field of the body's header is equal to the size of the payload given")        
+      THEN("The body's payload section correctly stores the paylaod as '0xc0 0x18 0x65 0xd1 0x03 0x5c'")
+      THEN("The output datagram is as expected '0xff 0x06 0xc0 0x18 0x65 0xd1 0x03 0x5c' ")
+
+      THEN("The second transform method reports ok status")      
+      THEN("The seq field of the body's header is equal to the next sequence")
```
ethanjli: Is the next sequence number here 0x00?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: I'd suggest having a test for seq 1 to seq 2. Right now we have seq 0 to seq 1, and seq 255 to seq 0; both of them are edge cases since they involve seq 0. seq 1 to seq 2 would be kind of like the inductive step in an inductive proof.

ethanjli (commented):
I reviewed the rest of datagrams.txt

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: I kept it consistent with CRCElement where we used different payload buffers for input to Parsed and Constructed CRCElements, do you want me to change it ?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Yes my bad it does return invalid_sequence

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: > The DatagramReceiver is expecting its next datagram to have sequence number 0xff. In this WHEN, do you mean that the headers will have sequence numbers 0xff and 0x00?

yep thanks for pointing it out, I did a derp in setting up the state before the WHEN. yes I meant 0xff  and 0x00

> I'd also suggest adding a WHEN for a header with sequence 0xff and any non-0x00 value, so we can guarantee that 0xff isn't a special case where any sequence number is accepted for the next datagram.

will do.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: this is how i structured these 2 tests:
`WHEN("The output datagram cannot hold enough data")`
Here input payload is 9 bytes, and output_buffer is 10 bytes long , so there is no space for the header

`WHEN("The input paylaod is too large for the output datagram to hold")`
input payload is 256 bytes, output buffer is of size 254 bytes 

Ya both will fail in `resize `and return `out_of_bounds`, I'll remove This WHEN and keep the previous one.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: > For certain THENs which are already fully covered by other scenarios, it may be fine to remove those THENs from this scenario if they're irrelevant to the test, e.g. THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer") and THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor") and THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header"). Removing some or all of these THENs could make it easier to understand what these tests care about and what their intent is, right now it might be a bit too distracting to keep all of them in this scenario. I think this is a judgement call for you to make.

Yes I concur, I will remove irrelevant cases.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: > And what's the sequence number of the next datagram it will generate? (I think it should still be 0)

next_seq is a private member, therefore we can't make any assertions on it, and as the output_datagram is empty, we can't assert if it has a seq header

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: yes

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: > For this scenario, you could consider using Catch2 data generators for higher coverage of possible values of sequence numbers; up to you.

I'm going to do this in a future PR, as generating buffers with sequences ranging from say 0 to 255 is not the issue, but passing that generated buffer in the next iteration is slightly complicated. I will try this later.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Added.

rohanpurohit:
@ethanjli updated with suggested changes, in some cases I had a follow-up comment, could you please review and let me know. thanks

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: Sounds good; for certain tests it might also be feasible to just use for loops - and that might be clearer and more readable

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: We can indirectly infer the value of next_seq by calling transform again in a THEN and checking the sequence number field of the resulting output_buffer. And ultimately as the user of DatagramSender I don't care about the existence or value of the next_seq private member, I just care that the sequence number fields of my output_buffers will increment on successful outputs and stay the same on errors

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: I don't mean using different payload buffers (we should do that), I mean that after the first write you assert that the payload is written into the body as `0x61 0x6a 0x1a 0x6a 0x29 0xcf 0x01 0x81 0xbe 0x9d`, then after the parse you assert that the payload is parsed from the body as `0x00`, and then after the second write you assert that the payload is written into the body as `0x56 0xd6 0x42 0xc5 0xe1 0xf0 0x30 0xe5 0xc8 0x4d`

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: I wonder if it's possible to add a `static_assert` to `Datagram::write` and `CRCElement::write` to check that `output_size >= PayloadBuffer::max_size() + ___HeaderProps::header_size`. Then this would be statically caught by the compiler instead of having to be covered by run-time tests.

rohanpurohit: Yes we can, I'll do that.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Yes will do that.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Yep added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: This case is redundant after adding `static_assert`. I'm leaving the comment unresolved if we need to discuss anything.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,456 @@
...
+      THEN("After the write method is called, The value returned by the seq accessor method remains unchanged")
+      THEN("the value returned by the length accessor method remains unchanged")
+      THEN("The payload returned by the paylaod accessor method remains unchanged")
+      THEN("The seq field of the body's header matches the value returned by the seq accessor method")
+      THEN("The length field of the body's header matches the value returned by the length accessor method")
+      THEN("The body's payload section correctly stores the paylaod as '0xb7 0xe1 0x8d 0xaa 0x4d'")
+      THEN("The output buffer is as expected '0x00 0x05 0xb7 0xe1 0x8d 0xaa 0x4d' ")
+
+    WHEN("The paylaod given to the constructor is altered ('0xb7 0xe1 0x8d 0xaa 0x4d 0x02') before write method is called")
+      THEN("Before the write method is called, The value returned by the length accessor method is inconsistent with the size of the altered paylaod")
```
ethanjli: In particular, we should specify what the inconsistent value is (i.e. the value returned by the length accessor method remains unchanged despite the payload alteration, and thus is inconsistent with the size of the altered payload)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,456 @@
...
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, The value returned by the seq accessor method remains unchanged")
+      THEN("the length accessor method returns a value equal to the size of the altered payload")
+      THEN("The paylaod returned by the paylaod accessor method is same as the altered paylaod buffer")
+      THEN("The seq field of the body's header matches the value returned by the seq accessor method")
+      THEN("The length field of the body's header matches the value returned by the length accessor method")
+      THEN("The body's payload section correctly stores the paylaod as '0xb7 0xe1 0x8d 0xaa 0x4d 0x02'")
+      THEN("The output buffer is as expected '0x00 0x06 0x01 0x02 0x03 0x04 0x05 0x02' ")
+
+  GIVEN("A Datagram constructed with a non-empty payload buffer of different lengths and sequence equal to 0")
```
ethanjli: Since the buffer can only have one length at a time, I'd suggest wording this as "with a non-empty payload buffer of any length from 0 to 252"; this also is closer to the language which would be used in a mathematical proof, and here the test code just happens to prove the claim by enumerating all the cases covered by the GIVEN.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,456 @@
...
+    WHEN("The parse method is called on a body with a complete 2-byte header, and a non-empty payload buffer where the length field of the header is inconsistent with the actual length of the payload buffer")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value returned by the seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The value returned by the length accessor method is not equal to the size of the payload buffer in the input_buffer body")
+      THEN("The payload returned from the payload accessor method is equal to the payload from the input_buffer body")
+      THEN("the payload accessor method returns a reference to the payload buffer given in the Datagram constructor") 
+      THEN("the payload given in the Datagram constructor is independent of the input buffer body")
+
+    WHEN("The parse method is called on a body with sequence field of the header inconsistent with sequence given in the constructor of the datagram")
```
ethanjli: I think it's better to avoid using the word "inconsistent" here, because there is no concept of data consistency between the sequence number given in the constructor and the sequence number being parsed from the body. Maybe "WHEN the parse method is called on a body whose header's sequence number field has a different value than the sequence number used to initialize the Datagram"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,456 @@
...
+      THEN("The transform method reports invalid length status")
+      THEN("After the transform method is called, the value returned by the output_datagram's seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The value returned by the output_datagram's length accessor method is equal to the length field of the input_buffer body's header")
+      THEN("The value returned by the output_datagram's length accessor method is inconsistent with the actual length of the payload buffer from the body")
+      THEN("The paylaod returned by the output_datagram's paylaod accessor method is equal to the paylaod from the input_buffer body")
+      THEN("The output_datagram's payload accessor method returns a reference to the payload given in its constructor")
+      THEN("The data in the output_datagram's payload buffer is independent of the data in input_buffer")
+
+    WHEN("transform is called on a body with a non-empty paylaod buffer and a complete 2-byte header where the sequence field of the header does not match sequence given in output_datagram's constructor")
```
ethanjli, line 295: I think what matters is that the sequence field of the header doesn't match the DatagramReceiver's expected sequence number. It should be irrelevant what sequence number is given in output_datagram's constructor, the fact that you initialized the DatagramReceiver to expect seq num 0 and you initialized output_datagram with seq num 0 makes it confusing

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
```diff
@@ -0,0 +1,456 @@
...
+      THEN("After the transform method is called, the value returned by the output_datagram's seq accessor method is equal to the sequence field of the input_buffer body's header")
+      THEN("The paylaod returned by the output_datagram's paylaod accessor method is equal to the paylaod from the input_buffer body")
+      THEN("The final sequence value returned by output_datagram's seq accessor method is 0xff")
+
+  GIVEN("A Datagram receiver of buffer size 254 bytes and expected sequence number equal to 1")     
+  // Setup:
+  // transform called on an input_buffer where the header's seq field has value 0
+
+  // assertions carried out in GIVEN:
+  // The output_datagram's seq accessor method returns 1
```
ethanjli: Is this true? If output_datagram is coming from the transform call on line 324, I'd expect the seq method to return 0, since the body's seq field has value 0

ethanjli (commented):
A few more comments in addition to the ones left unresolved above; otherwise everything looks good.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
ethanjli: Can you also add the static_assert to `CRCElement::write` in this PR?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: Sorry that's a typo. it does return 0 https://github.com/pez-globo/pufferfish-software/blob/e29b5a64925a1953ac69251b0a6f25a5b7764b68/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Datagrams.cpp#L1329

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: yes added.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/datagrams.txt
rohanpurohit: > then after the parse you assert that the payload is parsed from the body as 0x00

If this is the only concern then that was a mistake and I fixed it.

> then after the second write you assert that the payload is written into the body as 0x56 0xd6 0x42 0xc5 0xe1 0xf0 0x30 0xe5 0xc8 0x4d

Is this a concern as well? Here I constructed a new ConstructedDatagram object passing the payload as `0x56 0xd6 0x42 0xc5 0xe1 0xf0 0x30 0xe5 0xc8 0x4d` and called write after that.

ethanjli: The point of the roundtrip is that we're transforming data (generating a body), then passing the result of that (the body) into the inverse (parsing a body), then passing the result of that (the payload) into the inverse of the inverse (generating a body) - so we should be using the output of the parse step as the input of the second write step. Otherwise the last step is irrelevant to the test, right?

ethanjli: Pseudocode as discussed in meeting:
```
payload_input -> first write's ConstructedDatagram constructor
first write's ConstructedDatagram write -> first_body_output
payload_output -> ParsedDatagram constructor
first_body_output -> ParseDatagram parse
(check payload_output)
payload_output -> second write's ConstructedDatagram constructor
second write's ConstructedDatagram write -> second_body_output
```

rohanpurohit: > Pseudocode as discussed in meeting:
> 
> ```
> payload_input -> first write's ConstructedDatagram constructor
> first write's ConstructedDatagram write -> first_body_output
> payload_output -> ParsedDatagram constructor
> first_body_output -> ParseDatagram parse
> (check payload_output)
> payload_output -> second write's ConstructedDatagram constructor
> second write's ConstructedDatagram write -> second_body_output
> ```

I think you had mentioned in the meeting to keep the `ParsedDatagram constructor` payload empty.
other than that I have fixed and pushed.

Here is the code:
https://github.com/pez-globo/pufferfish-software/blob/feature/firmware-test-protocols-datagrams/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Datagrams.cpp#L736

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Datagrams.cpp
```diff
@@ -810,25 +809,26 @@ SCENARIO(
...
       }
 
       // Write
-      auto final_data = std::string("\x56\xd6\x42\xc5\xe1\xf0\x30\xe5\xc8\x4d", 10);
+      // payload_output = parsed_datagram.payload()
+      auto payload_output = std::string("\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d", 10);
       TestDatagramProps::PayloadBuffer input_payload;
-      PF::Util::convert_string_to_byte_vector(final_data, input_payload);
+      PF::Util::convert_string_to_byte_vector(payload_output, input_payload);
       TestConstructedDatagram write_datagram(input_payload, 1);
```
ethanjli, line 828: Was there a particular reason you're doing all this array copying and making a redundant hard-coded version of the data in line 738 (so that if we change line 738 but forget to change line 825, then the test continues to pass but it no longer behaves like a round-trip test), instead of just doing ` TestConstructedDatagram write_datagram(parsed_payload, 1);` or ` TestConstructedDatagram write_datagram(parsed_datagram.payload(), 1);`?

Along these lines of maintainability, but a question of saving you work rather than about the risk of a test no longer doing what we say it does: is there a reasonable way you can define `"\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d"` as a constant once, instead of hard-coded copy-pasting those bytes for all the REQUIREs?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Datagrams.cpp
```diff
@@ -0,0 +1,1928 @@
...
+        auto expected_payload = std::string("\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d", 10);
+        REQUIRE(parsed_datagram.payload() == expected_payload);
+      }
+
+      // Write
+      // payload_output = parsed_datagram.payload()
+      auto payload_output = std::string("\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d", 10);
+      TestDatagramProps::PayloadBuffer input_payload;
+      PF::Util::convert_string_to_byte_vector(payload_output, input_payload);
+      TestConstructedDatagram write_datagram(input_payload, 1);
```
ethanjli: Was there a particular reason you're doing all this array copying and making a redundant hard-coded version of the data in line 738 (so that if we change line 738 but forget to change line 825, then the test continues to pass but it no longer behaves like a round-trip test even though the scenario calls it a round-trip test), instead of just doing TestConstructedDatagram write_datagram(parsed_payload, 1); or TestConstructedDatagram write_datagram(parsed_datagram.payload(), 1);?

Along these lines of maintainability, but a question of potentially saving you work in the future rather than about the risk of a test no longer doing what we say it does: is there a reasonable way you can define `"\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d"` as a constant once, instead of hard-coded copy-pasting those bytes for all the REQUIREs?

rohanpurohit: > instead of just doing TestConstructedDatagram write_datagram(parsed_payload, 1); or TestConstructedDatagram write_datagram(parsed_datagram.payload(), 1);

RIght, so as I had used a different payload there, instead of doing something like `parsed_datagram.payload()` i simply changed the payload bytes https://github.com/pez-globo/pufferfish-software/commit/06c3eed2ce934f6997dbb49c157c5a0521e80e3d#diff-d2aefa6418e0db5f2229e9f40480ace84b32ba511eaef5a49d967b3b0b869a4fR825
I'll fix this now.

> Along these lines of maintainability, but a question of potentially saving you work in the future rather than about the risk of a test no longer doing what we say it does: is there a reasonable way you can define "\x61\x6a\x1a\x6a\x29\xcf\x01\x81\xbe\x9d" as a constant once, instead of hard-coded copy-pasting those bytes for all the REQUIREs?

Will try to do that.

ethanjli:
Great, let's merge this in! For records-keeping:

One thing I just learned about, which might make initialization of strings cleaner in future PRs, is C++'s `std::string` literal. It allows you to initialize C++ strings which may have null bytes inside them, without having to specify the length of the string: https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s . This feature was added in C++14, so we can use it.

rohanpurohit:

> One thing I just learned about, which might make initialization of strings cleaner in future PRs, is C++'s `std::string` literal. It allows you to initialize C++ strings which may have null bytes inside them, without having to specify the length of the string: https://en.cppreference.com/w/cpp/string/basic_string/operator%22%22s . This feature was added in C++14, so we can use it.

Great! will implement them in future PR's

ethanjli (approved):

PR #286, opened 2021-03-04
firmware: Vector::copy_from: detect when copy is truncated

### LINKED ISSUE #282: Vector::copy_from and calling code should detect when copy is truncated to prevent overflow
`Util::Vector::copy_from` returns silently, even if it couldn't copy all source data as requested because its internal buffer filled up in the middle as a result of the copy. It should return an `IndexStatus` indicating whether an out-of-bounds error was averted - i.e. it should copy what it can, and use `IndexStatus::out_of_bounds` to indicate if it was requested to copy data which would've been out of bounds of its internal buffer. Anything which uses Vector::copy_from should also check and handle its return code. For example, `Protocols::CRCElement::parse` and `Datagram::parse` currently silently truncate data if their payload buffers are shorter than necessary to hold all data from the input buffer, because that is the behavior of `Util::Vector::copy_from`.

For CRC elements, it may also be possible to enforce this through the compiler: we could add a static assertion into `CRCElement::parse` and/or `CRCElementReceiver::transform` comparing `input_size` against `PayloadBuffer::max_size()`, and into `CRCElement::write` and/or `CRCElementSender::transform` comparing `output_size` against `PayloadBuffer::max_size()`. These static assertions make the CRC element classes less flexible in ways we don't care about, so it could be good to add this extra level of checking if the compiler allows it. The same applies to the datagram classes.

rohanpurohit:
This PR resolves #282 by adding return codes to Vector::copy_from, in it's current state it only copies source data to internal buffer if the source data size is less than or equal to the internal buffer size, if the size of the source data is greater than internal buffer size it returns an `IndexStatus::out_of_bounds` error code.

Also Wherever copy_from is used:
`CRCElement::parse`, `CRCElement::write`, `Datagram::parse`, `Datagram::write`, 
- The copy_from method is executed before other members are updated. 
for ex: In CRCElement::parse , we would not want to update the CRC if the payload in the input_buffer is too large for the internal payload to hold

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Vector.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Vector.tpp
@@ -83,18 +83,14 @@ IndexStatus Vector<Element, array_size>::copy_from(
 template <typename Element, size_t array_size>
 IndexStatus Vector<Element, array_size>::copy_from(
     const Element *source_bytes, size_t source_size, size_t dest_start_index) {
-  size_ = array_size;
-  if (source_size + dest_start_index > size_) {
+  if (source_size + dest_start_index > array_size) {
     return IndexStatus::out_of_bounds;
   }
-  if (source_size + dest_start_index <= size_) {
-    size_ = source_size + dest_start_index;
-
-    memcpy(
-        buffer_.data() + dest_start_index,
-        source_bytes,
-        sizeof(Element) * (size_ - dest_start_index));
-  }
+  size_ = source_size + dest_start_index;
+  memcpy(
+      buffer_.data() + dest_start_index,
+      source_bytes,
+      sizeof(Element) * (size_ - dest_start_index));
   return IndexStatus::ok;
 }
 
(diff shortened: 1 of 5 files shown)


firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Vector.tpp
```diff
@@ -69,28 +69,33 @@ constexpr Element &Vector<Element, array_size>::operator[](size_t position) {
...
-void Vector<Element, array_size>::copy_from(
+IndexStatus Vector<Element, array_size>::copy_from(
     const Vector<Element, array_size> &source_bytes, size_t dest_start_index) {
-  copy_from(source_bytes.buffer(), source_bytes.size(), dest_start_index);
+  return copy_from(source_bytes.buffer(), source_bytes.size(), dest_start_index);
 }
 
 template <typename Element, size_t array_size>
-void Vector<Element, array_size>::copy_from(
+IndexStatus Vector<Element, array_size>::copy_from(
```
ethanjli, line 84: Pseudocode for copy_from:
check if we have enough capacity; if not, return early and don't change any internal state: if (source_size + dest_start_index > array_size) return out_of_bounds
resize to the exact size after copy: size_ = source_size + dest_start_index
do the memcpy

ethanjli (commented):
In all the Protocol classes, we want to always successfully parse the header, even if the payload doesn't fit. And when we write data, the lines should be in byte order (i.e. payload is written last).

Otherwise, this looks good. I will test on the Nucleo board before formally approving.

ethanjli:
Looks good in code review, and works fine in testing on the Nucleo board.

For records-keeping:

ethanjli (approved):

rohanpurohit:

PR #289, opened 2021-03-12
firmware: remove handling of data inputs from StateSynchronizer

### LINKED ISSUE #279: StateSynchronizer in firmware shouldn't be responsible for handling data inputs
Currently StateSynchronizer has an input method which basically calls the States input method, and adds return codes. We should just make the States input and output methods return return codes, and remove the input method from StateSynchronizer. We'll keep the StateSynchronizer's `input(time)` method though.

rohanpurohit:
This PR resolves #279 
Changes:
- Application/State input & output methods return error codes
- states is responsible for handling inupt in backend.tpp
- remove status code for StateSynchronizer::input 
- Add an invalid_type error code for StateSynchronizer::output
- Use Util::within_timeout in StateSynchronizer::should_output
- Update states::input method
- Move Application/States should_input method to backend.tpp

For records-keeping:

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.tpp
@@ -161,10 +161,10 @@ Backend::Status Backend::input(uint8_t new_byte) {
   }
 
   // Input into state synchronization
-  switch (synchronizer_.input(message.payload)) {
-    case BackendStateSynchronizer::InputStatus::ok:
+  switch (states_.input(message.payload)) {
+    case Application::States::InputStatus::ok:
       break;
-    case BackendStateSynchronizer::InputStatus::invalid_type:
+    case Application::States::InputStatus::invalid_type:
       // TODO(lietk12): handle error case
       return Status::invalid;
   }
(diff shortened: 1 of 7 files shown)


firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.tpp
```diff
@@ -160,11 +160,15 @@ Backend::Status Backend::input(uint8_t new_byte) {
       return Status::waiting;
   }
 
+  if (!states_.should_input(message.payload.tag)) {
```
ethanjli: Now that we're explicitly treating `States` as nothing more than a store, I think it might actually make sense to move the `should_input` method into the `Backend` class (and then rename the method to `Backend::accept_message` or `Backend::accept_state_segment`), as the logic inside `should_input` is actually specific to the behavior of `BackendReceiver` rather than `States`. If this sounds obvious to you, can you go ahead and implement it? Otherwise, let's first spend a bit of time discussing to see what consequences of this change I might have forgotten about or overlooked.

ethanjli (approved):
I've reviewed the code, and I haven't seen any problems in basic testing on the microcontroller. This is ready to merge, though I'm considering whether we might want to move the `should_input` method out of `States`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.tpp
rohanpurohit: I think it's a straightforward change, I pushed it just now @ethanjli please review it.

ethanjli:
Works fine on the microcontroller, let's merge this!

PR #291, opened 2021-03-12
Improve type code checking in firmware's Messages implementation

### LINKED ISSUE #278: Add proper type checking in the Messages.parse method
**Describe the bug**
When a uint8_t invalid type code which is not in `TaggedUnion::Tag` tries to get static casted into `TaggedUnion::Tag`, a segfault occurs.

**Expected behavior**
We could define a data structure which associates all tag uint8_t values with `TaggedUnion::Tag` enum values; if a uint8_t value isn't in the data structure, we return an error instead of trying to do a static_cast. We'll need to figure out what kind of data structure to use.

ethanjli:
This PR resolves #278 by:

- Defining a new `Util::EnumValues` variadic class template which provides a recursive constexpr function to check whether the integer provided as input is the underlying value of one of the enum values provided as a template parameter to the `Util::EnumValues` template specialization. To use this, specialize `Util::EnumValues` with an enum type and the enum values from that enum type (e.g. `using MyEnumValues = Util::EnumValues<MyEnum, MyEnum::first_value, MyEnum::second_value, MyEnum::third_value>`), and then call the `Util::EnumValues::includes(integer)` method (e.g. `if (MyEnumValues::includes(5)) {...}`). This way, the comparison of the integer against the list of enum values is statically expanded by the compiler - no need to create and dynamically iterate through a data structure or worry about bounds safety.
- Defining some basic unit tests for `Util::EnumValues`.
- Making `Protocols::Message` take a `Util::EnumValues` template specialization as a template parameter, so that the `Message::parse` can call the `EnumValues` specialization's `includes` method.
- Defining a `Application::MessageTypeValues` template specialization of `Util::EnumValues` which must be given all enum values of `MessageTypes`, and updating the `Driver::Serial::Backend::BackendReceiver::BackendMessage` template specialization to use `Application::MessageTypeValues`.
- Fixing an off-by-one error in `Protocols::Message::parse` in checking whether a message type code is within the bounds of `pb_message_descriptors`

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Application/States.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Application/States.h
@@ -7,6 +7,7 @@
 
 #pragma once
 
+#include "Pufferfish/Util/Enums.h"
 #include "Pufferfish/Util/TaggedUnion.h"
 #include "mcu_pb.h"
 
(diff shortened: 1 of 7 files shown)


ethanjli:
For records-keeping:

3. Does this work include, or is it based on, any third-party work which you did not create? Yes, the design of `Util::EnumValues` is based on a StackOverflow answer, which is licensed under CC-BY-SA, a license incompatible with our license (credited in a code comment above the declaration of `EnumValues`). However, my actual implementation of `EnumValues` is different and I only took inspiration in its approach of using template recursion (which is a standard pattern in template metaprogramming), so I think this should fall under fair use.

ethanjli:
@rohanpurohit If you can check that the tests run fine with the latest changes I've made, I think that'll be enough to merge this PR.

rohanpurohit (approved):
@ethanjli all the tests are passing.

PR #294, opened 2021-03-16
Fixed parse_network_order argument issue

renjipanicker:
The second parameter to parse_network_order() in Device.cpp should be 2 bytes (sizeof(uint16_t)) and not the buffer.size(). The latter will cause a buffer overflow error.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/SFM3019/Device.cpp
+++ b/firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/SFM3019/Device.cpp
@@ -58,11 +58,11 @@ I2CDeviceStatus Device::read_conversion_factors(ConversionFactors &conversion) {
   }
 
   conversion.scale_factor =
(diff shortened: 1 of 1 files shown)


ethanjli:
Notes from testing: before this fix, conversion had the following values:
```
scale_factor: 170
offset: -24576
flow_unit: 328
```

After this fix, conversion still has the same values - so this PR won't generate any regressions. From stepping through the code, I think what happens is that data beyond the 2nd byte for a uint16_t (which may be valid or may be out-of-bounds of the buffer) gets truncated in the conversion down to uint16_t. This would explain why the values don't change, even though the code had been incorrect before this PR. In this PR, let's also change `parse_network_order` to replace the `size` argument with `sizeof(T)`.

The other thing I'm seeing which we could address in this PR is that the files in `Driver/I2C/` could be refactored to use `Util/Endian.h`'s `read_ntoh` function, and then we could delete `parse_network_order` and `Util/Parse.h` entirely. The implementation of the `read_ntoh` functions is more readable to me, and to me the bounds-safety considerations feel more explicit in the `read_ntoh` interface than in the `parse_network_order` interface. Thoughts?

rohanpurohit:
closing check #302

PR #296, opened 2021-03-16
Alarms UX Implementation

Sudhir-dev:
This PR covers following points

* Alarms page & Alarm modal new UX implemented
* Pagination removed from Alarms page
* Multistep popup - Buttons Aligned & colored
* Multistep popup - Header Alignment issue fix
* Multistep popup - Added back icon in breadcrumbs

### DIFF
--- a/frontend/src/modules/alarms/AlarmsPage.tsx
+++ b/frontend/src/modules/alarms/AlarmsPage.tsx
@@ -26,25 +26,61 @@ const useStyles = makeStyles((theme: Theme) => ({
     borderRadius: theme.panel.borderRadius,
     flexWrap: 'nowrap',
     marginBottom: theme.spacing(2),
-    backgroundColor: theme.palette.background.paper,
   },
   leftContainer: {
     justifyContent: 'space-between',
-    borderRight: `2px dashed ${theme.palette.background.default}`,
-    padding: theme.spacing(2),
-  },
-  rightContainer: {
-    justifyContent: 'space-between',
-    padding: theme.spacing(2),
+    marginRight: theme.spacing(2),
   },
   paginationContainer: {
     alignItems: 'center',
-    width: '100%',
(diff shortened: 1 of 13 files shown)


Sudhir-dev:

Sudhir-dev:
@ethanjli I have updated PR with the following fixes

1. Disabled back button
2. Breadcrumbs fixed position
3. Rounded border of back button
4. Dashboard control section dotted line remove in the end
5. Disabled Submit button when no changes

ethanjli:
Looks good, this PR is ready to merge! A few more observations in general, to be addressed in either this PR or the next PR (I'd prefer merging in now, and the next changes can be addressed in a PR consisting entirely of various smaller UI fixes):

- The upper-right close button of the Events Log modal is not aligned with the upper-right close button of the breadcrumbs modal; and the titles of the two modals aren't consistent in position or text styling (bold vs. non-bold)
- The dotted line border around the table for the Events Log modal is awkward - the left and right edges are right next to the modal. Maybe it'd be better to completely remove the dotted edge border for the Events Log modal?
- When I switch between the Events Log modal and the Active Alarms modal, the sizes of the content container seem to be different, so the elements which are the same in both modals move. Also, the alarm mute button moves when I switch between the two.
- When I have 4 events in the Events Log modal, a vertical scrollbar appears; there should never be any scrollbar in the modals.
- The names of the second-layer breadcrumbs don't match the names specified in the Adobe XD design. For example, what the Adobe XD design specifies as "Flow Rate Settings" for the breadcrumb title is currently listed as "Set New"; and what the Adobe XD design specifies as "SpO2 Alarms" is currently listed as "Alarms". I would suggest following the Adobe XD design's breadcrumb naming so that different screens on the modal will have different breadcrumb path names

ethanjli (approved):

PR #297, opened 2021-03-17
Protocols: Add unit test cases for chunks

rohanpurohit:
This PR contains test cases for chunks.


firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+  GIVEN("A char chunk splitter with an empty internal buffer of 102 bytes")
+  // assertions:
+  // output buffer is empty
+
+    WHEN("102 non-delimiter bytes are passed as input, and 103rd byte is passed as a delimiter, and output is called between each input")
+      THEN("The input method reports ok status for all the 102 non-delimeter bytes")
+      THEN("After the input of 102 non-delimiter bytes, The output method reports waiting status")
+      THEN("The input method reports output_ready status on the 103rd byte")
+      THEN("After the delimiter is passed to input, The output method reports ok status")
+      THEN("The output buffer is as expected")
```
ethanjli: For all tests in this file, I would suggest specifying the length of the expected output buffer, e.g. THEN the output buffer has the expected sequence of 102 bytes. I think this will help make clear what the contents are, in the same way that the tests for CRCElements explicitly wrote out the expected bytes for short bodies/payloads.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+
+  GIVEN("A uint8_t chunksplitter with an empty internal buffer of capacity 256 bytes")
+  // assertions:
+  // output buffer is empty
+
+    WHEN("A delimiter is passed as input and output method is called after that")
+      THEN("The input method reports output_ready status")
+      THEN("The output method reports ok status")      
+      THEN("The output buffer remains unchanged")
```
ethanjli: When output buffers should remain empty in this file (e.g. also in line 34), it'll probably be clearer to say that specifically (THEN the output buffer remains empty)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+Scenario: A ChunkSplitter constructed with default value for delimeter and include_delimeter boolean as true correctly handles input chunks of acceptable lengths
+
+  GIVEN("A uint8_t chunksplitter with an empty internal buffer of capacity 256 bytes")
+  // assertions:
+  // output buffer is empty
+
+    WHEN("A delimiter is passed as input and output method is called after that")
+      THEN("The input method reports output_ready status")
+      THEN("The output method reports ok status")      
+      THEN("The output buffer remains unchanged")
```
ethanjli: Is this correct? With `include_delimiter == true` I would expect that the output buffer should be a 1-byte buffer, consisting of the delimiter

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("After the delimiter is passed as input, The output method reports ok status")
+      THEN("The output buffer is as expected")      
+
+    WHEN("255 non-delimiter bytes are passed as input, and 256th byte is passed as a delimiter, and output is called between each input")
+      THEN("The input method reports ok status for all the 255 non-delimeter bytes")
+      THEN("The output method reports waiting status")
+      THEN("The output buffer remains unchanged")
+      THEN("The input method reports output_ready status on the 256th delimeter byte")
+      THEN("After the delimiter is passed as input, The output method reports ok status")
+      THEN("The output buffer is as expected")
```
ethanjli: To elaborate on a previous comment, we might also describe something distinctive about this buffer, e.g. THEN the output buffer has the expected sequence of 256 bytes, where the last byte is a delimiter. This helps someone reading this understand what makes the behavior in this scenario different from the behavior in the previous scenario.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The input method reports invalid_length status")
+
+    WHEN("More than 256 non-delimiter bytes are passed as input")
+      THEN("The input method reports invalid_length status on 257th byte and for all the subsequent bytes")
+    
+    WHEN("More than 256 non-delimiter bytes are passed as input, after which a delimiter is passed and output is called")
+      THEN("The input method reports invalid_length status for all the 256 non-delimiter bytes")
+      THEN("The input method reports output ready status on the delimeter byte")
+      THEN("the output method reports invalid_length status")
+      THEN("The output buffer is as expected")
```
ethanjli: To elaborate on a previous comment, we might also describe something distinctive about the expected output buffers in this scenario, e.g. THEN the output buffer has the expected sequence of 256 bytes, consisting of the 256 bytes which had initially filled the internal buffer.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The input method reports invalid_length status")
+
+    WHEN("More than 256 non-delimiter bytes are passed as input")
+      THEN("The input method reports invalid_length status on 257th byte and for all the subsequent bytes")
+
+    WHEN("More than 256 non-delimiter bytes are passed as input, after which a delimiter is passed and output is called")
+      THEN("The input method reports invalid_length status for all the 256 non-delimiter bytes")
+      THEN("The input method reports output ready status on the delimeter byte")
+      THEN("the output method reports invalid_length status")
+      THEN("The output buffer is as expected")
```
ethanjli: In each test in this scenario, should I expect the last byte to be a delimiter byte? I think the answer is no

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+
+    WHEN("A delimeter byte is passed as input and output is called")
+      THEN("The input method reports invalid_length status")
+      THEN("the output method reports invalid_length status")
+      THEN("The output buffer is as expected")
+
+Scenario: After output is consumed, ChunkSplitter's behavior resets to that of a ChunkSplitter with an empty internal buffer
+
+  GIVEN("A uint8_t ChunkSplitter completely filled with input data, and no delimiters are consumed with an output call")
```
ethanjli: With the changes we've made to the input method, it's no longer possible to fill the internal buffer with inputs including multiple delimiter bytes inside the internal buffer. I think we can rephrase this like "GIVEN a uint8_t ChunkSplitter of capacity 256 bytes which has received 256 bytes as input, including multiple delimiter bytes".

However, I think this GIVEN should be moved into a different scenario (the scenario testing the input_overwritten feature), and other tests should be defined for this scenario, e.g. starting from an empty ChunkSplitter, filling its buffer, consuming with output, and then being able to completely fill its buffer again. We could even do this in a `for` loop for _n_ iterations.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+    WHEN("10 bytes of non-delimiter bytes followed by a delimiter byte are passed as input, and output is called between each input")
+      THEN("The input method reports ok status")
```
ethanjli: The GIVEN as currently written allows for the ChunkSplitter to be completely filled with non-delimiter input data, where (trivially) no delimiters were consumed with an output call; then the input method should return invalid_length. I think you would get this same result if you had a delimiter followed by at least ~247 bytes of non-delimiter bytes. So the GIVEN is underspecified for the set of WHEN/THENs you have here; I think it only works if there's a delimiter somewhere in the last ~247 bytes of input.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The output method reports ok status")
+      THEN("The output buffer is as expected")
+
+Scenario: A chunksplitter constructed with a non-zero delimeter parameter and include_delimeter boolean as false correctly handles input chunks
+
+  GIVEN("A uint8_t chunksplitter constructed with value of delimeter parameter as 0x01 is completely filled with non-delimeter input data")
+  // assertions:
+  // output buffer is empty
+
+    WHEN("257th non-delimiter byte is passed as input")
```
ethanjli: We could even specify that the 257th non-delimiter byte is `0x00`, to test that it's not erroneously treated as a delimiter by this ChunkSplitter which is supposed to use `0x01` as the delimiter.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+    WHEN("The output method is called")
+      THEN("The output method returns waiting status")
+      THEN("The output buffer remains unchanged")
+
+    WHEN("A delimeter byte is passed as input and output is called")
+      THEN("the input method reports output ready status on the delimeter byte")
+      THEN("the output method reports ok status")
+      THEN("The output buffer is as expected")
+
+Scenario: A chunksplitter constructed with a non-zero delimeter parameter and include_delimeter boolean as true correctly handles input chunks
```
ethanjli: You could merge this scenario with the previous one, and specify the value of `include_delimiter` in the GIVEN, since both of these scenarios are there to test the non-zero delimiter functionality.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+    WHEN("A delimeter byte is passed as input and output is called")
+      THEN("The input method reports output_ready status on the delimeter byte")
+      THEN("the output method reports ok status")
+      THEN("The output buffer is as expected")
+
+Scenario: A ChunkSplitter constructed with default parameters correctly handles input of multiple delimited chunks before output consumes them
+
+  GIVEN("A uint8_t chunksplitter with an empty internal buffer of capacity 256 bytes")
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after each delimeter input")
```
ethanjli: Output is called after each delimiter, so we should also have THENs about the outputs of the first output call, not just the last one. Since the next WHEN includes THENs about input_overwritten, and since this scenario is about input_overwritten, I think this WHEN should also have a THEN about input_overwritten after each input method call, as it should always be left unchanged (unless I've forgotten something). See my next comment about testing for whether input_overwritten is reset to false or is left unchanged.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The input method reports ok status for all the non-delimeter bytes")
+      THEN("The input method reports output_ready status for the delimeter bytes")
+      THEN("the output method reports ok status after the last delimeter byte")
+      THEN("The output buffer is as expected '\x04\x05' ")
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after all the bytes are passed")
+      THEN("The input method reports ok status on the first 3 bytes")
+      THEN("The input method reports output_ready status on the 4th byte") // delimeter byte
+      THEN("The input method reports ok status and sets the input_overwritten flag on the 5th byte")
```
ethanjli: What does input do to the input_overwritten flag on subsequent bytes? Does it reset it to false, or does it leave it unchanged from the previous value? If input leaves input_overwritten unchanged, we should design tests to check this (e.g. in one test, input_overwritten is true before input and we check that it stays true; and in another test, input_overwritten is false before input and we check that it stays false).

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+    WHEN("The output is called after the internal buffer is partially filled")
+      THEN("The output method reports waiting status")
+      THEN("The output buffer remains unchanged")
+
+    WHEN("A delimeter byte is passed as input and output is called")
+      THEN("The input method reports output_ready status on the delimeter byte")
+      THEN("the output method reports ok status")
+      THEN("The output buffer is as expected")
+
+Scenario: A ChunkSplitter constructed with default parameters correctly handles input of multiple delimited chunks before output consumes them
```
ethanjli: It might be more concise to say that this scenario is about detection/reporting of whether inputs are overwritten before output consumes them (I am not suggesting the final wording here), because the first WHEN does have output consume inputs before they can be overwritten

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The output buffer is as expected '\x04\x05' ")
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after all the bytes are passed")
+      THEN("The input method reports ok status on the first 3 bytes")
+      THEN("The input method reports output_ready status on the 4th byte") // delimeter byte
+      THEN("The input method reports ok status and sets the input_overwritten flag on the 5th byte")
+      THEN("The output method reports ok status")
+      THEN("The output buffer only contains the last chunk '0x04 0x05' ")
+
+Scenario: A chunksplitter constructed with default delimeter parameter and include_delimeter boolean parameter as true correctly handles input of multiple delimited chunks before output consumes them
```
ethanjli: Like in a previous comment, we could also merge this scenario with the previous one. The comments in the previous scenario apply here, too

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,262 @@
...
+      THEN("The input method reports output_ready status")
+      THEN("the output method reports ok status")
+      THEN("The output buffer is as expected '\x04\x05\x00' ")
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after all the bytes are passed")
+      THEN("The input method reports ok status on the first 3 bytes")
+      THEN("The input method reports output_ready status on the 4th byte") // delimeter byte
+      THEN("The input method reports ok status and sets the input_overwritten flag on the 5th byte")
+      THEN("The output method reports ok status")
+      THEN("The output buffer only contains the last chunk '0x04 0x05' ")
```
ethanjli: I think the lack of a 0x00 as the last byte is a typo

ethanjli (commented):
Here's my initial review of chunks.txt.

Since ChunkSplitter and ChunkMerger represent inverse operations, I think it's worth also designing at least one round-trip test.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
rohanpurohit: Ya it's wrong. its a 1-byte buffer with delimiter

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+    WHEN("A delimiter is passed as input and output method is called after that")
+      THEN("The input method reports output_ready status")
+      THEN("The output method reports ok status")      
+      THEN("The output buffer is a single null byte buffer")
+
+    WHEN("255 non-delimiter bytes and 256th byte as a delimeter are passed as input, after which output is called") {
+      THEN("The input method reports ok status for all the 255 non-delimeter bytes")
+      THEN("The input method reports output_ready status on the 256th delimeter byte")
+      THEN("After the delimiter is passed as input, The output method reports ok status")
+      THEN("The output buffer has the expected sequence of 255 bytes where the last byte is a delimeter")
```
ethanjli: Since we have `include_delimiter == true` and we passed 256 bytes as input, I think the output buffer should have a total of 256 bytes? (this also applies to the next WHEN0

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+
+  GIVEN("A uint8_t chunk splitter with a completely full internal buffer of 256 bytes with no delimeters")
+  // assertions:
+  // output buffer is empty
+
+    WHEN("257th non-delimiter byte is passed as input")
+      THEN("The input method reports invalid_length status")
+
+    WHEN("More than 256 non-delimiter bytes are passed as input")
+      THEN("The input method reports invalid_length status on 257th byte and for all the subsequent bytes")
```
ethanjli: The internal buffer is already completely full, so we should get `invalid_length` on every input byte passed in with the WHEN, not just bytes after the 256th byte. Maybe this test previously was structured as GIVEN with an empty ChunkSplitter and a WHEN passing in 256 bytes, and then passing in more bytes afterwards, and later the first 256 bytes of input were moved into the GIVEN. (this also applies to the next scenario)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+    
+    WHEN("More than 256 non-delimiter bytes are passed as input, after which a delimiter is passed and output is called")
+      THEN("The input method reports invalid_length status for all the 256 non-delimiter bytes")
+      THEN("The input method reports output ready status on the delimeter byte")
+      THEN("the output method reports invalid_length status")
+      THEN("The output buffer has the expected sequence of 512 bytes where 256 bytes were filled initially to the internal buffer")
```
ethanjli: Where was the idea of a 512-byte output buffer introduced for a 256-byte ChunkSplitter? (this also applies to the next WHEN, where a 257-byte output buffer is asserted) The corresponding THENs in the next scenario make more sense to me - each of them specifies the expected sequence of 256 bytes

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+  // output buffer is empty  
+
+    WHEN("The output is called after the internal buffer is partially filled")
+      THEN("The output method reports waiting status")
+      THEN("The output buffer remains empty")
+
+    WHEN("A delimeter byte (0x01) is passed as input and output is called")
+      THEN("The input method reports output_ready status on the delimeter byte")
+      THEN("the output method reports ok status")
+      THEN("The output buffer has expected sequence of 129 bytes consisting of the 128 bytes which had initially filled the internal buffer with delimeter as the last byte")
```
ethanjli, line 169: Since this THEN asserts the length of the output buffer, we should probably specify in the GIVEN how many bytes were in the chunksplitter: GIVEN A chunksplitter constructed with ... which is partially filled with 128 non-delimiter bytes

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+    WHEN("A delimeter byte (0x01) is passed as input and output is called")
+      THEN("the input method reports invalid_length status on the delimeter byte")
+      THEN("the output method reports invalid length status")
+      THEN("The output buffer has expected sequence of 256 bytes consisting of the 256 bytes which had initially filled the internal buffer without the delimeter as the last byte")
+
+  GIVEN("A uint8_t chunksplitter constructed with value of delimeter parameter as 0x01 and include_delimeter boolean as true is partially filled with non-delimeter input data")  
+  // assertions:
+  // output buffer is empty  
+
+    WHEN("The output is called after the internal buffer is partially filled")
```
ethanjli: Since GIVEN already declares that the internal buffer is partially filled and this WHEN isn't adding more bytes to the internal buffer, let's shorten this to `WHEN The output method is called`

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
+/* chunks
+
+Scenario: A ChunkSplitter constructed with default parameters correctly handles input chunks of acceptable lengths
```
ethanjli, line 3: We should either add a scenario like `After output is consumed, ChunkSplitter's behavior resets to that of a ChunkSplitter with an empty internal buffer` or add tests for that into this scenario. The behavior we would test is that after consuming output, we are able to add 256 bytes of input without getting an invalid_length error. Refer to https://github.com/pez-globo/pufferfish-software/pull/297#discussion_r597489271 for more details.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input method reports output_ready status on the delimeter byte")
+      THEN("the output method reports ok status")
+      THEN("The output buffer has expected sequence of 129 bytes consisting of the 128 bytes which had initially filled the internal buffer with delimeter as the last byte")
+
+Scenario: A ChunkSplitter correctly reports if input is overwritten before or after the output is consumed
+
+  GIVEN("A uint8_t chunksplitter constructed with default parameters with an empty internal buffer of capacity 256 bytes")
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after each delimeter input")
+      THEN("The input method reports ok status for all the non-delimeter bytes")
```
ethanjli: "all the non-delimiter bytes" means 0x01, 0x02, 0x03, 0x04, and 0x05. Maybe you could just say "for the first 3 non-delimiter bytes", since it seems that you're writing these THENs based on the sequence of input bytes

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input method reports output_ready status for the delimeter byte")
```
ethanjli: "the delimiter byte" is ambiguous because we have two delimiter bytes in the input. "the first delimiter byte" would solve this.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input_overwritten flag is false")
```
ethanjli: I think the actual behavior is that the `input_overwritten` flag should remain unchanged on the each input method call? Also, are you checking the state of this flag after every input method call, or only after the input method call of the delimiter?

I would suggest specifying the value of the `input_overwritten` output parameter in your WHEN, and having two WHENs: one where it's initialized as false and one where it's initialized as true. In both WHENs, these THENs would still be `THEN the input_overwritten flag is unchanged`.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input method reports ok status for the subsequent non-delimeter bytes")
+      THEN("The input_overwritten flag is unchanged")
+      THEN("The input method reports output_ready status for the delimeter byte")
+      THEN("the output method reports ok status after the last delimeter byte")
+      THEN("The output buffer is as expected '\x04\x05' ")   
+
+    WHEN("input to the chunksplitter is '0x01 0x02 0x03 0x00 0x04 0x05 0x00' and output is called after all the bytes are passed")
+      THEN("The input method reports ok status on the first 3 bytes")
+      THEN("The input method reports output_ready status on the 4th byte") // delimeter byte
+      THEN("The input method reports ok status and sets the input_overwritten flag on the 5th byte")
```
ethanjli: I would suggest specifying that `input_overwritten` gets set to true, since an incorrect implementation could set it to false.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input method reports ok status for the subsequent byte and input_overwritten flag remains unchanged")
```
ethanjli: Is this supposed to be "reports ok status for every subsequent byte"?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The output buffer only contains the last chunk '0x04 0x05' ")
+
+  GIVEN("A uint8_t ChunkSplitter constructed with default parameters, and capacity of 256 bytes which has received 256 non-delimiter bytes as input, and then a delimeter is passed as input")
+
+    WHEN("10 bytes of non-delimiter bytes followed by a delimiter byte are passed as input, and output is called between each input")
+      THEN("The input_overwritten flag is set to true on the first byte")
+      THEN("The input method reports ok status for all the non-delimeter bytes")
+      THEN("The input_overwritten flag remains unchanged after input of all 10 bytes")
+      THEN("The output method reports waiting status")
+      THEN("The input method reports output_ready status on passing the delimeter")
```
ethanjli, line 201: And the `input_overwritten` flag should remain unchanged on this input method call, too

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The output method reports ok status")
+      THEN("The output buffer has an expected sequence of 10 bytes without delimeters")
+
+  GIVEN("A uint8_t ChunkSplitter constructed with default parameters, and capacity of 256 bytes which is partially filled with non-delimeter input data")
+
+    WHEN("10 bytes of non-delimiter bytes followed by a delimiter byte are passed as input, and output is called between each input")
+      THEN("The input_overwritten flag is false")
```
ethanjli: Since this assertion is made before any method calls, do you mean for the statement in this THEN to be part of the WHEN? (e.g. WHEN 10 bytes of non-delimiter bytes followed by a delimiter byte are passed as input, with the input method's input_overwritten flag initialized as false, and output is called...)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+      THEN("The input method reports ok status for all the non-delimeter bytes")
+      THEN("The input_overwritten flag remains unchanged after input of all 10 bytes")
+      THEN("The output method reports waiting status")
+      THEN("The input method reports output_ready status on passing the delimeter")
+      THEN("The output method reports ok status")
+      THEN("The output buffer has an expected sequence of 10 bytes without delimeters")
```
ethanjli: The ChunkSplitter was initialized with non-delimiter input, so I'd expect the output buffer to have the initial data from the GIVEN, followed by the 10 bytes of input from the WHEN

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,282 @@
...
+
+  GIVEN("A ChunkMerger with delimeter equal to 0x01")
+    WHEN("A partially full buffer of size 10 bytes is passed as input to transform method")
+      THEN("The transform method reports ok status")
+      THEN("The ChunkMerger appends a delimeter")
+
+    WHEN("Completely full buffer of size 30 bytes is passed as input to transform method")
+      THEN("The transform method reports out_of_bounds status")
+      THEN("The delimeter is not appended to the buffer")
+
```
ethanjli, line 273: Since ChunkSplitter and ChunkMerger represent inverse operations, I think it's worth also designing at least one round-trip test involving two chunks (i.e. we pass the outputs of two ChunkMerger transforms into the inputs of ChunkSplitter and check the correctness of the outputs of the ChunkSplitter after each chunk). This would be a new scenario.

ethanjli (commented):
I did another round of review on the entire chunks.txt file; the comments are all about refining the changes made since the last review, so once they get addressed, I think this will be ready to merge.

For records-keeping:

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
rohanpurohit: yep! these cases are wrong I'm deleting them, as the next scenario covers the case when output is called after passing another byte

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
rohanpurohit: So I have added a for loop where we would pass in chunks of varying lengths like 1 byte , 3 bytes to 255 bytes all at once and that would report ok status each of those inputs.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
rohanpurohit: > I think the actual behavior is that the `input_overwritten` flag should remain unchanged on the each input method call? Also, are you checking the state of this flag after every input method call, or only after the input method call of the delimiter?
> 

only after passing a delimiter

rohanpurohit: > I think the actual behavior is that the `input_overwritten` flag should remain unchanged on the each input method call? Also, are you checking the state of this flag after every input method call, or only after the input method call of the delimiter?
> 

In most cases after passing in a delimiter

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
ethanjli: Sketch of round-trip test from meeting:

```
Vector<ByteVector> merged1.extend([0x01 0x02 0x03] -ChunkMerger-> [0x01 0x02 0x03 0x00])
Vector<ByteVector> merged1.extend([0x04 0x05 0x06] -ChunkMerger-> [0x04 0x05 0x06 0x00])
Vector<ByteVector> merged1.extend([0x07 0x08 0x09] -ChunkMerger-> [0x07 0x08 0x09 0x00])

ByteVector merged1 == [0x01 0x02 0x03 0x00 0x04 0x05 0x06 0x00 0x07 0x08 0x09 0x00]

merged1 -ChunkSplitter-> Vector<ByteVector> split
    0x01 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x02 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x03 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x00 -> ChunkSplitter.input -> output_ready, so split.push_back(ChunkSplitter.output -> [0x01 0x02 0x03])
    0x04 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x05 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x06 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x00 -> ChunkSplitter.input -> output_ready, so split.push_back(ChunkSplitter.output -> [0x04 0x05 0x06])
    0x07 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x08 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x09 -> ChunkSplitter.input -> ok, so don't acll ChunkSplitter.output or split.push_back
    0x00 -> ChunkSplitter.input -> output_ready, so split.push_back(ChunkSplitter.output -> [0x07 0x08 0x09])

Vector<ByteVector> split == [[0x01 0x02 0x03] [0x04 0x05 0x06] [0x07 0x08 0x09]]

split -ChunkMerger-> ByteVector merged2
    merged2.extend(split[0] -ChunkMerger-> [0x01 0x02 0x03 0x00])
    merged2.extend(split[1] -ChunkMerger-> [0x04 0x05 0x06 0x00])
    merged2.extend(split[2] -ChunkMerger-> [0x07 0x08 0x09 0x00])

ByteVector merged2 == [0x01 0x02 0x03 0x00 0x04 0x05 0x06 0x00 0x07 0x08 0x09 0x00]
```

rohanpurohit: Added here : https://github.com/pez-globo/pufferfish-software/blob/fb237485b1e668d6bc9fa4b0bc9a2bfce24e84e1/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt#L274

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,298 @@
...
+      THEN("The output buffer has the expected sequence of 256 bytes that had initially filled the internal buffer without the delimeter as the last byte")
+
+    WHEN("A delimeter byte is passed as input and output is called")
+      THEN("The input method reports invalid_length status")
+      THEN("the output method reports invalid_length status")
+      THEN("The output buffer has the expected sequence of 256 bytes that had initially filled the internal buffer without the delimeter as the last byte")
+
+Scenario: After output is consumed, ChunkSplitter's behavior resets to that of a ChunkSplitter with an empty buffer      
+  
+  GIVEN("A uint8_t chunksplitter constructed with default parameters is completely filled with 256 bytes of input data and output is consumed by passing a delimeter")
```
ethanjli:
```suggestion
  GIVEN("A uint8_t chunksplitter constructed with default parameters, which is completely filled with 256 bytes of input data followed by a delimiter, and whose output is then consumed")
```
This is my understanding of what you mean in this GIVEN. This rephrasing turns the contents of the GIVEN from a complete sentence into a noun.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,298 @@
...
+
+    WHEN("Completely full buffer of size 30 bytes is passed as input to transform method")
+      THEN("The transform method reports out_of_bounds status")
+      THEN("The delimeter is not appended to the buffer")
+
+Scenario: ChunkMerger and Chunksplitter roundtrip behaves properly
+
+  GIVEN("A uint8_t chunksplitter with empty internal buffer and a ChunkMerger constructed with default parameters")
+
+    WHEN("A merged chunk is generated by passing 3 different byte vectors and calling transform, which is then passed as input to a chunksplitter, then the merged output is passed to a new Chunkmerger")
```
ethanjli:
```suggestion
    WHEN("A merged chunk is generated by passing 3 different byte vectors and calling transform, and its results are then passed as input to a ChunkSplitter, and its outputs are passed to a new ChunkMerger")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,298 @@
...
+    //ChunkMerger
+      THEN("The transform method reports ok status for first bytevector")
```
ethanjli:
```suggestion
      THEN("The transform method of the first ChunkMerger reports ok status for first ByteVector")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,298 @@
...
+      THEN("The ChunkMerger appends the delimeter '\xda\xb3\x1b\x94\xa5\x00'")    
+      THEN("The transform method reports ok status for second bytevector")
+      THEN("The ChunkMerger appends the delimeter '\x1b\x1a\xc5\xd9\xf3\x00'")
+      THEN("The transform method reports ok status for third bytevector")
+      THEN("The ChunkMerger appends the delimeter '\xbb\x7d\xa8\x01\xf1\x00'")
+    
+    //ChunkSplitter
+      //setup : input to chunksplitter '\xda\xb3\x1b\x94\xa5\x00\x1b\x1a\xc5\xd9\xf3\x00\xbb\x7d\xa8\x01\xf1\x00'
+      THEN("The input method reports ok status for the non-delimeter bytes")
```
ethanjli:
```suggestion
      THEN("The input method of the ChunkSplitter reports ok status for the non-delimeter bytes")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/chunks.txt
```diff
@@ -0,0 +1,298 @@
...
+      THEN("The input method reports output_ready status for the delimeter bytes")
```
ethanjli: What does the output method report? What does the output method produce?

ethanjli (commented):
Here's my review of the changes for the test of resetting of behavior after the output method is called, and the round-trip test.

rohanpurohit:

ethanjli (approved):

PR #298, opened 2021-03-18
Protocols: Add unit test cases for messages

rohanpurohit:
This PR contains test cases for messages


rohanpurohit:
Initial Draft, I think wording of the test descriptions needs more refining.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
+/* Write function
+Scenario: Message: The write method correctly generates an encoded body from the protobuf message descriptor payload
+
+  GIVEN("A Message object constructed with StateSegment Taggedunion, default message types and a max size of 252 bytes")
```
ethanjli: Since no explicit constructor is defined for `Message`, I think it's more accurate to say "A Message object default-constructed with ... as its class template parameters". Or you could use this GIVEN to also initialize the values of the `type` and `payload` members.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
+/* Write function
+Scenario: Message: The write method correctly generates an encoded body from the protobuf message descriptor payload
```
ethanjli, line 2: All tests in this scenario need to describe what the value of the `type` and `payload` members should be after `write` is called.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("After the write method is called, the output buffer remains unchanged")
+
+    WHEN("The output buffer is too small to hold the generated body")
+      THEN("The write method reports invalid length status")
+      THEN("After the write method is called, the output buffer remains unchanged")
+
+    WHEN("The size of the output buffer is zero")
+      THEN("The write method reports invalid length status")
+
+    // Intent: To show that the write method does not change the internal "type" member of the message class
```
ethanjli: I'm surprised, because the first line of the `write` method is: `auto type = static_cast<uint8_t>(payload.tag);`. I must be missing/misunderstanding something here

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+
+    WHEN("The value of the message payload tag is greater than the size of the message descriptor array")
+      THEN("The write method reports invalid type status")
+      THEN("After the write method is called, the output buffer remains unchanged")
+
+    WHEN("The message payload tag is such that it corresponds to Unrecognized message descriptor in descriptor array")
+      THEN("The write method reports invalid type status")
+      THEN("After the write method is called, the output buffer remains unchanged")
+
+    WHEN("The output buffer is too small to hold the generated body")
```
ethanjli: Either:
1. In this WHEN and the next WHEN, you should describe the values of the `type` and `payload` members, or
2. You could split these WHENs into another GIVEN which describes the values of those members.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+    // Intent: To show that the write method does not change the internal "type" member of the message class
+    WHEN("The internal type member is set to an arbitrary value and then write method is called")
+    // setup:
+    // type is set to 4
+    // A sensor measurements payload message is set, which has value of type as 2
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, type member remains unchanged)
+
+
+  GIVEN("A Message object constructed with StateSegment Taggedunion, default message types and a max size of 126 bytes")
```
ethanjli: It looks like I might not be using the `max_size` template parameter anywhere in the implementation of the `Messages` classes. I would suggest adding `static_asserts` comparing `max_size` against the sizes of the `input_size` and/or `output_size` method template parameters into the definitions of the `parse` and `write` methods.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+    WHEN("The payload is a sensor measurements message and write method is called")
+      THEN("The write method reports ok status")
+      THEN("The body's header correctly stores the value of message type obtained from the message payload tag")
+      THEN("The output buffer is as expected")
```
ethanjli: I mentioned in the review of #297 that it could be good to describe the output buffer (e.g. by giving its expected length), and I think it would be useful here too

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+
+/* Parse function
+
+Scenario: Message: The parse method correctly updates internal message type and sets the message payload fields after decoding from the input buffer
+
+  GIVEN("A Message object constructed with StateSegment Taggedunion, default message types and a max size of 252 bytes")
+
+    WHEN("A body with an encoded payload and 1 byte header of value equal to 0 is parsed")
+      THEN("The parse method reports invalid type status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
```
ethanjli: The header will be changed to add a reserved field for future use, so this THEN should specify that it's comparing against the type field of the header.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("The payload Taggedunion tag is set to Unknown")
+      THEN("The message payload fields are not set")
+
+    WHEN("An empty input buffer body is parsed")
+      THEN("The parse method reports invalid length status")
+      THEN("The payload Taggedunion tag is set to Unknown")
```
ethanjli: I don't think `parse` sets the internal members if it returns `invalid_length`, I think it just leaves them at whatever their previous values were. If this is accurate, we should design tests to check that it actually does leave them at the previous values (this will involve adding a new GIVEN with different initial values of the members, and maybe the GIVENs should be reorganized under a separate scenario)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("The message payload fields are not set")
+
+    WHEN("A body with an empty payload and 1 byte header of value equal to the message descriptor array size is parsed")
+      THEN("The parse method reports invalid type status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to Unknown")
```
ethanjli: There are three ways `parse` can return `invalid_type`; in one of them, it does not change `payload.tag`, and in the other two, it depends on the GIVEN/WHEN (this will vary depending on the `MessageTypes` template parameter and the `pb_protobuf_descriptors` argument). We should cover each of the three ways `invalid_type` can be returned. It could make sense for this to be a separate scenario, because it may involve multiple GIVENs.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("The message payload fields are not set")
+
+    WHEN("A body with a 1 byte header and payload of sensor_measurements message with invalid encoding is parsed")
+      THEN("The parse method reports invalid encoding status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to sensor_measurements")
+      THEN("The message payload fields are not set")
```
ethanjli: I think this contradicts the previous line, as `payload.tag` is a field of `payload`. Maybe you mean that the `payload.value` field is not set?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+
+    WHEN("The input buffer size is greater than the payload max size")
```
ethanjli: This doesn't make sense to me, because Messages.tpp never refers to `payload_max_size`.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("The status of the parse function returns invalid encoding")
+
+/* test cases for parse functions for each message type
+    WHEN("A body with 1 byte header and a paylod of {Message type} message type is parsed")
```
ethanjli: I think this underspecifies the contents of the header - the type field should be consistent with the payload.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,181 @@
...
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to {Message type}")
+      THEN("The message payload field values are as expected")
```
ethanjli: The language is a bit ambiguous. I think rewording it will resolve this: "THEN the payload.values field has the expected {Message type} field values" or "THEN the payload field's values field has the expected {Message type} field values" or something similar.

Whatever you choose, let's use the same pattern when we refer to the payload.tag field (alternately, "the payload field's tag field" or something similar).

ethanjli (commented):
Here's my review of the write and parse methods of messages.txt. I would suggest going over every test again and comparing against the test patterns on Notion as a checklist, because I might've missed other items which are missing in the current version. For some of the comments, it might help for us to review+discuss some of logic (or illogical decisions I might've made) in the design and implementation of the `Message` methods.

Comments made here also apply to roundtrip and MessageReceiver and MessageSender, so I'll review those after the next round of changes.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: Let's make the `Message::write` method's interface more consistent with the `Datagram::write` and `CRCElement::write` method interfaces. So the contact the method should uphold is that it will update the type member, and the method is not const. This also means that `MessageSender::transform` should take a `const TaggedUnion &payload` (where `TaggedUnion` is now a template parameter for `MessageSender`) instead of `const Message &input_message`; and `BackendSender::transform` should similary take a `const Application::StateSegment &state_segment` instead of a `const BackendMessage &input_message`. Then `Backend::output`, instead of instantiating `BackendMessage message` and calling `synchronizer_.output(message.payload)`, should instantiate an `Application::StateSegment state_segment` and call `synchronizer_.output(state_segment)`.

ethanjli: `MessageSender::transform` will need to instantiate its own `Message input_message` object, and then set `input_message.payload = input_payload`, and hopefully we get copy-by-value (we should verify this by checking that the value of `input_payload` hasn't changed after calling `transform`).

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/Messages.tpp
```diff
@@ -98,15 +98,17 @@ MessageStatus MessageReceiver<Message, num_descriptors>::transform(
...
 
-template <typename Message, size_t num_descriptors>
+template <typename Message, typename TaggedUnion, size_t num_descriptors>
 template <size_t output_size>
-MessageStatus MessageSender<Message, num_descriptors>::transform(
-    const Message &input_message, Util::ByteVector<output_size> &output_buffer) const {
+MessageStatus MessageSender<Message, TaggedUnion, num_descriptors>::transform(
+    const TaggedUnion &input_payload, Util::ByteVector<output_size> &output_buffer) const {
+  Message input_message;
+  input_message.payload = input_payload;  // NOLINT(cppcoreguidelines-pro-type-union-access)
```
ethanjli: As discussed in meeting, we'll add a copy assignment operator to `TaggedUnion` where assigning the `value` member is suppressed with a `NOLINTNEXTLINE` (and we'll need to add a unit test or two of a TaggedUnion where the union type has constituent types of different sizes (e.g. a union of a `uint8_t` and a `uint16_t`)

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
```diff
@@ -53,7 +53,7 @@ static const auto state_sync_schedule = Util::make_array<const StateOutputSchedu
 using BackendMessage = Protocols::Message<
     Application::StateSegment,
     Application::MessageTypeValues,
-    Protocols::DatagramProps<
+    Protocols::CRCElementProps<
```
ethanjli: We caught a bug, where the CRCElement body size was defined based on Frame payload size rather than Datagram payload size! Here's the fix:
```
using CRCElementProps = Protocols::CRCElementProps<
    Driver::Serial::Backend::FrameProps::payload_max_size>;
using DatagramProps = Protocols::DatagramProps<
    Protocols::CRCElementProps<CRCElementProps::payload_max_size>;

using Message = Protocols::Message<
    Application::StateSegment,
    Application::MessageTypeValues,
    DatagramProps::payload_max_size>;
```

I will test this on the microcontroller after this fix.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/TaggedUnion.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+  GIVEN("A Taggedunion constructed with a union of uint8_t and uint16_t values and an enum class") {
+    union TestUnion {
+      uint8_t first;
+      uint16_t second;
+    };
+
+    using TestTaggedUnion = PF::Util::TaggedUnion<TestUnion, TestEnum>;
+
+    WHEN("Two instances of given taggedunion are copied by value") {
```
ethanjli: The two WHENs need to be differentiated from each other.

ethanjli (commented):
Here's my review of TaggedUnion.cpp

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
ethanjli: This PR seems to work fine on the microcontroller.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/Messages.tpp
```diff
@@ -19,8 +19,11 @@ template <typename TaggedUnion, typename MessageTypes, size_t max_size>
...
 MessageStatus Message<TaggedUnion, MessageTypes, max_size>::write(
     Util::ByteVector<output_size> &output_buffer,
-    const Util::ProtobufDescriptors<num_descriptors> &pb_protobuf_descriptors) const {
-  auto type = static_cast<uint8_t>(payload.tag);
+    const Util::ProtobufDescriptors<num_descriptors> &pb_protobuf_descriptors) {
+  static_assert(
+      Util::ByteVector<output_size>::max_size() >= max_size,
+      "Write method unavailable as output buffer is too small");
+  type = static_cast<uint8_t>(payload.tag);
   if (type > pb_protobuf_descriptors.size()) {
```
ethanjli: Should this be `>` or `>=`? In `parse` we changed `>` to `>=` because `>` was an off-by-one error. In either case, let's add a unit test for this edge case, for both `parse` and `write`.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+  // message class type member is 0
+
+    // setup:
+    // message class type member set to 5
+    // payload of parameters_request is set
+    // message descriptor array containing one element corresponding to Unknown message
+
+    WHEN("Write method is called on a message object with it's value of the payload tag greater than the size of the descriptor array")
+      THEN("The write method reports invalid type status")
+      THEN("After the write method is called, the type member remains unchanged")
```
ethanjli: Now that we removed `auto type = ...`, the type member should be changed on every `write` method call, right? (This applies to all tests of the write method)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("The payload.values field data remains unchanged")
+      THEN("The output buffer remains empty")
```
ethanjli: I think this should be "the output buffer remains unchanged"?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+
+    // setup:
+    // message class type member is unchanged
+    // payload tag is set to Unknown
+
+    WHEN("Write method is called on a message object where it's payload tag value is such that it corresponds to Unrecognized message descriptor in descriptor array")
```
ethanjli:
```suggestion
    WHEN("Write method is called on a message object whose payload.tag value corresponds to an Unrecognized message descriptor in the descriptor array")
```

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+
+    // setup:
+    // message class type member set to 5
+    // payload of parameters_request is set
+    // message descriptor array containing one element corresponding to Unknown message
+
+    WHEN("Write method is called on a message object with it's value of the payload tag greater than the size of the descriptor array")
+      THEN("The write method reports invalid type status")
+      THEN("After the write method is called, the type member remains unchanged")
+      THEN("The payload.values field data remains unchanged")
```
ethanjli, line 28: Does the `payload.tag` field remain unchanged too? (This applies to all tests of the write method)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("The output buffer remains empty")
+
+  GIVEN("A Message object default-constructed with StateSegment Taggedunion, default message types and a max size of 126 bytes as its class template parameters")
+    // setup:
+    // sensor_measurements.flow = 30;
+    // sensor_measurements.fio2 = 85;
+    // sensor_measurements.spo2 = 72;  
+    WHEN("The payload is a sensor measurements message and write method is called")
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, the type member is set to the value of the message payload tag")
```
ethanjli, line 73: Does the `payload.value` field remain unchanged too? (This applies to all tests of the write method)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+    // setup:
+    // parameters.fio2 = 60;
+    // parameters.mode = VentilationMode_hfnc;
+    // parameters.ventilating = true;
+    WHEN("The parameters message data is written")
+      THEN("The write method reports ok status")
+      THEN("After the write method is called, the type member is set to the value of the message payload tag")
+      THEN("The type field of the body's header correctly stores the value of message type obtained from the message payload tag")
+      THEN("The output buffer has an expected sequence of 10 bytes")
+
```
ethanjli, line 92: We're still missing a test for the `write` method returning `invalid_length`, which is one of the three possible error codes returned by this method. I believe there are two cases for this: 1) less than header size, and 2) greater than header size but less than the required size (e.g. for the sensor_measurements test which should generate a 16-byte output, setting max_size to 14 bytes).

If we're unable to generate tests which trigger `invalid_encoding`, we should add a comment to that effect in the test files, and explaining what we tried and why we think we won't be able to generate such a test.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: Did you find that parse sets the internal members, such as `tag`, even if the method returns `invalid_length`?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+  GIVEN("A Message object constructed with StateSegment Taggedunion, default message types and a max size of 252 bytes")
+    // setup:
+    // message class type member is unchanged
+    // payload is uninitialised
+
+    WHEN("An empty input buffer body is parsed")
+      THEN("The parse method reports invalid length status")
+      THEN("The payload Taggedunion tag is set to Unknown")
+      THEN("the payload.values field remains unchanged")
```
ethanjli, line 126: What happens to the `type` field after the method call? (This applies to all parse and MessageReceiver tests)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+
+    WHEN("A body with an encoded payload and 1 byte header of value equal to 0 is parsed")
```
ethanjli: It'd be helpful if you put the WHENs for the tests of the other ways we can get `invalid_type` return code above this WHEN, since this WHEN is constructed so that `MessageTypes::includes(type) == true && type < pb_protobuf_descriptors.size()`; this way, the order of tests matches the order of code in the implementation.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("The parse method reports invalid type status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the type field of the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to Unknown")
+      THEN("the payload.values field remains unchanged")
+
+    WHEN("A body with an empty payload and 1 byte header of value equal to the message descriptor array size is parsed")
+      THEN("The parse method reports invalid type status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the type field of the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to Unknown")
```
ethanjli: In this test of `invalid_type` resulting from `type >= pb_protobuf_descriptors.size()`, shouldn't `tag` be set to the value `pb_protobuf_descriptors.size() + 1` by the method?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("the payload.values field remains unchanged")
+
+    WHEN("A body with an empty payload and 1 byte header of value greater than the message descriptor array size is parsed")
+      THEN("The parse method reports invalid type status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the type field of the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to Unknown")
```
ethanjli: In this test of `invalid_type` resulting from `type >= pb_protobuf_descriptors.size()`, shouldn't `tag` be set to the value `pb_protobuf_descriptors.size() + 1` by the method?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: Where is the test of `invalid_type` resulting from `!MessageTypes::includes(type)`?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("the payload.values field remains unchanged")
+
+    WHEN("A body with a 1 byte header and payload of sensor_measurements message with invalid encoding is parsed")
```
ethanjli: What's the value of the 1-byte header?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: To rephrase: What's the value of the 1-byte header in each WHEN?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: It looks like this comment was marked as resolved without addressing either of the suggestions in this comment?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+/* test cases for parse functions for each message type
+    WHEN("A body with 1 byte header and a paylod of {Message type} message type is parsed")
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to {Message type}")
+      THEN("The message payload field values are as expected")
+      THEN("The input buffer is unchanged after parse")
+/*
+
+Scenario: The parse method correctly parses bodies when the payload and field members have data
```
ethanjli, line 181: Right now the test under this isn't about correctly parsing bodies, but about error-handling when parsing fails. However, we should also have a test about correctly parsing bodies when the payload and field members have data.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+  // setup:
+  // type = 5
+  // parameters_request.fio2 = 80;
+  // parameters_request.mode = VentilationMode_hfnc;
+  // parameters_request.ventilating = true;
+
+    WHEN("An empty input buffer body is parsed")
+      THEN("The parse method reports invalid length status")
+      THEN("After the parse method is called, The value assigned to the type member remains unchanged")
+      THEN("The payload Taggedunion tag is set to Unknown")
```
ethanjli: Are you sure that the value of the tag is changed to be unknown, instead of being left unchanged?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -0,0 +1,207 @@
...
+      THEN("The parse method reports ok status")
+      THEN("After the parse method is called, The value assigned to the type member is equal to the input_buffer body's header")
+      THEN("The payload Taggedunion tag is set to {Message type}")
+      THEN("The message payload field values are as expected")
+      THEN("The input buffer is unchanged after parse")
+/*
+
+Scenario: The parse method correctly parses bodies when the payload and field members have data
+
+  GIVEN("A Message object constructed with StateSegment Taggedunion, default message types and a max size of 252 bytes")
```
ethanjli: Was there a reason you left `payload.tag` uninitialized?

ethanjli (commented):
Here's my review of the write and parse methods of messages.txt. Since there are a bunch of comments, I will hold off on reviewing Roundtrip and MessageReceiver/MessageSender for now.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: > We're still missing a test for the `write` method returning `invalid_length`, which is one of the three possible error codes returned by this method. I believe there are two cases for this: 1) less than header size, and 2) greater than header size but less than the required size (e.g. for the sensor_measurements test which should generate a 16-byte output, setting max_size to 14 bytes).

added.

> If we're unable to generate tests which trigger `invalid_encoding`, we should add a comment to that effect in the test files, and explaining what we tried and why we think we won't be able to generate such a test.

will do.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: yes. fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: yes. fixed.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: So the tests where the body with `header` value greater than and equal to descriptor array size (8), both actually failed this test instead, as `MessageTypes` enum has enums till 7. Therefore I have re-structured these tests now

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Protocols/Messages.tpp
rohanpurohit: > Should this be > or >=? In parse we changed > to >= because > was an off-by-one error.

Yes, I think so too, updated. added for both with `invalid_type `result.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: I had updated everywhere else and forgot to update here. fixed now.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: No, it does not, it remains unchanged as the payload is uninitialized in this GIVEN, it is unknown before or after parse

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: I think the order should be fine now.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: Ya my bad, it is left unchanged.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: Since I use the `set` method to initialize the payload, I forgot to mention explicitly that the tag is set to something. I fixed it.

rohanpurohit:
@ethanjli I have resolved most of the review comments, maybe you can have a quick overview and check if I have missed anything. Other than that looks good to me. Please let me know thanks.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
```diff
@@ -64,144 +108,223 @@ Scenario: Message: The write method correctly generates an encoded body from the
...
       THEN("the payload.values field remains unchanged")
 
-    WHEN("A body with an encoded payload and 1 byte header of value equal to 0 is parsed")
+    WHEN("A body with an empty payload and 1-byte header of value equal to the message descriptor array size is parsed")
+    // setup:
+    // custom descriptor array of size 4
       THEN("The parse method reports invalid type status")
       THEN("After the parse method is called, The value assigned to the type member is equal to the type field of the input_buffer body's header")
-      THEN("The payload Taggedunion tag is set to Unknown")
+      THEN("The payload.tag field is set to parameters")
```
ethanjli: I don't understand the relationship between payload.tag == parameters and the size of the message descriptor array. Maybe there's some information in the test code which isn't described in the test description? (similar for the next WHEN with payload.tag == parameters_request)

ethanjli (commented):
I've skimmed over the changes to the tests for Message class, without aiming to understand the changes and new tests (and I've skipped MessageReceiver and MessageSender because they are basically thin wrapped around the Message class). As we agreed in the meeting, I am applying a less extensive depth of review for unit test definitions, and I am approving this PR because it has been through a few rounds of review already, with the understanding that these tests are prototype quality and will require a additional review later. I will assume that you've checked your each GIVEN-WHEN-THEN set with the specifications for firmware test patterns [from our Notion page](https://www.notion.so/pufferfish/Testing-13ce58d72a154b6883fbd57b71d37f76#3d4d5953522a477cb641469cd3edd798).

For records-keeping:

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
rohanpurohit: Right the information in the test code would be the header of the input_buffer which in this case is `0x04` (which is equal to the descriptor array size), the reason I created a different descriptor array rather than using the one from the backend for these two cases was that to return `invalid_type` on an array of 8 elements, `MessageTypes::includes(type)` would return `invalid_type` instead.  therefore created an array with 4 elements, and added the test cases for equal to and greater than the array size.

And also that's why the `payload.tag is set to parameters`  same is the case for the next WHEN, where the header is`0x05`. so I'm just asserting that the payload.tag parameter is updated.

rohanpurohit:

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/messages.txt
ethanjli: I see, so the context here is "THEN(The payload.tag field is set to parameters, _which corresponds to value 0x04_)"

ethanjli (approved):

rohanpurohit:
test

PR #300, opened 2021-03-26
Refactor COBS interface and add unit test cases

rohanpurohit:
This PR: 
Updates the COBS interface:

- pass ByteVectors instead of pointers to buffers.
- return IndexStatus code instead of decode and encode sizes.
- return early if buffer is empty while decoding
- return early if resize is not possible for output encoded buffer while encoding

Updates to COBSDecoder and COBSencoder classes:
- use static asserts to check for input and output sizes ( we are doing this to avoid writing test cases )
- return IndexStatus instead of FrameProps::OutputStatus

Add test cases for Util encode_cobs and decode_cobs

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
@@ -63,7 +63,7 @@ class BackendReceiver {
     available = 0,
     waiting,
     invalid_frame_length,
-    invalid_frame_decoding,
+    invalid_frame_encoding,
     invalid_crcelement_parse,
     invalid_crcelement_crc,
     invalid_datagram_parse,
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
@@ -63,7 +63,7 @@ class BackendReceiver {
     available = 0,
     waiting,
     invalid_frame_length,
-    invalid_cobs_decoding,
+    invalid_frame_decoding,
     invalid_crcelement_parse,
     invalid_crcelement_crc,
     invalid_datagram_parse,
@@ -105,7 +105,7 @@ class BackendSender {
     invalid_message_encoding,
     invalid_datagram_length,
     invalid_crcelement_length,
-    invalid_cobs_encoding,
+    invalid_frame_encoding,
     invalid_frame_length,
     invalid_return_code
   };
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
@@ -63,6 +63,7 @@ class BackendReceiver {
     available = 0,
     waiting,
     invalid_frame_length,
+    invalid_cobs_decoding,
     invalid_crcelement_parse,
     invalid_crcelement_crc,
     invalid_datagram_parse,
@@ -104,6 +105,7 @@ class BackendSender {
     invalid_message_encoding,
     invalid_datagram_length,
     invalid_crcelement_length,
+    invalid_cobs_encoding,
     invalid_frame_length,
     invalid_return_code
   };
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
@@ -23,13 +23,10 @@ IndexStatus COBSDecoder::transform(
       Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
       "COBSDecoder unavailable as the input buffer size is too large");
   static_assert(
-      Util::ByteVector<output_size>::max_size() <= FrameProps::payload_max_size,
-      "COBSDecoder unavailable as the output buffer size is too large");
+      Util::ByteVector<output_size>::max_size() >= FrameProps::payload_max_size,
+      "COBSDecoder unavailable as the output buffer size is too small");
 
-  if (Util::decode_cobs(input_buffer, output_buffer) != IndexStatus::ok) {
-    return IndexStatus::out_of_bounds;
-  }
-  return IndexStatus::ok;
+  return Util::decode_cobs(input_buffer, output_buffer);
 }
 
 // COBSEncoder
@@ -39,18 +36,13 @@ IndexStatus COBSEncoder::transform(
     const Util::ByteVector<input_size> &input_buffer,
     Util::ByteVector<output_size> &output_buffer) const {
   static_assert(
-      Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
-      "COBSDecoder unavailable as the input buffer size is too large");
+      Util::ByteVector<input_size>::max_size() <= FrameProps::payload_max_size,
+      "COBSEncoder unavailable as the input buffer size is too large");
+  static_assert(
+      Util::ByteVector<output_size>::max_size() >= FrameProps::encoded_max_size,
+      "COBSEncoder unavailable as the output buffer size is too small");
 
-  size_t encoded_size = Util::get_encoded_cobs_buffer_size(input_buffer.size());
-  if (output_buffer.max_size() < encoded_size) {
-    return IndexStatus::out_of_bounds;
-  }
-
-  if (Util::encode_cobs(input_buffer, output_buffer) != IndexStatus::ok) {
-    return IndexStatus::out_of_bounds;
-  }
-  return IndexStatus::ok;
+  return Util::encode_cobs(input_buffer, output_buffer);
 }
 
 }  // namespace Pufferfish::Driver::Serial::Backend
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
@@ -16,40 +16,41 @@ namespace Pufferfish::Driver::Serial::Backend {
 // COBSDecoder
 
 template <size_t input_size, size_t output_size>
-FrameProps::OutputStatus COBSDecoder::transform(
+IndexStatus COBSDecoder::transform(
     const Util::ByteVector<input_size> &input_buffer,
     Util::ByteVector<output_size> &output_buffer) const {
-  if (input_buffer.size() > FrameProps::payload_max_size + 1) {
-    return FrameProps::OutputStatus::invalid_length;
-  }
+  static_assert(
+      Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
+      "COBSDecoder unavailable as the input buffer size is too large");
+  static_assert(
+      Util::ByteVector<output_size>::max_size() <= FrameProps::payload_max_size,
+      "COBSDecoder unavailable as the output buffer size is too large");
 
-  if (output_buffer.max_size() < input_buffer.size()) {
-    return FrameProps::OutputStatus::invalid_length;
+  if (Util::decode_cobs(input_buffer, output_buffer) != IndexStatus::ok) {
+    return IndexStatus::out_of_bounds;
   }
-
-  output_buffer.resize(
-      Util::decode_cobs(input_buffer.buffer(), input_buffer.size(), output_buffer.buffer()));
-  return FrameProps::OutputStatus::ok;
+  return IndexStatus::ok;
 }
 
 // COBSEncoder
 
 template <size_t input_size, size_t output_size>
-FrameProps::OutputStatus COBSEncoder::transform(
+IndexStatus COBSEncoder::transform(
     const Util::ByteVector<input_size> &input_buffer,
     Util::ByteVector<output_size> &output_buffer) const {
-  if (input_buffer.size() > FrameProps::payload_max_size) {
-    return FrameProps::OutputStatus::invalid_length;
-  }
+  static_assert(
+      Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
+      "COBSDecoder unavailable as the input buffer size is too large");
 
   size_t encoded_size = Util::get_encoded_cobs_buffer_size(input_buffer.size());
   if (output_buffer.max_size() < encoded_size) {
-    return FrameProps::OutputStatus::invalid_length;
+    return IndexStatus::out_of_bounds;
   }
 
-  output_buffer.resize(
-      Util::encode_cobs(input_buffer.buffer(), input_buffer.size(), output_buffer.buffer()));
-  return FrameProps::OutputStatus::ok;
+  if (Util::encode_cobs(input_buffer, output_buffer) != IndexStatus::ok) {
+    return IndexStatus::out_of_bounds;
+  }
+  return IndexStatus::ok;
 }
 
 }  // namespace Pufferfish::Driver::Serial::Backend
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
@@ -61,6 +61,7 @@ IndexStatus encode_cobs(
 /// \brief Decode a COBS-encoded buffer.
 /// \param encodedBuffer A ByteVector to the \p encodedBuffer to decode.
 /// \param decodedBuffer The target ByteVector for the decoded bytes.
+/// The decoded buffer is resized to fit the decoded data, returns out_of_bounds otherwise
 /// \returns IndexStatus as ok/out_of_bounds
 template <size_t input_size, size_t output_size>
 IndexStatus decode_cobs(
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
@@ -51,21 +51,16 @@ namespace Pufferfish::Util {
 
 /// \brief Encode a byte buffer with the COBS encoder.
 /// \param buffer A ByteVector to the unencoded buffer to encode.
-/// \param size  The number of bytes in the \p buffer.
 /// \param encodedBuffer The ByteVector for the encoded bytes.
-/// \returns The number of bytes written to the \p encodedBuffer.
-/// \warning The encodedBuffer must have at least getEncodedBufferSize()
-///          allocated.
+/// \returns IndexStatus as ok/out_of_bounds
 template <size_t input_size, size_t output_size>
 IndexStatus encode_cobs(
     const Util::ByteVector<input_size> &buffer, Util::ByteVector<output_size> &encoded_buffer);
 
 /// \brief Decode a COBS-encoded buffer.
 /// \param encodedBuffer A ByteVector to the \p encodedBuffer to decode.
-/// \param size The number of bytes in the \p encodedBuffer.
 /// \param decodedBuffer The target ByteVector for the decoded bytes.
-/// \returns The number of bytes written to the \p decodedBuffer.
-/// \warning decodedBuffer must have a minimum capacity of size.
+/// \returns IndexStatus as ok/out_of_bounds
 template <size_t input_size, size_t output_size>
 IndexStatus decode_cobs(
     const Util::ByteVector<input_size> &encoded_buffer,
@@ -74,7 +69,7 @@ IndexStatus decode_cobs(
 /// \brief Get the maximum encoded buffer size for an unencoded buffer size.
 /// \param unencodedBufferSize The size of the buffer to be encoded.
 /// \returns the maximum size of the required encoded buffer.
-inline size_t get_encoded_cobs_buffer_size(size_t unencoded_buffer_size);
+constexpr size_t get_encoded_cobs_buffer_size(size_t unencoded_buffer_size);
 
 }  // namespace Pufferfish::Util
 
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
@@ -52,6 +52,7 @@ namespace Pufferfish::Util {
 /// \brief Encode a byte buffer with the COBS encoder.
 /// \param buffer A ByteVector to the unencoded buffer to encode.
 /// \param encodedBuffer The ByteVector for the encoded bytes.
+/// The encoded buffer is resized to fit the encoded data, returns out_of_bounds otherwise
 /// \returns IndexStatus as ok/out_of_bounds
 template <size_t input_size, size_t output_size>
 IndexStatus encode_cobs(
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
@@ -30,6 +30,7 @@
 #include <cstddef>
 #include <cstdint>
 
+#include "Pufferfish/Util/Vector.h"
 namespace Pufferfish::Util {
 
 /// \brief A Consistent Overhead Byte Stuffing (COBS) Encoder.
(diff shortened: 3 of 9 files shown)


rohanpurohit:
@ethanjli initial changes for updating the cobs interface.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
```diff
@@ -63,6 +63,7 @@ class BackendReceiver {
     available = 0,
     waiting,
     invalid_frame_length,
+    invalid_cobs_decoding,
```
ethanjli: I think of "encoding" like a property of the data, the same way that "length" is also a property of frames or "encoding" is a property of messages (see line 74), so I would suggest renaming this to either "invalid_cobs_encoding" or "invalid_frame_cobs" or "invalid_frame_encoding"

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Backend.h
```diff
@@ -104,6 +105,7 @@ class BackendSender {
     invalid_message_encoding,
     invalid_datagram_length,
     invalid_crcelement_length,
+    invalid_cobs_encoding,
```
ethanjli: Is it actually possible for the COBS encoder to generate an encoding error which would be distinct from a frame length error?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
```diff
@@ -16,40 +16,41 @@ namespace Pufferfish::Driver::Serial::Backend {
...
+IndexStatus COBSDecoder::transform(
     const Util::ByteVector<input_size> &input_buffer,
     Util::ByteVector<output_size> &output_buffer) const {
-  if (input_buffer.size() > FrameProps::payload_max_size + 1) {
-    return FrameProps::OutputStatus::invalid_length;
+  static_assert(
+      Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
+      "COBSDecoder unavailable as the input buffer size is too large");
+  static_assert(
+      Util::ByteVector<output_size>::max_size() <= FrameProps::payload_max_size,
```
ethanjli: You meant to use `>=` here, right? We should get an error if the provided output buffer is too small to hold the largest possible payload

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
```diff
@@ -16,40 +16,41 @@ namespace Pufferfish::Driver::Serial::Backend {
...
+      "COBSDecoder unavailable as the output buffer size is too large");
+
+  if (Util::decode_cobs(input_buffer, output_buffer) != IndexStatus::ok) {
+    return IndexStatus::out_of_bounds;
```
ethanjli: I think it'd be better to just do `return Util::decode_cobs(...)` if there's no code afterwards. When we do have code afterwards, it would be better to do `IndexStatus status = Util::decode_cobs(...)` and return status if `status` is not `ok`, so that we don't assume that `IndexStatus` will never have any other possible error statuses besides `ok` or `out_of_bounds`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
```diff
@@ -16,40 +16,41 @@ namespace Pufferfish::Driver::Serial::Backend {
...
 template <size_t input_size, size_t output_size>
-FrameProps::OutputStatus COBSEncoder::transform(
+IndexStatus COBSEncoder::transform(
     const Util::ByteVector<input_size> &input_buffer,
     Util::ByteVector<output_size> &output_buffer) const {
-  if (input_buffer.size() > FrameProps::payload_max_size) {
-    return FrameProps::OutputStatus::invalid_length;
-  }
+  static_assert(
+      Util::ByteVector<input_size>::max_size() <= FrameProps::encoded_max_size,
```
ethanjli: Shouldn't we be comparing the size of the input (the payload) against the max allowed size of the payload, rather than the max allowed size of the encoded result?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Backend/Frames.tpp
```diff
@@ -16,40 +16,41 @@ namespace Pufferfish::Driver::Serial::Backend {
...
+      "COBSDecoder unavailable as the input buffer size is too large");
 
   size_t encoded_size = Util::get_encoded_cobs_buffer_size(input_buffer.size());
   if (output_buffer.max_size() < encoded_size) {
-    return FrameProps::OutputStatus::invalid_length;
+    return IndexStatus::out_of_bounds;
   }
```
ethanjli: Let's instead just statically assert that `output_buffer.max_size()` should be big enough to hold the max encoded size.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
```diff
@@ -49,25 +50,32 @@ namespace Pufferfish::Util {
 /// \sa http://www.jacquesf.com/2011/03/consistent-overhead-byte-stuffing
 
 /// \brief Encode a byte buffer with the COBS encoder.
-/// \param buffer A pointer to the unencoded buffer to encode.
+/// \param buffer A ByteVector to the unencoded buffer to encode.
 /// \param size  The number of bytes in the \p buffer.
```
ethanjli: We are removing the size parameter from both functions.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
```diff
@@ -49,25 +50,32 @@ namespace Pufferfish::Util {
...
-/// \param encodedBuffer The buffer for the encoded bytes.
+/// \param encodedBuffer The ByteVector for the encoded bytes.
 /// \returns The number of bytes written to the \p encodedBuffer.
```
ethanjli: We are changing the return type of both functions.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
```diff
@@ -49,25 +50,32 @@ namespace Pufferfish::Util {
...
 /// \warning The encodedBuffer must have at least getEncodedBufferSize()
 ///          allocated.
```
ethanjli: Since we're now passing in ByteVectors, we can do the size checks instead of assuming that the output buffer has at least a certain size allocated (or else we go out-of-bounds in memory), and we can return an error if this condition is not met, so we can remove this warning from both functions.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
```diff
@@ -49,25 +50,32 @@ namespace Pufferfish::Util {
...
+template <size_t input_size, size_t output_size>
+IndexStatus decode_cobs(
+    const Util::ByteVector<input_size> &encoded_buffer,
+    Util::ByteVector<output_size> &decoded_buffer);
 
 /// \brief Get the maximum encoded buffer size for an unencoded buffer size.
 /// \param unencodedBufferSize The size of the buffer to be encoded.
 /// \returns the maximum size of the required encoded buffer.
-size_t get_encoded_cobs_buffer_size(size_t unencoded_buffer_size);
+inline size_t get_encoded_cobs_buffer_size(size_t unencoded_buffer_size);
```
ethanjli: If we need this to be inline, we might as well just make it constexpr (which also implies inline).

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.tpp
```diff
@@ -31,13 +31,19 @@ namespace Pufferfish::Util {
...
+IndexStatus encode_cobs(
+    const Util::ByteVector<input_size> &buffer, Util::ByteVector<output_size> &encoded_buffer) {
   size_t read_index = 0;
   size_t write_index = 1;
   size_t code_index = 0;
   uint8_t code = 1;
 
-  while (read_index < size) {
+  if (encoded_buffer.empty()) {
+    encoded_buffer.resize(buffer.size() + 1);
```
ethanjli: Shouldn't we always do this resize? And shouldn't we use `get_cobs_encoded_buffer_size` to tell us the size to resize to?

We should return an error code if the resize fails and returns an error.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.tpp
```diff
@@ -57,39 +63,42 @@ size_t encode_cobs(const uint8_t *buffer, size_t size, uint8_t *encoded_buffer)
...
+    if (read_index + code > encoded_buffer.size() && code != 1) {
+      return IndexStatus::out_of_bounds;
     }
 
     read_index++;
 
     for (uint8_t i = 1; i < code; i++) {
-      decoded_buffer[write_index++] = encoded_buffer[read_index++];
+      uint8_t byte = encoded_buffer[read_index++];
+      decoded_buffer.push_back(byte);
```
ethanjli, line 91: Do we need to resize `decoded_buffer` to size 0 at the start of this function?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.tpp
```diff
@@ -57,39 +63,42 @@ size_t encode_cobs(const uint8_t *buffer, size_t size, uint8_t *encoded_buffer)
...
     }
 
-    if (code != max_block_size + 1 && read_index != size) {
-      decoded_buffer[write_index++] = '\0';
+    if (code != max_block_size + 1 && read_index != encoded_buffer.size()) {
+      decoded_buffer.push_back(0x00);
```
ethanjli, line 95: The deleted line incremented `write_index` after adding a null byte to `decoded_buffer`, do we not need to do the same thing in the new version?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,63 @@
...
+
+   WHEN("The COBS::decode method is called on a encoded buffer of capacity 256 bytes whose original buffer was of capacity 254 with null byte at the end")   
+    THEN("The decoded size is 254")    
+    THEN("The decoded buffer has expected sequence of 254 bytes with last byte as null byte")    
+
+   WHEN("The COBS::decode method is called on a encoded buffer of capacity 257 bytes whose original buffer was of capacity 255 with no null bytes") 
+    THEN("The decoded size is 255")
+    THEN("The decoded buffer has expected sequence of 255 bytes as '0xff 0x71 0xcf ......0x02 0xa2 0xd2')
+
+Scenario: The Util encode_cobs method correctly encodes buffers
```
ethanjli: We will need to add tests checking what happens if our encoded_buffer is too small to fit the encoded result.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,63 @@
+Scenario: The cobs::decode method correctly decodes encoded buffers
+
+ GIVEN("The Util COBS::decode function")
```
ethanjli, line 71: The tests for `COBS::decode` need to be updated so that the first THEN checks the return status code, like what you did for `COBS::encode`. In general, both functions will need to have tests triggering all possible distinct ways that error code returns can be triggered

ethanjli (commented):
Great start! I reviewed the changes to `Core/` files, as well as `cobs.txt`. Based on the changes you made, it looks like this will make our COBS interface less brittle (in terms of being able to detect errors and stop early and report them)

rohanpurohit:
@ethanjli This PR is ready for review. the failing test case is now resolved, we can discuss that in the meeting, both encode and decode work fine.

Changes from last review: 
- returned error code if resize fails during encoding
- changed invalid_frame_decoding to invalid_frame_encoding
- added more test cases.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,131 @@
...
+
+  WHEN("The COBS::decode method is called on a buffer containing a encoded sensor measurements message payload")
+    THEN("The decode_cobs method reports ok status")
+    THEN("The decoded buffer is as expected")
+
+  WHEN("The COBS::decode method is called on a buffer that contains these bytes '0x03, 0x9c, 0xb8, 0x03, 0xbe, 0xce'")
+    THEN("The decode_cobs method reports ok status")
+    THEN("The decoded buffer is as expected '\x9c\xb8\x00\xbe\xce' ")
+
+  WHEN("The COBS::decode method is called on a encoded buffer of capacity 254 bytes whose original buffer was of capacity 253 with no null bytes")
```
ethanjli: Let's figure out how to state this WHEN and the next two WHENs more clearly and unambiguously.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,131 @@
+Scenario: The cobs::decode method correctly decodes encoded buffers
```
ethanjli: Since COBS.h defines the encode function before the decode function, the encode tests should come before the decode tests. Also, these are functions and not methods (as reflected in the GIVEN), so the scenario description should be consistent with that.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,131 @@
...
+    THEN("The decoded buffer is empty")
+     
+  WHEN("The COBS::decode method is called on a buffer that contains these bytes '0x05 0x02 0x03' ")    
+    THEN("The decode_cobs method reports out_of_bounds status")
+    THEN("The decoded buffer is empty")
+
+  WHEN("The COBS::decode method is called on a buffer that contains these bytes '0x05 0x02 0xff' ")    
+    THEN("The decode_cobs method reports out_of_bounds status")
+    THEN("The decoded buffer is empty")
+
```
ethanjli: We need a test that captures the essence of the test we had been troubleshooting last call - since these three above WHENs the first byte is too big compared to the length of the input buffer, I don't think these WHENs cover the case we were looking at last time.

ethanjli (commented):
Everything looks generally good, some minor changes are suggested in this review of everything in the PR (though maybe a few more things might come up during our meeting today)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
ethanjli: `0x02 0x02 0xff` has been added as a simplified one, let's also add `0xff 0x10 ... 0x10 0x02 0x10 0x10` as the exact example which we were getting confused by.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
ethanjli: We'll use "is called on a 257-byte buffer filled with the manual encoding of a (e.g.) 255 non-null-byte payload"

ethanjli:
I'll test this on the microcontroller prior to merging. For records-keeping:

ethanjli:
I tested this on the microcontroller and found a COBS-related regression: now many of the frames being sent by the microcontroller have invalid COBS encodings. For example, the first COBS body sent is valid (7-byte input of `0x92 0xfd 0x4b 0xfa 0x00 0x01 0x02`, 8-byte encoded output of `0x05 0x92 0xfd 0x4b 0xfa 0x03 0x01 0x02`), but the second COBS body sent has an input of 7 bytes (`0x11 0x29 0x3e 0x6c 0x01 0x01 0x04`) and an encoded output of 9 bytes (`0x08 0x11 0x1d 0x3e 0x6c 0x01 0x01 0x04 0x00`), which is clearly an incorrect length. Can you see if you can reproduce this behavior in unit tests? e.g. if calling `COBS::encode('0x11 0x29 0x3e 0x6c 0x01 0x01 0x04')` gives a 9-byte buffer as output, this suggests that there is an issue with the algorithm; otherwise, we've probably revealed or introduced an issue in how `BackendSender` or `FrameSender` is either resetting/clearing or not resetting/clearing `input_buffer` and `output_buffer`.

rohanpurohit:
> I tested this on the microcontroller and found a COBS-related regression: now many of the frames being sent by the microcontroller have invalid COBS encodings. For example, the first COBS encoder result is valid (7-byte input of `0x92 0xfd 0x4b 0xfa 0x00 0x01 0x02`, 8-byte encoded output of `0x05 0x92 0xfd 0x4b 0xfa 0x03 0x01 0x02`), but the second COBS body sent has an input of 7 bytes (`0x11 0x1d 0x3e 0x6c 0x01 0x01 0x04`) and an encoded output of 9 bytes (`0x08 0x11 0x1d 0x3e 0x6c 0x01 0x01 0x04 0x00`), which is clearly an incorrect length. According to the debugger, what happens in the second COBS encoder function call is that we pass it the input buffer as listed above and a 9-element output buffer which was already dirty before the function call (with data `0x05 0x92 0xfd 0x4b 0xfa 0x03 0x01 0x02 0x00`). The output buffer should be getting resized down to 8 elements, but it doesn't seem to be getting resized down.
> 
> Can you see if you can reproduce this behavior in unit tests? e.g. if calling `COBS::encode('0x11 0x29 0x3e 0x6c 0x01 0x01 0x04')`, where the `output_buffer` before the function call is empty or has 9 bytes of arbitrary data inside, results in a 9-byte buffer after the function call, this suggests that there is an issue with the algorithm; otherwise, we've probably revealed or introduced an issue in how `BackendSender` or `FrameSender` is either resetting/clearing or not resetting/clearing `input_buffer` and `output_buffer`.
> 
> Since this issue escaped both unit tests on `Util/COBS.h` and code review, I suspect that our change to the COBS functions and calling code has broken the contract around the `output_buffer` parameter - for example, if the new version of the COBS encoder function isn't correctly resizing `output_buffer`s for whatever reason, now that we've stopped resizing buffers in the calling code. But if there's any relationship to the external interface of the COBS functions, we should add a unit test to reproduce and assert the correct behavior or point out an incorrect way to use the functions. We should also make sure that, for functions with buffers as output parameters, we have unit tests checking whether the value of the output parameter after the function call depends on the value of the output parameter before the function call.

I will try to reproduce that.

ethanjli:
I tested this on the microcontroller and it works fine. I deleted the previous comment which you've quoted, and I realized that issue was because my git GUI client lied to me that it had the most recent commits on this branch when it was actually stuck on 8ec1a68, and as a result I found that the COBS encoder function was behaving incorrectly because it wasn't resizing the output buffer (passed in as 9 bytes) down to 8 bytes (the appropriate size for a 7-byte input buffer), which is the exact problem pointed out in https://github.com/pez-globo/pufferfish-software/pull/300#discussion_r601961080 and resolved in 44f6e96.

Since this resizing behavior is an important part of the contract of the COBS util functions, we should add a note in the code comment documentation for the functions about how `output_buffer` is supposed to be modified - it should be resized to the appropriate size for the input, and if it can't be properly resized then it returns a bounds error. And we should make sure we have unit test cases to check that the results we get from the encode and decode functions are independent of the data (and buffer length) which had been in `output_buffer` before the encode/decode functions were called. This is implied in the specification for Firmware Test Pattern 1's WHEN: `WHEN: the function is called with ... as input parameters and ... as output parameters, listed from left to right in order of parameters` - so we need to specify what the values given in the output parameters, which lets us determine in review if we have sufficient coverage of the possible values which can be given in the output parameters. Should we add some sort of clarification to the test pattern's specification about coverage of the input values of output parameters?

rohanpurohit:

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.h
```diff
@@ -49,25 +50,28 @@ namespace Pufferfish::Util {
...
 
 /// \brief Decode a COBS-encoded buffer.
-/// \param encodedBuffer A pointer to the \p encodedBuffer to decode.
-/// \param size The number of bytes in the \p encodedBuffer.
-/// \param decodedBuffer The target buffer for the decoded bytes.
-/// \returns The number of bytes written to the \p decodedBuffer.
-/// \warning decodedBuffer must have a minimum capacity of size.
-size_t decode_cobs(const uint8_t *encoded_buffer, size_t size, uint8_t *decoded_buffer);
+/// \param encodedBuffer A ByteVector to the \p encodedBuffer to decode.
+/// \param decodedBuffer The target ByteVector for the decoded bytes.
```
ethanjli, line 63: I believe the decoded buffer should be resized to fit the decoded data, and return out_of_bounds otherwise.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -16,7 +16,11 @@ Scenario: The Util encode_cobs function correctly encodes buffers
 
   WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33'")
     THEN("The encode_cobs function reports ok status")
-    THEN("The encoded buffer is as expected '\x03\x11\x22\x02\x33'") {
+    THEN("The encoded buffer is as expected '\x03\x11\x22\x02\x33'")
+
+  WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33' and a non-empty output_buffer is passed")
```
ethanjli: Regression coverage would be stronger if we had a test case where output_buffer needs to be resized up and a test case where output_buffer needs to be resized down, since we can't guarantee that the implementation will always handle both cases the same way. If you don't want to do this, let's at least specify the length of output_buffer (e.g. if we gave a non-empty output buffer of size 5 we might not catch any bugs preventing resizing, since the target size is 5 bytes anyways; so the test should specify that the length is not 5, or it should specify the exact length passed).

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -0,0 +1,142 @@
...
+
+  WHEN("The cobs::encode is called on a buffer that contains these bytes '0x00 0x00'")
+    THEN("The encode_cobs function reports ok status")
+    THEN("The encoded buffer is as expected '\x01\x01\x01' ")
+
+  WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33'")
+    THEN("The encode_cobs function reports ok status")
+    THEN("The encoded buffer is as expected '\x03\x11\x22\x02\x33'")
+
+  WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33' and a non-empty output_buffer is passed")
```
ethanjli: I see that you added a test on the independence of cobs::encode vs. the state of output_buffer before the function call, but I don't see a comparable test for cobs::decode yet.

ethanjli (commented):
The changes look good, but they've only been made for encode_cobs and not for decode_cobs. Once these are addressed, let's merge this PR.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
rohanpurohit: completely forgot about decode. will add now

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
rohanpurohit: the current case is resizing up as it's filled with 2-bytes , I'll add for down

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/cobs.txt
```diff
@@ -18,10 +18,14 @@ Scenario: The Util encode_cobs function correctly encodes buffers
     THEN("The encode_cobs function reports ok status")
     THEN("The encoded buffer is as expected '\x03\x11\x22\x02\x33'")
 
-  WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33' and a non-empty output_buffer is passed")
+  WHEN("The cobs::encode function is called on a buffer that contains these bytes '0x11 0x22 0x00 0x33' and a output_buffer filled with 2-bytes is passed")
```
ethanjli: When you say that the output_buffer is "filled", is it actually full (which would imply that its capacity is 2 bytes? Or is it just supposed to be partially filled? This applies to the other WHENs changed in this commit.

rohanpurohit: It is partially filled, so a 254-byte buffer is filled with 2-bytes.

ethanjli (approved):

PR #301, opened 2021-03-26
Generate settings & alarm limits change events in backend simulator

ethanjli:
This PR:

Updates protocol buffer message definitions:
- Reorders fields so that HFNC-related fields come first
- **Updates the ventilation modes enum**, including adding and removing some modes
- Organizes the log event codes enum into sections, and adds values for control and alarm limits events
- Defines a new log event type enum value for alarm limits
- **Adds more old/new fields for different data types into log events, instead of just having `old_value` and `new_value`**, and reorders fields

Updates the frontend:
- Reorders the cases of switch statements for ventilation mode, to match the order of values in the ventilation modes protobuf enum
- Updates the modes page layout to show the new modes and to remove modes which will not be implemented (this change is untested, because I have no way of viewing the modes page)
- In the Event Log, renames the Alarm column to the Event column; makes patient alarms be shown in red rather than green; changes the format of the date/time column to display the date first, since date has a higher priority than time in the column's sort order; **adds support for alarm limits change events**
- In the Event Log, fixes capitalization for event labels and details; makes details strings more brief; adds missing units into details strings; adds a special detail string for ventilation started/stopped events; removes undesirable space between value and unit when the unit is a percentage; reorganizes event type definitions to match the organization of the log event codes enum
- Fixes a misspelling of "Cannula"

Updates the backend:
- Splits off events log management functionality from alarms management
- **Generates parameter change log events for FiO2 and Flow Rate**
- **Generates alarm limits change log events for FiO2, SpO2, and Heart Rate**
- **Adds backend simulation and alarms for heart rate**
- **Disables many unit tests in `backend/tests/unit/protocols/test_file.py`** because I didn't write those tests and I don't have instructions for how to update the test examples for the new protobuf field orderings. This will need to be fixed in a future PR, and it should be fixed before any changes are made to the filesystem-related protocols. (@Tanmay06 do you have any personal notes about how you generated these examples and/or how I should go about updating them? If not, @rohanpurohit this might be a new task for you, though lower priority than firmware-related tests)


ethanjli:
@Sudhir-dev Can you try running the frontend with the backend simulator to see if you can catch any issues or suspicious behaviors which might've slipped past me? And can you also review the changes to the protobuf definitions and the events log-related frontend files, and share any observations or questions you might have?

ethanjli:
For records-keeping:

Sudhir-dev (approved):
Looks good to me.

PR #302, opened 2021-04-01
Driver/I2C: switch to read_ntoh instead of parse_network_order

rohanpurohit:
This PR proposes switching to read_ntoh instead of parse_network_order:
since the implementation of parse_network_order could have lead to a buffer overflow if the correct size wasn't passed


rohanpurohit:
@ethanjli This PR is ready for review.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Endian.h
```diff
@@ -14,6 +14,23 @@
 
 namespace Pufferfish::Util {
 
+constexpr void read_ntoh(const uint8_t *network_endian, int16_t &host_endian) noexcept {
```
ethanjli, line 17: For the `int16_t` and `int64_t` implementations, did you copy-paste+edit the implementations of the `uint16_t` and `uint64_t` versions of `read_ntoh` because you found it was infeasible (i.e. not allowed by the compiler) to make the `int16_t` and `int64_t` versions just call the `uint16_t` and `uint64_t` versions?

rohanpurohit: there is no `int64_t `implementation that is the `uint64_t `overloaded function, for the `int16_t  ` version, ya the compiler/clang-tidy complained if I called uint16_t version by static_cast

ethanjli (commented):
The code looks fine, and I don't see any regressions in the SFM3019 sensor readings when I test this PR on the microcontroller. For records-keeping:

rohanpurohit:

ethanjli (approved):

ethanjli:
As discussed in our meeting: the naming of `read_ntoh` is kind of confusing. The naming will be clearer (and consistent with how Python structs require us to specify whether a field is bigend or littleend rather than network order or host order) if we do something like:

```
read_bigend(const uint8_t *big_endian, uint16_t &result):
    result = set_byte<0, uint16_t>(network_endian[1]) + set_byte<1, uint16_t>(network_endian[0])
```

PR #303, opened 2021-04-01
Improve COBS & add unit test cases for Backend/Frames

rohanpurohit:
This PR includes:
- Update COBS interface to fix resizing and change encode_cobs algorithm to account for edge cases for eg: a 254-byte buffer and buffers of size greater than 255.
- rename the read_ntoh function to read_bigend for better understanding.
- add Unit test cases for backend/Frames.h


rohanpurohit:
@ethanjli PR is ready for review. thanks

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Driver/Serial/Backend/frames.txt
```diff
@@ -0,0 +1,138 @@
...
+   THEN("The input method reports ok status for all 255 non-delimiter bytes and invalid_length for the 256th byte")
+   THEN("The output method reports invalid_length status")
+   THEN("The output buffer remains empty")
+
+  WHEN("128 non-delimiter frame body followed by a delimiter are passed as input, and output is called after that")
+   THEN("The input method reports ok status for all 128 non-delimiter bytes and output_ready for 129th byte")
+   THEN("The output method reports ok status")
+   THEN("The output buffer has an expected sequence of 128 bytes")
+
+  WHEN("256-byte frame body with a delimiter at the 129th and 256th-byte position is passed as input, and output is called after that")
```
ethanjli: WHEN 256 bytes are passed as input, where the input's delimiter bytes are at positions 129 and 256, and then output is called afterwards

ethanjli (commented):
Tests to add:

- Correct resetting of FrameReceiver behavior after calling output from a valid input (like in chunks.txt)
- Correct resetting of FrameReceiver behavior after calling output from an input error (like in chunks.txt); cover input errors triggered by ChunkSplitter and errors triggered by COBSDecoder

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.tpp
```diff
@@ -39,25 +39,37 @@ IndexStatus encode_cobs(
...
-  size_t encoded_buffer_size = 0;
-
-  if (buffer.size() == max_block_size) {
-    for (size_t i = 0; i < max_block_size; i++) {
-      if (buffer[i] == 0) {
-        encoded_buffer_size = max_block_size + 1;
-        break;
-      }
-      encoded_buffer_size = max_block_size + 2;
+  // calculate the exact encoded buffer size
```
ethanjli: As discussed, let's decompose the size-computation logic out into a separate function.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/COBS.tpp
```diff
@@ -115,7 +116,7 @@ constexpr size_t get_encoded_cobs_buffer_size(
       write_index++;
     }
     if (read_index > unencoded_buffer.size()) {
```
ethanjli: What should happen if `read_index == unencoded_buffer.size()`? Right now it looks like then line 121 would try to do a memory access at `unencoded_buffer[unencoded_buffer.size()]`. Is that within the bounds of `unencoded_buffer`? (I think the answer is no, but I always get confused about potential off-by-one cases)

(also, would it be possible to generate a test case where this happens?)

rohanpurohit: I think read_index is always less than `unencoded_buffer.size()` as it won't enter the while loop if it's equal to it. So in a way we are implicitly doing the bounds check, then the question is do we need to explicitly catch `out_of_bounds` ?

ethanjli: Oops!!! I had been thinking of this function as reading encoded data, when clearly it's reading pre-encoded data. The only way we read data is checking if a byte is zero, and deciding based on that whether to increment some (index) counters. So the bounds check of the while loop is sufficient; invalid data has no way to force an out-of-bounds read.

rohanpurohit:

ethanjli:
Let's merge the latest version of develop back into this branch, and then try running the firmware with these changes to ensure that there are no regressions or subtle merge conflicts. Afterwards, I'll mark the formal approval on Github.

ethanjli (approved):
From light testing, it seems to work fine without any issues.

PR #304, opened 2021-04-01
Frontend UI bugs

Sudhir-dev:
Addressing following Bug fixes mentioned in PR #296 

> * The upper-right close button of the Events Log modal is not aligned with the upper-right close button of the breadcrumbs modal; and the titles of the two modals aren't consistent in position or text styling (bold vs. non-bold)
> * The dotted line border around the table for the Events Log modal is awkward - the left and right edges are right next to the modal. Maybe it'd be better to completely remove the dotted edge border for the Events Log modal?
> * When I switch between the Events Log modal and the Active Alarms modal, the sizes of the content container seem to be different, so the elements which are the same in both modals move. Also, the alarm mute button moves when I switch between the two.
> * When I have 4 events in the Events Log modal, a vertical scrollbar appears; there should never be any scrollbar in the modals.
> * The names of the second-layer breadcrumbs don't match the names specified in the Adobe XD design. For example, what the Adobe XD design specifies as "Flow Rate Settings" for the breadcrumb title is currently listed as "Set New"; and what the Adobe XD design specifies as "SpO2 Alarms" is currently listed as "Alarms". I would suggest following the Adobe XD design's breadcrumb naming so that different screens on the modal will have different breadcrumb path names


Sudhir-dev:

ethanjli:
As discussed in the meeting for this PR:

- We will globally replace the "x" icon of the close button with the "back" button used elsewhere in the frontend
- The "Events Log" button from the topbar should always open the "Events Log" view, rather than the last view selected

For one or two separate PRs:
- It'd be great to have absolute positioning of the pagination controls in the Events Log modal. It's fine if they're absolutely positioned to the lower left or the lower right, if one is simpler
- The Events Log should pull alarm limits from the log event objects, rather than from the store

Sudhir-dev:
@ethanjli Updated PR with fixes for following points
 
> * We will globally replace the "x" icon of the close button with the "back" button used elsewhere in the frontend
> * The "Events Log" button from the topbar should always open the "Events Log" view, rather than the last view selected

ethanjli (approved):
Looks great! I just realized that I had forgotten to write down another issue to fix: making the height of the table rows in the Event Log constant, so that it's the same whether or not the "Settings" button is shown in a row. We could fix this in this PR, or in the next one; I'll approve this PR for merging in case you want to fix it in the next PR.

PR #305, opened 2021-04-05
Frontend UI fixes

Sudhir-dev:
This PR addresses following comments mentioned in PR #304 

> For one or two separate PRs:
> 
> * It'd be great to have absolute positioning of the pagination controls in the Events Log modal. It's fine if they're absolutely positioned to the lower left or the lower right, if one is simpler
> * The Events Log should pull alarm limits from the log event objects, rather than from the store

> Looks great! I just realized that I had forgotten to write down another issue to fix: making the height of the table rows in the Event Log constant, so that it's the same whether or not the "Settings" button is shown in a row. We could fix this in this PR, or in the next one; I'll approve this PR for merging in case you want to fix it in the next PR.


ethanjli:
Looks great! For records-keeping:

Sudhir-dev:

ethanjli (approved):

PR #306, opened 2021-04-06
add unit test cases for states

rohanpurohit:
This PR includes:
- test cases for Protocols/states.h


rohanpurohit:
@ethanjli the GIVEN WHEN'S THEN'S have been modified for all test cases, therefore requires one review, if there are major changes then we can have one more. In terms of test cases we have discussed these quite a few times and I think they should be enough, please let me know. Thanks

ethanjli (commented):
From looking over the structure of the tests (but not looking in depth at coverage, wording, or clarity), this looks fine to me. To make the intent of each WHEN clearer, I would recommend grouping the WHENs into more, smaller scenarios - e.g. one scenario is about the behavior of advancing schedule entries when long amounts of time (relative to the schedule entry timeouts) pass between clock updates, one scenario is about the behavior when very short amounts of time (relative to schedule entry timeouts) pass between clock updates, one scenario is about cycling back from the end of a schedule to the next scenario, maybe there are some scenarios about what happens when the exact timeout of a schedule entry is advanced by a clock update, etc. For example, your "Scenario: The States output method correctly updates output StateSegment tag and value fields on an output schedule array of multiple message types" scenario looks like it overlaps with your "Scenario: The States output method behaves correctly when an exceptionally long time passes between subsequent entries" scenario in the kinds of things it's trying to test.

I think this PR is fine to merge as prototype-quality test specs with the understanding that the tests will be iterated on later as part of software quality assurance. It's up to you whether/how to reorganize the scenarios in this PR. For records-keeping:

rohanpurohit:
I will restructure the scenarios in future iterations.

ethanjli (approved):

PR #316, opened 2021-04-16
deploy: split up deploy.sh (#314)

### LINKED ISSUE #314: Split up deploy.sh into install/setup and locking up the system
Currently, running the deploy.sh will do the following:
- install all the dependencies.
- setup all the required services.
- lockout the file system by generating hashes for all the files to avoid tampering.
- locks and removes sudo permissions from pi users.
- disables all network-related services.

If there is any error while installing/setting up the services, the dev is locked out of the pi and needs to reflash the raspberry pi sd card and re-do the process from the beginning.
therefore, we propose to split up the process of install/setup and locking the user out of the system into two different scripts, one as deploy/install/setup and the second one as lock.sh

rohanpurohit:
This PR includes-
- split up deploy.sh into install, deploy and lock
- change backend_file in helper.sh from simulator to application (so now deployment installs the application backend rather than the simulator backend)
- update deployment README


ethanjli (commented):
Looks good! One thing we could iterate on is suppressing the "Reboot required for changes to take effect. Please restart!" message when it is not desirable. Specifically, when install.sh and deploy.sh and maybe lock.sh are called from within a pipeline (such as deploy_full.sh), we want to suppress those messages (though suppression may not be needed for lock.sh). You could use the pattern included in the end of install.sh to implement suppression of the reboot message - note that it also runs the reboot command for the operator if confirmation is given, which is nice.

I would suggest that, instead of using "deploy" as the argument to install.sh, deploy.sh, and lock.sh, you could use "pipeline".

deploy/deploy_full.sh
```diff
@@ -14,12 +14,12 @@ function exit_script {
 }
 
 # Install required components
-$script_path/install.sh "deploy" || exit_script "Development components installation failed"
+$script_path/install.sh "pipeline" || exit_script "Development components installation failed"
 
 # Deploy services
-$script_path/deploy.sh "deploy" || exit_script "Development of services failed"
+$script_path/deploy.sh "pipeline" || exit_script "Development of services failed"
```
ethanjli: I think you meant to write "Deployment of services" rather than "Development of services"

deploy/deploy.sh
```diff
@@ -37,4 +37,12 @@ $script_path/scripts/live_usb.sh || exit_script "USB Mass Storage setup failed"
 # Modify eeprom to setup watchdog to handle Raspberry Pi failures and crashes
 $script_path/scripts/setup_watchdog.sh || exit_script "Watchdog setup failed"
 
-echo -e "Reboot required for changes to take effect. Please restart!"
+if [[ "$@" != "pipeline" ]]
+then
+    echo -n "Reboot required for changes to take effect. Do you want to reboot now? [y/N]: "
```
ethanjli: Let's use this PR to fix the message of possible inputs, from `[y/N]` to `[y/n]` - in all the scripts.

deploy/deploy_full.sh
```diff
@@ -14,12 +14,12 @@ function exit_script {
...
 
 # Deploy lock services
-$script_path/lock.sh "deploy" || exit_script "Development of lock services failed"
+$script_path/lock.sh || exit_script "Development of lock services failed"
```
ethanjli: Is there a particular reason we don't have a "pipeline" flag for conditional printing of a reboot message in `lock.sh`?

ethanjli (changes requested):
Just some minor changes, and then we're ready to merge (EDIT: I submitted this review as "Request Changes" but actually I should've submitted this review as a regular "Comment"). I haven't tested (and won't test) the functionality of this PR, because it's simple enough for just a review of the code - I'll assume that it works correctly for you when you run it.

For records-keeping:

deploy/deploy_full.sh
rohanpurohit: Done.

deploy/deploy_full.sh
rohanpurohit: seems like a derp, we should suppress for lock as well and reboot after all 3 are complete.

deploy/deploy.sh
rohanpurohit: done

rohanpurohit:

ethanjli (approved):

rohanpurohit:
the changes were tested on a headless raspberry pi 4b!

PR #317, opened 2021-04-17
messages: fix failing test cases

rohanpurohit:
This PR fixes the failing assertions in messages.cpp,
These tests should have failed and I fixed them
These were leftover from a previous iteration where the message types structs were not initialized to 0 and therefore the output buffer that was generated was incorrect. Although it's still unclear why GitHub actions passed in the PR whereas it should have failed for each commit, certainly might be worth looking at either TestCatch2 or GitHub actions behavior for the same. Another interesting observation I had was the test cases on my computer always passed and failed only after removing the `std::string` operator, whereas they should have failed each time, this might be something related to my computer or how I was running the tests and not the library itself as I had the behavior mirroring the GitHub actions and merged commit on my laptop. This is definitely isolated to the test cases in this file and not the test cases that are merged previously.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Messages.cpp
+++ b/firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Protocols/Messages.cpp
@@ -25,8 +25,6 @@ namespace PF = Pufferfish;
 namespace BE = PF::Driver::Serial::Backend;
 using namespace std::string_literals;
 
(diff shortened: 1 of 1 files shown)


ethanjli:
The tests run fine on my computer now. In the PR description, can you describe what was behaving incorrectly, why it was behaving incorrectly, and what you did to make it behave correctly?

For records-keeping:

rohanpurohit:

ethanjli (approved):

PR #318, opened 2021-04-19
display version number in frontend (#258)

### LINKED ISSUE #258: Display of version number/hash in frontend

rohanpurohit:
This PR:
- adds functionality to display either commit-hash or tag depending on the latest commit on the current branch

### DIFF
--- a/deploy/scripts/helper.sh
+++ b/deploy/scripts/helper.sh
@@ -13,17 +13,6 @@ backend_dir=$script_dir/../../backend/ventserver
 backend_file=$(realpath $backend_dir)/simulator.py
 frontend_dir=$script_dir/../../frontend
 
-# query git for the latest commit
-latest_commit_hash=$(git rev-parse --short --verify HEAD)
-tag=$(git tag --contains $latest_commit_hash)
-
-if [ -z $tag ]
-then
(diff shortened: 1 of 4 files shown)


ethanjli:
I like how you modified the yarn commands - very clean! It works for me, for both `yarn start` and `yarn build`. For records-keeping:

rohanpurohit:

ethanjli (approved):

Issue #322, opened 2021-04-21
Debounce firmware alarms


ethanjli:
Currently, if a measurement jumps around an alarm limit (even if it's within what the frontend will smooth/round/truncate away), it produces many alarm events. Obviously this is undesirable. In the firmware we need signal smoothing of values for alarms, as well as rounding for alarms, so that the frontend's displayed value is consistent with the alarm activation/inactivation behavior, in edge cases where roundoff makes a difference. This may be enough to fully debounce firmware alarms, or it may not; I suspect we might need a separate debouncing mechanism anyways, for events where we don't have a floating-point number to smooth.

This debouncing mechanism needs to support a "lockout" for FiO2 and flow rate (whose alarm limits should be automatically adjusted when FiO2 and flow rate settings are adjusted, as described in #323), where alarms don't trigger during an initial period in which the ventilator expects those measurements to still be transitioning to their new settings. However, the alarms do need to sound if the measurements can't reach their new settings within a reasonable amount of time.

PR #327, opened 2021-04-23
Implement driver for power management

### LINKED ISSUE #33: Power state indicator

### LINKED ISSUE #334: Add a power source indicator and Trigger an alarm if ventilator is running on battery
we would implement the feature to check if the unit is running on battery or not through the LTC4015 driver in the firmware and pass that data to the backend & frontend. Currently, we won't be reporting the remaining battery charge, which will be implemented later. We will also trigger an alarm if the unit is running on battery

rohanpurohit:
This PR includes:
- added a low driver for LTC4015 with following functions:
  - is_charger_enabled
  - read_battery_voltage
  - read_input_voltage
  - read_system_voltage
  - read_battery_current
  - read_input_current
- Added a high level driver that currently just checks if the charger is enabled and triggers an alarm
- moved AlarmsManager from BreathingCircuit to Application
- added a overloaded function for actiavte_alarm with no alarm_limits
- added a new LogEventCode - charger_disconnected
- made changes to add BatteryPower messagetype


ethanjli:
From an overview-level code review, this looks good. We discussed moving alarms-related functionality out of the high-level Sensor class for LTC4015, and changing the function interface and name to `read_charging_status` and renaming the `BatteryPower` message field from `charging_status` to `charging`. The next step is to make any necessary fixes once hardware gets tested, then I'll do another quick review and we can merge.

rohanpurohit:
Changes:
- Found in testing that while doing read from HAL_I2C_Mem_Read, failed when the device address was left-shifted by 1, thus the abstraction was changed to fix that.
- Renamed BatteryPower MessageType to Battery
- Renamed charging_status field of battery to charging
- moved alarm related code from the high-level sensor driver to power module 
- update read_charging_status method in low-level driver

ethanjli:
Results from testing on commit 875b8ed, with `ltc4015_sensor` removed from the list of firmware initializables:
1. the firmware behaves as expected: battery level starts from 0% and increases at a reasonable rate to 100% but then stays there, and a "Charger is disconnected" alarm is active once it's at 100%.
2. When I remove the `if (!parameters.ventilating)` branch from `Pufferfish::Driver::Power::transform`, everything still seems to behave correctly. I don't see any spamming of the event log; and I get this alarm on the quickstart screen, while ventilation is inactive. The only weird behavior I see here is that I get the "Charger is disconnected" event twice in the event log, occurring 23 seconds apart.

Results from testing on commit 352f3a3, with `ltc4015_sensor` removed from the list of firmware initializables:
- When I remove the `if (!parameters.ventilating)` branch from `Pufferfish::Driver::Power::transform` (and delete all code in `deactivate_alarms`, to make the commit compile), the event log is **not** spammed with "Charger is disconnected" events.

Results from testing on commit 2abfd33, with `ltc4015_sensor` removed from the list of firmware initializables:
1. When I run the firmware without any modifications, the firmware simulator for power management does not seem to work: the battery level is always shown as 100%, and charging is always `false`. I got a single alarm that the battery charger is disconnected, but the event log wasn't spammed by that alarm.
2. When I remove the `if (!parameters.ventilating)` branch from `Pufferfish::Driver::Power::transform`, the event log is spammed with "Charger is disconnected" events, but only when the ventilator is not ventilating; once the ventilator starts ventilating, it works fine
3. When I run the simulator backend, the power management simulation seems to work, and the alarms seem to work.

Results from testing on commit 01948ce, with `ltc4015_sensor` removed from the list of firmware initializables:
- This commit doesn't have a `if (!parameters.ventilating)` branch in `Pufferfish::Driver::Power::transform`. Battery level is always at 100%, and the event log is spammed with "Charger is disconnected" events.

Thus, some change made on 352f3a3 fixed the problem of spamming the event log even without the `if (!parameters.ventilating)` branch. My hypothesis is that `BreathingCircuit::AlarmsService::deactivate_alarms` was always deactivating the `LogEventCode::charger_disconnected` alarm whenever the breathing circuit found that the parameters had set `ventilating` to false. Thus, the changes made in 2abfd33 to remove `charger_disconnected` from the list of alarms which the Breathing Circuit alarms service could deactivate had a side effect of also solving the root cause of the event log spamming issue. By adding that event code back to the list of alarm codes for the Breathing Circuit alarm service, I was able to bring the spamming back. This also explains why your workaround to disable the `charger_disconnected` alarm was able to stop the spamming: then your Power alarms service was no longer fighting against the Breathing Circuit alarms service about whether to activate or deactivate `charger_disconnected`.

Since we have identified the root cause of the event log being spammed by `charger_disconnected` events, it is now correct to remove the `if (!parameters.ventilating)` branch from `Pufferfish::Driver::Power::transform`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+// NOLINTNEXTLINE(cert-msc51-cpp)
+std::ranlux24_base prng;
+// NOLINTNEXTLINE(readability-magic-numbers)
+std::uniform_int_distribution<int> uniform_int(1, 100);
+static constexpr float uniform_width = 1;
+std::uniform_real_distribution<float> uniform_centered(-uniform_width / 2, uniform_width / 2);
+
+void Simulator::input_clock(uint32_t current_time) {
+  if (initial_time_ == 0) {
```
ethanjli, line 15: The `initial_time` related code in the simulator backend was only so that the simulator backend's timestamps would resemble the firmware's timestamps, as Python's `time.time()` has time=0 at Jan 1 1970 while the firmware's `HAL::time()` has time=0 when the firmware first started. We should remove `initial_time_` from the firmware's power simulator.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
```diff
@@ -0,0 +1,75 @@
...
+    initial_time_ = current_time;
+  }
+  if (update_needed()) {
+    previous_time_ = current_time_;
+  }
+  current_time_ = current_time - initial_time_;
+}
+
+void Simulator::transform(uint32_t current_time, PowerManagement &power_management) {
+  input_clock(current_time);
```
ethanjli, line 25: When you call `input_clock` here and it determines that an update is needed, then it sets `previous_time_ = current_time_`. Next, you call `transform_charge`/`transform_discharge`, which does nothing if `previous_time_ == current_time_` (as determined by an `update_needed()` check). I haven't tried running the latest version of the firmware on this PR yet, but this code seems suspicious to me.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+void Simulator::transform_charge(PowerManagement &power_management) {
+  if (!update_needed()) {
+    return;
+  }
+  power_management.charging = true;
+  power_management.power_left +=
+      (1 + (power_responsiveness * uniform_centered(prng))) / uniform_int(prng);
```
ethanjli: This formula is seems more convoluted than necessary. If you're simply trying to add a random percentage to `power_left`, it would be clearer to just define a uniform random distribution ` std::uniform_real_distribution<float> uniform_percentage(min_increase_per_step, max_increase_per_step)` (and you can just delete `uniform_int` and `uniform_centered`) and add `uniform_percentage(prng)` to `power_left`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+void Simulator::transform_discharge(PowerManagement &power_management) {
+  if (!update_needed()) {
+    return;
+  }
+  power_management.charging = false;
+  power_management.power_left -=
+      (1 + (power_responsiveness * uniform_centered(prng))) / uniform_int(prng);
+  if (power_management.power_left == 0) {
```
ethanjli: If `transform_discharge` should be symmetric with `transform_charge`, should this `if` check be `power_management.power_left <= 0`? And should the body of the if branch also set `power_management.power_left = 0?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/HAL/Interfaces/I2CDevice.h
```diff
@@ -43,6 +43,15 @@ class I2CDevice {
    */
   virtual I2CDeviceStatus read(uint8_t *buf, size_t count) = 0;
 
+  /**
+   * Reads data from the device
```
ethanjli, line 47: It would be helpful to add a note here about when this `read` method should be used instead of the other version of the `read` declared on line 44.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/Simulator.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include "Pufferfish/Application/States.h"
+
+namespace Pufferfish::Driver::Power {
+
+class Simulator {
+ public:
+  void transform(uint32_t current_time, PowerManagement &power_management);
+  void transform_charge(PowerManagement &power_management);
+  void transform_discharge(PowerManagement &power_management);
```
ethanjli: These should be private methods.

proto/mcu_pb.proto
```diff
@@ -186,9 +187,9 @@ message ActiveLogEvents {
 
 // Power Management
 
-message BatteryPower {
+message PowerManagement {
```
ethanjli: Maybe it would be better to name this `MCUPowerStatus` (in the future we'll also be adding a `GUIPowerStatus` message for the Raspberry Pi, which is powered by a separate battery and a separate copy of the LTC4015 chip).

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/LTC4015/Device.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::array<uint8_t, sizeof(uint16_t)> buffer{};
+  I2CDeviceStatus ret = i2cdevice_.read(static_cast<uint16_t>(Command::vbat_addr), buffer);
+  if (ret != I2CDeviceStatus::ok) {
+    return ret;
+  }
+
+  uint16_t vbat_value = 0;
+  Util::read_bigend(buffer.data(), vbat_value);
+  // NOLINTNEXTLINE(readability-magic-numbers)
+  v_bat = vbat_value * 128.176;  // Vbat = [VBAT] • 128.176µV for lead-acid
```
ethanjli: All magic numbers for this driver should be defined as static constants, either private in the class declaration of `Device` (if the numbers aren't needed elsewhere) or in `Types.h` (if the numbers are needed elsewhere).

ethanjli: It would be great if you could add code comments to these formulas (and/or to the definitions of static constants for magic numbers) indicating where to find these formulas in the datasheet (e.g. a URL and page number)

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/LTC4015/Sensor.h
```diff
@@ -0,0 +1,60 @@
...
+class StateMachine {
+ public:
+  // More actions will be added later
+  enum class Action { initialize, measure };
+
+  [[nodiscard]] Action update(uint32_t current_time_us);
+
+ private:
+  Action next_action_ = Action::initialize;
+  uint32_t current_time_us_ = 0;
```
ethanjli: Do you still need `current_time_us_` in `StateMachine`? It doesn't look like it's being used anywhere.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/LTC4015/Sensor.cpp
```diff
@@ -0,0 +1,88 @@
...
+      return initialize(time_.micros());
+    case Action::measure:
+      return InitializableState::ok;
+  }
+  return InitializableState::failed;
+}
+
+InitializableState Sensor::output(PowerManagement &power_management) {
+  switch (next_action_) {
+    default:
```
ethanjli: Since you put the `default` case up here, I think you'll always fall through to `case Action::measure` and return from inside it, and you'll never reach the `return InitializableState::failed` line at the end of this method. Maybe you meant to move the `default` case to the end of the `switch` statement?

ethanjli (commented):
Here's my review of the code. The firmware implementation for #322 depends on modifying the Alarms Manager (which was refactored in this PR), so let's discuss how we should proceed (e.g. should we make a separate PR with just the Alarms Manager refactor and merge that in immediately, or should I copy the Alarms Manager refactor onto my branch for #322, or should we just merge my branch for #322 after we merge in #327, or some other strategy)

ethanjli:
Results from testing the power simulator in the firmware: simulated charging happens slowly (I think it's a reasonable charging rate), while discharging happens way too quickly (i.e. it goes from 100% to 0% in a few seconds). This bug only occurs in the firmware's power simulator, not in the simulator backend.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
ethanjli: Actually, the Breathing Circuit simulator also has `initial_time` (though maybe it shouldn't either). We can leave `initial_time` in the power simulator for this PR, just to be consistent.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
ethanjli: It looks like this same pattern is used in the BreathingCircuit's simulator, so that needs further investigation too.

rohanpurohit:
@ethanjli I have updated the power simulator and higher-level driver, this is ready for testing and code review, I will update the changes you suggested in #378 once we test and merge that.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
rohanpurohit: I'm now simply incrementing from 0 to 100 every second and decrementing on discharge, works fine and we can get rid of random distributions and randomness in results too, I'm resolving this!

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
rohanpurohit: I had done the same previously, although `cppcheck` complaints about `Core/Src/Pufferfish/Driver/Power/Simulator.cpp:53:35: style: Checking if unsigned expression 'power_management.power_left' is less than zero. [unsignedLessThanZero]
`

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/LTC4015/Device.cpp
rohanpurohit: Yep! added link for datasheet along with page number, they are only required now in `Device` so i have added it there.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/LTC4015/Sensor.h
rohanpurohit: Yep not using it, cleaned up the driver

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/LTC4015/Sensor.cpp
rohanpurohit: Yes fixed!

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/LTC4015/Device.cpp
```diff
@@ -0,0 +1,117 @@
...
+  if (ret != I2CDeviceStatus::ok) {
+    return ret;
+  }
+
+  uint16_t vbat_value = 0;
+  Util::read_bigend(buffer.data(), vbat_value);
+  v_bat =
+      vbat_value *
+      static_cast<uint16_t>(
+          batt_voltage_conversion);  // https://www.analog.com/media/en/technical-documentation/data-sheets/4015fb.pdf
```
ethanjli: The formatting here is weird, I think clang-format will fix it if you move the comment to its own line before the line where you assign `v_bat`. This also applies to the other comments in this file.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
ethanjli: Looks like I missed a basic thing which explains all of this: right after `previous_time` is set to `current_time_`, `current_time_` is changed to a new value.

ethanjli (commented):
Here's my review, just minor formatting changes. I tested the backend simulator and firmware simulator and they both work fine now. If we merge this into develop ahead of #378, I'll resolve any merge conflicts and make any changes in #378. For records-keeping:

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/Simulator.cpp
ethanjli: Ah, ok - I had thought `power_management.power_left` was a `float` rather than a `uint*_t`. Do you think we should change it to `float`, or leave it as `uint*_t`?

rohanpurohit: Let's change it to `float`?

rohanpurohit:
> Here's my review, just minor formatting changes. I tested the backend simulator and firmware simulator and they both work fine now. If we merge this into develop ahead of #378, I'll resolve any merge conflicts and make any changes in #378. 

either way is fine with me, I guess we need to use `Util::TImer` in the simulator and remove `Util::UnrecognizedMessage` as now it uses EnumMap.

ethanjli:
I just merged #378 into develop before I saw your latest comment. One other change which was buried in that PR's description is that we should change `AlarmsService::transform(PowerManagement &power_management,...` to `AlarmsService::transform(const PowerManagement &power_management,...` if that alarms service only needs to read values from `power_management` and doesn't need to write values to it.

ethanjli (approved):

PR #329, opened 2021-04-23
FiO2 & flow alarms

### LINKED ISSUE #323: Automatically set & update FiO2 and flow rate alarm limits
Right now there's no way to change the FiO2 and flow rate alarm limits in the firmware, from a frontend action. The alarm limits are locked at the maximum possible range for those values, so alarms should never trigger. What we actually need is to automatically change the alarm limits to be some tolerance range around the FiO2 and flow rate parameter settings; whenever the parameter settings change, the alarm limits should be updated. This could be done either from ParametersService (not elegant) or from AlarmLimitsService (better, but could there be some kind of synchronization issue? Maybe not, if we can guarantee the AlarmLimitsService always runs immediately after ParametersService and before parameter updates are actuated by the valves; I think we can make this guarantee).

ethanjli:
This PR resolves #323 by having the firmware automatically set FiO2 and flow alarm limits as a tolerance window (+/- 2) around the FiO2 and flow parameters, and filling the remaining gaps for support of FiO2 and flow rate alarms.

Additional minor changes:
- Range protobuf messages now use int32 instead of uint32, to allow for negative numbers in alarm limits and ranges
- The frontend now shows more useful initial values for the FiO2 and flow rate settings on the quick start screen, reducing the number of increment/decrement button-presses needed to adjust to clinically relevant values (in the absence of the rotary encoder).

### DIFF
--- a/backend/ventserver/protocols/protobuf/mcu_pb.py
+++ b/backend/ventserver/protocols/protobuf/mcu_pb.py
@@ -21,14 +21,14 @@ class LogEventCode(betterproto.Enum):
     """Log Events"""
 
     # Patient
-    fio2_too_low = 0
-    fio2_too_high = 1
-    spo2_too_low = 2
-    spo2_too_high = 3
-    rr_too_low = 4
-    rr_too_high = 5
-    hr_too_low = 6
-    hr_too_high = 7
+    spo2_too_low = 0
+    spo2_too_high = 1
+    hr_too_low = 2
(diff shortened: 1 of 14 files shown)


ethanjli:
Because this PR has been sufficiently (though lightly) tested on basic functionality and is part of the v0.7 milestone, I'm going to merge this into develop as-is, and as prototype-grade code. For records-keeping:

Issue #324, opened 2021-04-21
Set Alarm Limits Screen's Apply Changes button is missing a confirmation dialogue


ethanjli:
When the Set Alarm Limits screen's "Apply Changes" button is clicked, it should bring up a confirmation dialogue displaying the changed alarm limits. This button should only be enabled if unsaved changes to alarm limits have been made. Maybe there should also be a "Discard Changes" button (enabled only if unsaved changes have been made, and which also brings up a confirmation dialogue displaying the unsaved changes to discard), just like the "Cancel" button of our modals for changing settings and alarm limits?

(this issue was created from https://github.com/pez-globo/pufferfish-software/pull/313#issuecomment-818406959 )

ethanjli:
From basic testing, there's some confusing and inconsistent behavior from the quickstart screen: I can adjust FiO2 and flow rate settings, and they're applied when I press "Start Ventilation". I can also adjust SpO2 and HR alarm limits, but they're discarded when I press "Start Ventilation". The catch here is that in order for those changes on the alarm limits page to be used when I press "Start Ventilation", first I have to press "Apply Changes". This behavior is really counter-intuitive. To make it consistent, I've made the "Start Ventilation" button automatically apply/commit the AlarmLimitsStandby values when I press it (4ee0ee4). And we need to hide the "Apply Changes" button from the alarm limits screen in standby mode (I don't know how to do this).

ethanjli:
Additionally, when the Set Alarms screen is opened from another screen (e.g. the dashboard), its values should be initialized to the current values in either AlarmLimits or AlarmLimitsRequest - so it shouldn't keep any unsaved changes which were to be discarded previously.

Issue #307, opened 2021-04-08
Settings overview in breadcrumbs modal shows the wrong numbers


ethanjli:
The "HFNC Control" screen of the Ventilation Controls breadcrumbs modal shows the wrong numbers in its black panels. The Heart Rate panel's measurement appears to be stuck at 0 (instead of showing the sensor data which is shown by the Heart Rate panel on the dashboard), while the SpO2 shows rapidly-flickering unfiltered measurements (instead of showing the filtered data which is shown by the SpO2 panel on the dashboard). 

Unsaved changes to alarm limits are not reflected in the small numbers on the upper right corners of the alarm limits panels.
I think the FiO2 and Flow Rate panels should show the settings as the big numbers; right now it looks like they're configured to show the actual sensor measurements.

**Software:**
 - Domain: frontend
 - Version: I see this on https://github.com/pez-globo/pufferfish-software/tree/feature/firmware-log-events (e.g. on commit d0d5436), but I suspect it should be reproducible on develop.

**Additional context**
I am testing with ventserver.application, but I expect this would be reproducible with ventserver.simulation too.

ethanjli:
We might be able to show unsaved changes to ParametersRequest and AlarmLimitsRequest if we commit all increment/decrement operations to ParametersRequestStandby and AlarmLimitsRequestStandby. In this case, we should probably rename "Standby" to "Draft" (e.g. "ParametersRequestDraft").

rohanpurohit:
after #338 unsaved changes to alarm limits are visible in the HFNC control alarm limit panels, but only in the `dashboard` page, since we decided we would commit changes from increment/decrement buttons to AlarmLimitsRequestStandby which is now AlarmLimitsRequestDraft. In a scenario where we are on the alarm limits screen and open the events log and then the alarm limits modal to make changes, on making changes and pressing `submit` everything is as expected, But if we make changes and press `cancel` and `confirm`, we find ourselves also losing the `unsaved` changes on the alarm limits screen if any, as the changes made in alarm limits screen and breadcrumbs modal are independent. thus it would be easier if we propagate the unsaved changes in the alarm limits screen to the breadcrumbs modal and vice-versa

PR #330, opened 2021-04-24
Clean up the HFNC dashboard

### LINKED ISSUE #309: Dashboard should show both measured FiO2 & flow and set FiO2 & flow
Currently the frontend's HFNC dashboard shows a large area with three plots. We need to remove these for the HFNC-only release. I propose that we replace the entire right half of the dashboard with another panel collection (so the dashboard will consist of two panel collections: one on the left, and one on the right).

We could add two more black panels to the collection on the left (perhaps as an additional column), which will show the measured FiO2 and flow rates (these come from SensorMeasurements), as well as alarm limits. The panel collection on the right could add two black panels to show the set FiO2 and flow rates (these come from Parameters), but not alarm limits. Note that the alarm limits for FiO2 and flow rate are not adjustable, so all four of these panels will open "Set Value" modals when tapped.

We should remove the ROX Index panel, since we have no way to determine respiratory rate for the HFNC release (respiratory rate is needed to calculate ROX index).

If we have too much whitespace after these changes, it'd be great if we could increase the font size of all text in the dashboard's panels.

ethanjli:
This PR closes #309 by changing the layout of the HFNC-only dashboard (removing plots, and showing both set and measured values for FiO2 and flow rate) and increasing the size of text on the dashboard. Additionally, in preparation for the v0.7 release, it:

- Disables the screensaver
- Reenables the FDO2 sensor in the firmware by default
- Fixes unintuitive behavior between the set alarms screen and the "Start Ventilation" button in standby mode (see https://github.com/pez-globo/pufferfish-software/issues/324#issuecomment-825985574 for more details)
- Fixes parts of #307 
- Adds an emergency use notice to the landing screen

### DIFF
--- a/backend/ventserver/simulation/alarm_limits.py
+++ b/backend/ventserver/simulation/alarm_limits.py
@@ -97,7 +97,7 @@ class HFNC(Service):
     """Alarm limits servicing for HFNC mode."""
 
     FLOW_TOLERANCE = 2  # L/min
-    FLOW_MIN = 0  # L/min
+    FLOW_MIN = -2  # L/min
     FLOW_MAX = 80  # L/min
 
     def transform(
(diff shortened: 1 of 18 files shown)


ethanjli:
Because this PR has been sufficiently (though lightly) tested on basic functionality and is completes the v0.7 milestone, I'm going to merge this into develop as-is, and as prototype-grade code. This version of software should be tested more thoroughly by other people before any deployment. Any serious bugs which need to be fixed on v0.7 will be addressed in hotfix branches and merged onto develop in future PRs. For records-keeping:

PR #333, opened 2021-04-26
Restructure Honeywell ABP sensor driver

### LINKED ISSUE #231: Bring up absolute pressure sensor for FiO2 calculations
In order to have correct FiO2 calculations which don't make assumptions about the ambient/current pressure, we will need to measure the absolute ambient pressure instead of assuming it is 1 atm. This will require updating and bringing up the driver for ~the Honeywell ABP~ a TBD sensor with our protoboard, in our new architectural pattern as done for the SFM3019.

rohanpurohit:
This PR:
- update structure similar to SFM drivers
- fix includes
- Add high level driver


firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Device.cpp
```diff
@@ -10,39 +10,35 @@
...
+
+I2CDeviceStatus Device::read_sample(ABPSample &sample) {
   std::array<uint8_t, 2> data{{0, 0}};
   I2CDeviceStatus ret = dev_.read(data.data(), data.size());
   if (ret != I2CDeviceStatus::ok) {
     return ret;
   }
 
-  static const uint8_t status_shift = 6;
-  static const size_t bridge_high = 0;
```
rohanpurohit: @ethanjli is this a typo? shouldn't it be 1? and I think by using `read_bigend` now we won't be facing the problem there.

ethanjli: It could be a typo, this was implemented by someone else and I'm not familiar with the I2C interface of the ABP sensor so I don't know anything about the bridge data. If it's trying to read multiple bytes of data into a single number, read_bigend is probably an appropriate refactor here.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Types.h
```diff
@@ -1,17 +0,0 @@
...
-#pragma once
-
-#include <cstdint>
-
-namespace Pufferfish {
-
-/**
- * A unit of pressure read from sensors
- */
-enum class PressureUnit { unknown = 0, psi, pa, k_pa, m_pa, m_bar, bar };
```
ethanjli, line 15: If/when we add drivers for other pressure sensors, we will probably want to have an application-wide definition of the pressure unit enum, probably in `Core/Inc/Pufferfish/Types.h` or `Core/Inc/Pufferfish/Units.h`. It's fine to move `PressureUnit` into `Driver/I2CHoneywellABP/Types.h` for now, as we can always move the declaration later if/when we actually need it and once it becomes clear what the filename should be.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -11,53 +11,16 @@
...
-static const ABPConfig abpxxxx001pg2a3 = {0x28, 0.0, 1.0, PressureUnit::psi};
-static const ABPConfig abpxxxx005pg2a3 = {0x28, 0.0, 5.0, PressureUnit::psi};
-static const ABPConfig abpxxxx030pg2a3 = {0x28, 0.0, 30.0, PressureUnit::psi};
+namespace Pufferfish::Driver::I2C::HoneywellABP {
 
 /**
  * Driver for Honeywell ABP pressure sensor
  */
-class HoneywellABP : public Testable {
+class Device : public Testable {
```
ethanjli: Let's remove `Testable` and the `test` method from this class (see also another comment below).

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Device.cpp
```diff
@@ -10,39 +10,35 @@
...
 }
 
-float HoneywellABP::raw_to_pressure(uint16_t output) const {
-  if (output < output_min) {
-    output = output_min;
-  } else if (output > output_max) {
-    output = output_max;
-  }
+float Device::raw_to_pressure(uint16_t output) const {
+  Util::clamp<uint16_t>(output, output_min, output_max);
```
ethanjli: Does `Util::clamp` produce the clamped result as its returned value, or does it modify the first parameter as an output parameter?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Sensor.h
```diff
@@ -0,0 +1,54 @@
...
+  InitializableState setup() override;
+  InitializableState output(float &output);
+
+ private:
+  static const size_t max_retries_setup = 8;    // max retries for setup
+  static const size_t max_retries_measure = 8;  // max retries for measuring
+
+  Device device_;
+  ABPSample sample_{};
+  using Action = StateMachine::Action;
```
ethanjli: This `using` declaration should be in its own blank-line-delimited section at the top of the `private` section. i.e.:

```
private:
  using declaration;
  using declaration;
  ...

  static const declaration;
  static const declaration;
  ...

  regular const declaration;
  regular non-const declaration;
  ...
```

If there are other files in this PR which need to be adjusted to match this, let's reorder declarations in those files too. We should follow this convention everywhere in the firmware.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Sensor.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+InitializableState Sensor::initialize() {
+  if (retry_count_ > max_retries_setup) {
+    return InitializableState::failed;
+  }
+
+  retry_count_ = 0;
+  // Test
+  while (device_.test() != I2CDeviceStatus::ok) {
```
ethanjli: Let's refactor the `test` method out of the `Device` class; test functionality to be carried out during initialization should instead be implemented in the `Sensor::initialize()` method, or else in private methods called by `Sensor::initialize()`.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -338,7 +346,7 @@ PF::Driver::Power::Simulator power_simulator;
 
 // Initializables
 auto initializables =
-    PF::Driver::make_initializables(sfm3019_air, sfm3019_o2, fdo2, nonin_oem, ltc4015);
+    PF::Driver::make_initializables(sfm3019_air, sfm3019_o2, fdo2, nonin_oem, ltc4015, abp);
```
ethanjli: Let's move `abp` to be either right before `fdo2` or right after `fdo2`, so that all sensors for the breathing circuit are next to each other in this list.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -612,6 +620,7 @@ int main(void)
   // Normal loop
   while (true) {
     uint32_t current_time = hal_time.millis();
+    float abp_pressure = 0;
```
ethanjli: `hfnc.sensor_vars()` will need to be modified to provide a place to store the measurement(s) produced by `abp.output(...)`. For now, maybe we can use `hfnc.sensor_vars().pout` (we'll probably need to rename this to a more standard name later).

ethanjli (commented):
Here's my review from looking at the changes.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Device.cpp
rohanpurohit: Oops my bad! it returns the clamed result.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Sensor.h
rohanpurohit: oh okay Noted! I have updated sensor.h and device.h accordingly.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Sensor.h
```diff
@@ -37,15 +37,19 @@ class Sensor : public Initializable {
   InitializableState output(float &output);
 
  private:
+  using Action = StateMachine::Action;
+
   static const size_t max_retries_setup = 8;    // max retries for setup
   static const size_t max_retries_measure = 8;  // max retries for measuring
 
+  const float pmin = 0.0;  // minimum pressure for abpxxxx001pg2a3
+  const float pmax = 1.0;  // maximum pressure for abpxxxx001pg2a3
```
ethanjli: We should probably rename these variables `p_min` and `p_max` instead of `pmin` and `pmax`.

Anywhere we have numeric literals for physical quantities, we should also annotate them with units if the units are known, e.g.:

```
  const float p_min = 0.0;  // psi; minimum pressure for abpxxxx001pg2a3
  const float p_max = 1.0;  // psi; maximum pressure for abpxxxx001pg2a3
```

(assuming the abpxxxx001pg2a3 uses PSI)

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -32,23 +31,20 @@ class Device : public Testable {
...
 
-  I2CDeviceStatus test() override;
-  I2CDeviceStatus reset() override;
-
  private:
   Pufferfish::HAL::Interfaces::I2CDevice &dev_;
 
+  static const uint8_t status_shift = 6;
+  static const size_t bridge_high = 0;
+  static const size_t bridge_low = 0;
```
ethanjli: How many of these do we actually need? Can we remove `bridge_high` and `bridge_low` now that we use `read_bigend`?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Sensor.cpp
```diff
@@ -76,6 +70,12 @@ InitializableState Sensor::measure(float &output) {
     retry_count_ = 0;  // reset retries to 0 for next measurement
     output = sample_.pressure;
     next_action_ = fsm_.update();
+    if (sample_.status != ABPStatus::no_error && sample_.status != ABPStatus::stale_data) {
+      return InitializableState::failed;
+    }
+    if (Util::within(sample_.pressure, pmin, pmax)) {
+      return InitializableState::failed;
+    }
```
ethanjli: We should definitely check the pressure against `pmin` and `pmax` in `initialize` (currently I don't see that check there)

Currently the `SFM3019` `Sensor` class does not check its measured values against `flow_min` and `flow_max` in the `measure` method, only in the `check_range` method (which is part of the setup routine). I'm not sure whether we want `measure` to include range-checking, but we should be consistent across sensors. If we want to include range-checking in `measure`, then we will need to figure out how to handle such errors. We'll probably need to do some risk analysis (and/or look at what was identified and recorded in Matrix Requirements) to determine whether we do want range-checking in `measure`. For now, let's have the HoneywellABP `Sensor` only do range-checking in `initialize`, and I'll file a Github issue about this (#422).

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -32,23 +31,20 @@ class Device : public Testable {
...
+
   // pressure range (refer to datasheet)
   const float pmin;  // minimum pressure
   const float pmax;  // maximum pressure
```
ethanjli: Do we need still these?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -32,23 +31,20 @@ class Device : public Testable {
...
-
   // sensor 14-bit output range
   const uint16_t output_min = 0x0666;  // 10% of 2^14
   const uint16_t output_max = 0x399A;  // 90% of 2^14
```
ethanjli: Should these be `static const`?

ethanjli (commented):
Here are some more comments.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
rohanpurohit: Yeah we can remove `bridge_high`  and `bridge_low`

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
rohanpurohit: Yes they are initialized in the constructor with the config from main.cpp, and used in the raw_to_pressure formula in `Device.cpp`

rohanpurohit: Oh wait do you mean since we are only using one config? right

rohanpurohit:
@ethanjli we tested this PR on the sensor and found these outputs:
**measurements:**
1. `idle (without any external pressure)` - 0
2. `when applying pressure on the inlet of the sensor with my finger` - `0.157613`
3. `when blowing air into the sensor` - `0.00740`, `0.02`,`0.00167836`

also regarding `stale_data` there is nothing more we need to do.
according to the [datasheet](https://sensing.honeywell.com/index.php/ci_id/45841/la_id/1/document/1/re_id/0) `When the status bits read “1 0”, “stale” data is indicated,`
if you see [Device.cpp L31](https://github.com/pez-globo/pufferfish-software/pull/333/files#diff-23451c400d356534c491f7cd01ea4c370083c758ca89edcfe17e55a1139bb7deR31) we need to check the 2 bits `S0 and S1` and if we get 1 & 0 (000010 = 2 ,enum value of ABPStatus)  we store the status as `stale_data`.

In the datasheet, we are advised to ignore `command_mode` and `diagnostic` status, which we are doing here [Sensor.cpp L65](https://github.com/pez-globo/pufferfish-software/pull/333/files#diff-9c610f0934573bc0c976233d888d132a978f848d44300cd307d17c87fd080519R65)

rohanpurohit:

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -0,0 +1,48 @@
...
+      : dev_(dev), pmin(cfg.pmin), pmax(cfg.pmax), unit(cfg.unit) {}
+
+  [[nodiscard]] float raw_to_pressure(uint16_t output) const;
+
+  /**
+   * Reads out the pressure from the sensor
+   * @param sample[out] the sensor reading; only valid on success
+   * @return ok on success, error code otherwise
+   */
+  I2CDeviceStatus read_sample(ABPSample &sample);
```
ethanjli: Perhaps we can rename `ABPSample` just to `Sample`?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/I2C/HoneywellABP/Device.h
```diff
@@ -0,0 +1,48 @@
...
+
+ private:
+  Pufferfish::HAL::Interfaces::I2CDevice &dev_;
+
+  static const uint8_t status_shift = 6;
+  // sensor 14-bit output range
+  static const uint16_t output_min = 0x0666;  // 10% of 2^14
+  static const uint16_t output_max = 0x399A;  // 90% of 2^14
```
ethanjli, line 38: `static const` members should be above non-`static` members

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -641,6 +649,8 @@ int main(void)
     // Independent Sensors
     fdo2.output(hfnc.sensor_vars().po2);
     nonin_oem.output(store.sensor_measurements_raw().spo2, store.sensor_measurements_raw().hr);
+    // *temporary* should be used in the breathing circuit
+    abp.output(hfnc.sensor_vars().p_out);
```
ethanjli: We'll need to rename `p_out` to `p_out_above_atm`, as p_out = p_out_above_atm (from ABP) + p_atm (from BMP388)

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Sensor.cpp
```diff
@@ -0,0 +1,108 @@
...
+  retry_count_ = 0;  // reset retries to 0 for measuring
+  return InitializableState::setup;
+}
+
+InitializableState Sensor::measure(uint32_t current_time, float &output) {
+  if (device_.read_sample(sample_) == I2CDeviceStatus::ok) {
+    retry_count_ = 0;  // reset retries to 0 for next measurement
+    output = sample_.pressure;
+    next_action_ = fsm_.update(current_time);
+    if (sample_.status != ABPStatus::no_error && sample_.status != ABPStatus::stale_data) {
```
ethanjli: We should just do `if (sample_.status == ABPStatus::no_error) return ok`, else fall through to the retry logic.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/I2C/HoneywellABP/Sensor.cpp
```diff
@@ -0,0 +1,108 @@
...
+    ++retry_count_;
+    if (retry_count_ > max_retries_setup) {
+      return InitializableState::failed;
+    }
+    if (sample_.status != ABPStatus::no_error && sample_.status != ABPStatus::stale_data) {
+      return InitializableState::failed;
+    }
+    if (!Util::within(sample_.pressure, p_min, p_max)) {
+      return InitializableState::failed;
+    }
```
ethanjli: We'll instead have a separate `check_range` state in the state machine, like what is done for the SFM3019 `Sensor` class.

ethanjli (commented):
Here are the changes discussed in our meeting today:

rohanpurohit:
working fine in testing with the sensor.

ethanjli (approved):

PR #337, opened 2021-05-03
Play audio in frontend for all active alarms

rohanpurohit:
Currently:
- plays alarms for the patient and system log events

**Bugs:**
- start ventilation -> pause ventilation -> start ventilation again-> in dashboard alarm gets triggered, to mute we need to toggle `mute icon` twice to get it working.

Can be reproduced on develop as well.
seems to be a state issue or useEffect

### DIFF
--- a/frontend/src/modules/app/OverlayScreen.tsx
+++ b/frontend/src/modules/app/OverlayScreen.tsx
@@ -77,8 +77,10 @@ const AudioAlarm = (): JSX.Element => {
   const alarmMuteStatus = useSelector(getAlarmMuteStatus, shallowEqual);
   const [audio] = useState(new Audio(`${process.env.PUBLIC_URL}/alarm.mp3`));
   audio.loop = true;
+  const storeData = store.getState();
+  const events = storeData.controller.eventLog.nextLogEvents.elements;
   const [playing, setPlaying] = useState(alarmMuteStatus.active);
-
+  const patientLogEvent = events.find((el: LogEvent) => (el.type as number) < 2);
   useEffect(() => {
     if (playing) {
       audio.play();
@@ -93,22 +95,21 @@ const AudioAlarm = (): JSX.Element => {
   useEffect(() => {
     if (popupEventLog) {
       setPlaying(false);
-      if (popupEventLog.code === BACKEND_CONNECTION_LOST_CODE) {
+      if (popupEventLog && patientLogEvent) {
         setPlaying(true);
+        dispatch({ type: RED_BORDER, status: true });
       }
-      dispatch({ type: RED_BORDER, status: true });
     } else {
       setPlaying(false);
       dispatch({ type: RED_BORDER, status: false });
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [popupEventLog]);
+  }, [popupEventLog, patientLogEvent, dispatch]);
 
   useEffect(() => {
     if (popupEventLog) {
       dispatch({ type: RED_BORDER, status: !alarmMuteStatus.active });
     }
-    if (popupEventLog && popupEventLog.code === BACKEND_CONNECTION_LOST_CODE) {
+    if (popupEventLog && patientLogEvent) {
       setPlaying(!alarmMuteStatus.active);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


frontend/src/modules/app/OverlayScreen.tsx
```diff
@@ -93,22 +95,21 @@ const AudioAlarm = (): JSX.Element => {
   useEffect(() => {
     if (popupEventLog) {
       setPlaying(false);
-      if (popupEventLog.code === BACKEND_CONNECTION_LOST_CODE) {
+      if (popupEventLog && patientLogEvent) {
```
ethanjli: A BACKEND_CONNECTION_LOST event is a system log event rather than a patient log event. The variable `patientLogEvent` needs to be renamed and/or changed to reflect that both system alarms and patient alarms will trigger audio in the frontend.

frontend/src/modules/app/OverlayScreen.tsx
```diff
@@ -77,8 +77,10 @@ const AudioAlarm = (): JSX.Element => {
   const alarmMuteStatus = useSelector(getAlarmMuteStatus, shallowEqual);
   const [audio] = useState(new Audio(`${process.env.PUBLIC_URL}/alarm.mp3`));
   audio.loop = true;
+  const storeData = store.getState();
+  const events = storeData.controller.eventLog.nextLogEvents.elements;
```
ethanjli: Is there any specific reason why we're not just using the `getNextLogEvents` selector?

frontend/src/modules/app/OverlayScreen.tsx
```diff
@@ -77,8 +77,10 @@ const AudioAlarm = (): JSX.Element => {
...
   const [playing, setPlaying] = useState(alarmMuteStatus.active);
-
+  const patientLogEvent = events.find((el: LogEvent) => (el.type as number) < 2);
```
ethanjli: This function looks like it was inspired by the `getPopupEventLog` selector. Maybe we should make this be its own selector?

ethanjli (commented):
I did light testing and it seems to work. I haven't been able to reproduce your bug, maybe it requires a more specific sequence of events. Perhaps it needs to be filed as a separate bug report.

ethanjli (commented):
This PR works on the latest changes, and the new approach to implementation looks good (clear and simple). Let's just rename the selector to something more specific, i.e. `getHasActiveAlarms` (this name indicates that the selector just decides whether there are active alarms) rather than `getAlarms` (which I would expect to return a list of all alarms). After this minor rename, let's merge. For records-keeping:

rohanpurohit:
Few observations that still exist and are mostly related to how the alarmMuteStatus store and then the local state is behaving:

1.  start ventilation -> any alarm is triggered -> pause ventilation -> start ventilation again-> if an alarm is still active, it is muted on start, maybe that behavior is not preferred?
2.  alarm sound is playing, and a page is changed (for instance to set_alarms page) and then we try to mute the alarm, need to toggle it twice to get it working as expected.
3. item 2 can also be observed sometimes when ventilation is paused and restarted

other than that, muting the alarm seems to be fine when we don't navigate from the dashboard.
probably is a very small fix to solve the above-mentioned issues but needs debugging.


1. This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? yes.
2. Were any of these contributions also part of work you did for an employer or a client? no.
3. Does this work include, or is it based on, any third-party work which you did not create? no.

ethanjli:
I'm unable to reproduce item 1 when I test with the simulator. Patient alarms should indeed be cancelled in quickstart (this is taken care of by AlarmsService, and new alarms get triggered when ventilation starts again if the condition is still there to trigger the alarm), though we'll want to have the low battery alarm still active in quickstart. I'm also unable to reproduce item 2. Can you file a new issue for each of these things, and we can follow up on them separately from this PR? I suspect they are independent problems from the audio playing: it sounds like item 2 related to the implementation of the alarm muting (which will definitely be changed), while item 1 is something else.

ethanjli (approved):

rohanpurohit:
> I'm unable to reproduce item 1 when I test with the simulator. Patient alarms should indeed be cancelled in quickstart (this is taken care of by AlarmsService, and new alarms get triggered when ventilation starts again if the condition is still there to trigger the alarm), though we'll want to have the low battery alarm still active in quickstart. I'm also unable to reproduce item 2. Can you file a new issue for each of these things, and we can follow up on them separately from this PR? I suspect they are independent problems from the audio playing: it sounds like item 2 related to the implementation of the alarm muting (which will definitely be changed), while item 1 is something else.

maybe I dint describe the first item properly
- If I restart ventilation while an alarm is playing, on starting again should the alarm be muted or playing? 

will create issues and we can take it from there.

Issue #308, opened 2021-04-08
Reorganize firmware's HAL file namespace scheme and file & class names


ethanjli:
Our firmware follows a convention that the folder path for each file should match its namespace structure. For example, the classes in Pufferfish/Driver/BreathingCircuit/Alarms.h are all in the `Pufferfish::Driver::BreathingCircuit` namespace. However, the files in Pufferfish/HAL predate this convention. They should all be updated to follow this convention, and class names should be updated to remove redundant words. For example, with the files related to AnalogInput:
- Pufferfish/HAL/Interfaces/AnalogInput.h should declare its classes in the `Pufferfish::HAL::Interfaces` namespace, rather than the `Pufferfish::HAL` namespace
- Pufferfish/HAL/Mock/MockAnalogInput.h should be renamed to Pufferfish/HAL/Mock/AnalogInput.h, and it should declare its classes in the `Pufferfish::HAL::Mock` namespace, rather than the `Pufferfish::HAL` namespace. The `Pufferfish::HAL::Mock::MockAnalogInput` class (which will inherit from `Pufferfish::HAL::Interfaces::AnalogInput`) should be renamed to `Pufferfish::HAL::Mock::AnalogInput`.
- Pufferfish/HAL/STM32/HALAnalogInput.h should be renamed to Pufferfish/HAL/STM32/AnalogInput.h, and it should declare its classes in the `Pufferfish::HAL::STM32` namespace, rather than the `Pufferfish::HAL` namespace. The `Pufferfish::HAL::STM32::HALAnalogInput` class (which will inherit from `Pufferfish::HAL::Interfaces::AnalogInput`) should be renamed to `Pufferfish::HAL::STM32::AnalogInput`.

Finally, we should delete the `Pufferfish/HAL/HAL.h` file, as half of it is redundant with `Pufferfish/HAL/STM32/HAL.h`, and the remaining includes (which are `#includes` of interface files) should all be specified individually in any dependent files on an as-needed basis.

@rohanpurohit Since this is purely a renaming task and it'll give exposure to all of the HAL classes, this could be a good first task for Raavi to do together with you (I'm not sure Raavi has accepted the invitation I sent her to our Github team, she needs to do it from her email before we can assign her issues; once she does, can you add her as an assignee?), as a "hello world" of making some changes, building and testing in STM32Cube IDE, and using our Github Actions infrastructure.

PR #338, opened 2021-05-03
frontend: Improve set alarms screen 

### LINKED ISSUE #324: Set Alarm Limits Screen's Apply Changes button is missing a confirmation dialogue

### LINKED ISSUE #345: Navigating to a different screen should show confirmation dialogue
Currently, when ventilation is running if we change alarm limits in set alarms page and navigate to dashboard it should show a confirmation dialogue displaying what has changed and ask for confirmation.

rohanpurohit:
This PR:
- adds a Cancel button that resets the changes, and renames `apply changes` to `Submit`
- disables `Submit` and `Cancel` button if there is no change in limits and vice versa.
- hide both these buttons when in standby mode.
- rename AlarmLimitsRequestResponse:standby ParametersRequestResponse:standby to draft and rename accordingly
- fix a few console errors related to Grid tag in modalpopup:
 Failed prop type: The prop `alignItems` of `Grid` must be used on `container`. ,
 Following component has two or more children with the same key attribute: "indicator-[object Object]". This may cause glitches and misbehavior in rendering process. Component:
- modal popup content shows the changed limits for both Submit and Cancel
- handle discarding of changes in modal popup
- clicking `dashboard ` button set alarms page shows a confirmation dialog
- show unsaved changes in alarm limits on HFNC control in multistepwizard


frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -287,30 +292,56 @@ export const AlarmsPage = (): JSX.Element => {
...
+  };
+
+  const handleDiscardConfirm = () => {
+    setDiscardOpen(false);
+    handleDiscard();
+    setIsDisabled(true);
   };
+
+  const handleDiscard = () => {
+    const storeData = store.getState();
```
ethanjli: Are you directly accessing the store because it won't work to use the `getAlarmLimits` selector?

ethanjli (commented):
Observations from basic testing, after these are addressed I'll do another round of testing to see if any other emergent behaviors have issues:

- When I am in standby mode, I still see the "Discard Changes" button. If I adjust an alarm limit, it becomes enabled; when I click it, it brings up a confirmation dialogue and if I confirm then it resets the changes. I didn't envision having this functionality, but it could be useful. However, the confirmation dialogue says something like "**Discard Changes?** Change HR to 60 - 105?", where instead it should say something like "**Keep Previous Values?** Keep HR alarm range at 60 - 105?" - it just needs to be consistent with the confirmation text from the Alarm Limits modal dialog.
- When ventilation is running, I see both "Discard Changes" as a green button and "Apply Changes" as a green button. For consistency, maybe this should be "Cancel" as a blue button and "Submit" as a green button. However, "Discard Changes" is more informative as a button label than "Cancel" in standby mode - so if the simplest way to make the button say "Cancel" in ventilating mode would to just hide the "Discard Changes" button in standby mode, it's fine to hide that button in standby mode.
- When ventilation is running and I adjust a value from the Alarm Limits screen, I can navigate away to another screen to bypass the confirmation messages. The changes remain in the screen, but they are not committed to AlarmLimitsRequest. Actually, navigating to a different screen (Dashboard or Settings) should also trigger the confirmation dialog if there are unsaved changes. If fixing this is out-of-scope of this PR, can you create a new issue to track it separately?

ethanjli:
@Sudhir-dev Item number 3 is necessary but seems like it might be complicated because it probably interacts with routing. We'll need to have documentation before proceeding with this

ethanjli:
Observations from testing:

1. When I first start the frontend and am in standby mode, the alarm limits are initialized as SpO2 = [90 - 100], HR = [60 - 100]. If I change the alarm limits and press the "Discard Changes" button, the confirmation dialog which pops up asks if I want to keep SpO2 at [90 - 100] and HR at [60 - 100]. If I press "Confirm", it does what it says - good.
2. Now I think the presence of "Discard Changes" and absence of the "Apply Changes" button in standby mode is confusing. When I see a button in the lower-right corner, I'm primed to think it means "Apply Changes" - this is the pattern set everywhere else in the frontend. I found myself instinctively trying to click it to save changes, without reading the button, Let's just hide the "Cancel"/"Discard Changes" button in standby mode, to make button placement in the frontend more consistent with assumptions we want to encourage the user to make. This also would make the Alarm Limits screen more consistent with the Quickstart screen.
3. When I start the frontend, start ventilation without changing any alarm limits, change the alarm limits from the alarm limits page, save the changes (e.g. to SpO2 = [90 - 99], HR = [60 - 99]), navigate to the dashboard, the dashboard reflects the new AlarmLimitsRequest values. But if I then navigate back to the alarm limits page, it shows the alarm limits as the values from before saving changes (namely it shows them as SpO2 = [90 - 100] and HR = [60 - 100]). Initially the "Cancel" and "Submit" buttons are disabled, even though the values shown on the screen are different from the actual values of AlarmLimitsRequest. Once I start adjusting the alarm limits, then the enabled/disabled state behaviors of the "Cancel" and "Submit" buttons become correct with respect to the actual values of the AlarmLimitsRequest (so e.g. if I use the alarm limits buttons to change the SpO2 and HR upper limits to 99 again, then the Cancel & Submit buttons are disabled; but once I change them back to 100, then the Cancel & Submit buttons become enabled)

Perhaps we should pause work on this PR until after we merge #340, because that PR uses AlarmLimitsRequestStandby as the only place to store unsaved changes for AlarmLimitsRequest - so it will interact with the bug I describe in item 3, and it might even fix that behavior if the "Cancel" button confirmation dialog triggers a dispatch which changes the values of AlarmLimitsRequestStandby to be overwritten with the values of AlarmLimitsRequest.

rohanpurohit:
> Observations from testing:
> 
> 1. When I first start the frontend and am in standby mode, the alarm limits are initialized as SpO2 = [90 - 100], HR = [60 - 100]. If I change the alarm limits and press the "Discard Changes" button, the confirmation dialog which pops up asks if I want to keep SpO2 at [90 - 100] and HR at [60 - 100]. If I press "Confirm", it does what it says - good.
> 2. Now I think the presence of "Discard Changes" and absence of the "Apply Changes" button in standby mode is confusing. When I see a button in the lower-right corner, I'm primed to think it means "Apply Changes" - this is the pattern set everywhere else in the frontend. I found myself instinctively trying to click it to save changes, without reading the button, Let's just hide the "Cancel"/"Discard Changes" button in standby mode, to make button placement in the frontend more consistent with assumptions we want to encourage the user to make. This also would make the Alarm Limits screen more consistent with the Quickstart screen.

I concur, I will disable it.
> 3. When I start the frontend, start ventilation without changing any alarm limits, change the alarm limits from the alarm limits page, save the changes (e.g. to SpO2 = [90 - 99], HR = [60 - 99]), navigate to the dashboard, the dashboard reflects the new AlarmLimitsRequest values. But if I then navigate back to the alarm limits page, it shows the alarm limits as the values from before saving changes (namely it shows them as SpO2 = [90 - 100] and HR = [60 - 100]). Initially the "Cancel" and "Submit" buttons are disabled, even though the values shown on the screen are different from the actual values of AlarmLimitsRequest. Once I start adjusting the alarm limits, then the enabled/disabled state behaviors of the "Cancel" and "Submit" buttons become correct with respect to the actual values of the AlarmLimitsRequest (so e.g. if I use the alarm limits buttons to change the SpO2 and HR upper limits to 99 again, then the Cancel & Submit buttons are disabled; but once I change them back to 100, then the Cancel & Submit buttons become enabled)
> 

Yep, I found this bug too, and was working on fixing that!
> Perhaps we should pause work on this PR until after we merge #340, because that PR uses AlarmLimitsRequestStandby as the only place to store unsaved changes for AlarmLimitsRequest - so it will interact with the bug I describe in item 3, and it might even fix that behavior if the "Cancel" button confirmation dialog triggers a dispatch which changes the values of AlarmLimitsRequestStandby to be overwritten with the values of AlarmLimitsRequest.

Yes, I agree, meanwhile I'll try to locally get this PR working with those changes.

rohanpurohit:
> 3. When I start the frontend, start ventilation without changing any alarm limits, change the alarm limits from the alarm limits page, save the changes (e.g. to SpO2 = [90 - 99], HR = [60 - 99]), navigate to the dashboard, the dashboard reflects the new AlarmLimitsRequest values. But if I then navigate back to the alarm limits page, it shows the alarm limits as the values from before saving changes (namely it shows them as SpO2 = [90 - 100] and HR = [60 - 100]). Initially the "Cancel" and "Submit" buttons are disabled, even though the values shown on the screen are different from the actual values of AlarmLimitsRequest. Once I start adjusting the alarm limits, then the enabled/disabled state behaviors of the "Cancel" and "Submit" buttons become correct with respect to the actual values of the AlarmLimitsRequest (so e.g. if I use the alarm limits buttons to change the SpO2 and HR upper limits to 99 again, then the Cancel & Submit buttons are disabled; but once I change them back to 100, then the Cancel & Submit buttons become enabled)
> 

This seems to be have been fixed after light testing on top of your store changes.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -114,11 +149,19 @@ export const ToolBar = ({
...
   const ventilating = useSelector(getParametersIsVentilating);
+  const alarmLimitsRequestStandby = useSelector(getAlarmLimitsRequestStandby);
+  const alarmLimitsCurrent = useSelector(getAlarmLimitsCurrent);
+  const alarmLimits = (alarmLimitsCurrent as unknown) as Record<string, Range>;
+  const alarmLimitsStandby = (alarmLimitsRequestStandby as unknown) as Record<string, Range>;
   const [isVentilatorOn, setIsVentilatorOn] = React.useState(ventilating);
   const [label, setLabel] = useState('Start Ventilation');
   const [isDisabled, setIsDisabled] = useState(false);
+  const [open, setOpen] = useState(false);
+  const setAlarmLimits = (data: Partial<AlarmLimitsRequest>) => {
```
ethanjli: I'm a bit confused by the variable naming. Just to confirm, `commitStandbyRequest` should save data into `store.controller.alarmLimits.standby` (the AlarmLimitsRequestStandby), and not `store.controller.alarmLimits.request` (the AlarmLimitsRequest). If this is what you want, maybe it's better to rename `setAlarmLimits` to `setAlarmLimitsRequestStandby`.

ethanjli (commented):
It seems to work fine in basic testing, and nothing glaring stands out in code review. For records-keeping:

ethanjli:
I tested triggering of confirmation dialog to discard unsaved changes when we attempt to go to the dashboard from the alarm limits page, and it works well. However, when we attempt to go to the settings page from the alarm limits page, this confirmation does not get triggered (but it should).

I realized another test procedure we should apply to this PR: go to the alarm limits page, change some limits, then open the events log and adjust alarm limits from there (the HR and SpO2 alarms need to be active for this to be possible), and see what the behavior of the Cancel and Submit buttons is like. With some light testing in this procedure, it already seems to work (even though we have local state controlling the enable/disable states rather than a selector), but I don't understand how it works.

rohanpurohit:
> I tested triggering of confirmation dialog to discard unsaved changes when we attempt to go to the dashboard from the alarm limits page, and it works well. However, when we attempt to go to the settings page from the alarm limits page, this confirmation does not get triggered (but it should).
> 

thanks for testing. I will try to add that!
> I realized another test procedure we should apply to this PR: go to the alarm limits page, change some limits, then open the events log and adjust alarm limits from there (the HR and SpO2 alarms need to be active for this to be possible), and see what the behavior of the Cancel and Submit buttons is like. With some light testing in this procedure, it already seems to work (even though we have local state controlling the enable/disable states rather than a selector), but I don't understand how it works.

https://github.com/pez-globo/pufferfish-software/pull/338/files#diff-a3a6688cdb1d790fc9bf6d27c0b6ff6b075351c0f1216481f61b5cb1f84cfc2eR365 it's because of this where this useEffect will check if there are any changes in Standby and current and disable/enable buttons accordingly

rohanpurohit:
@ethanjli about opening confirmation dialog on clicking settings page, I think we'll have to add that in Navigation.tsx which might complicate things, one thing to note here is even after going to the settings page, pressing the dashboard button will still shows the confirmation dialog which I think is fair enough, only other thing there is pressing `Home` in the sidebar menu will go straight to dashboard, but even then set alarms page will have the unsaved changes? so if it's ok with you I can maybe create a new issue for this and address it in a future PR?

ethanjli:
> it's because of this where this useEffect will check if there are any changes in Standby and current and disable/enable buttons accordingly

Oh! That useEffect almost looks like a selector already - unless I've missed something else in the code (which is likely), changing to a `getAlarmLimitsUnsavedChanges` selector (which could maybe return a list of the unsaved changes, so that they're also usable with the unsaved changes confirmation dialogs) should simplify the code.

> so if it's ok with you I can maybe create a new issue for this and address it in a future PR?

Yes, let's track it in a separate issue for a future PR

ethanjli:
Here's my test procedure - the code generally passes these tests, except for a weird edge case:

1. Start the simulator backend and the frontend. Set SpO2 alarm limits to [90 - 100] and HR limits to [60 - 100]. Start ventilation.
2. "Return to dashboard" button: Set SpO2 alarm limits to [90 - 95] and HR limits to [60 - 110]. When I press the topbar button to return to the dashboard, the confirmation dialogue shows the changes I'm discarding. When I press "Cancel" on the confirmation dialogue, I keep my previous values ✅. When I press "Cancel" and then "Confirm", SpO2 alarm limits are reset to [90 - 100] and HR limits are reset to [60 - 100] ✅.
3. Cancel button: Set SpO2 alarm limits to [90 - 95] and HR limits to [60 - 110]. When I press "Cancel", the confirmation dialogue shows the changes I'm discarding. When I press "Cancel" on the confirmation dialogue, I keep my previous values ✅. When I press "Cancel" and then "Confirm", SpO2 alarm limits are reset to [90 - 100] and HR limits are reset to [60 - 100] ✅.
4. Submit button: Set SpO2 alarm limits to [90 - 95] and HR limits to [60 - 110]. When I press "Submit", the confirmation dialogue shows the changes I'm submitting. When I press "Cancel" on the confirmation dialogue, I keep my unsaved values ✅. When I press "Submit" and then "Confirm", SpO2 alarm limits are changed to [90 - 95] and HR limits are reset to [60 - 110] ✅.
5. Making changes via the Active Alarms modal from the alarm limits screen with no unsaved changes: Go to the alarm limits screen (SpO2 alarm limits are [90 - 95] and HR limits are [60 - 110]), and open the Active Alarms modal where both SpO2 is too high and HR is too high alarms are active. Open the SpO2 Alarms modal, and set SpO2 alarm limits to [90 - 100]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows SpO2 alarm limits as [90 - 100], and the Cancel/Submit buttons are disabled ✅.
6. Overriding an unsaved change on the alarm limits screen from the Active Alarms modal: Go to the alarm limits screen (SpO2 alarm limits are [90 - 100] and HR limits are [60 - 110]), and make an unsaved change to set HR alarm limits to [60 - 100]. Open the Active Alarms modal where the HR is too high alarm is active. Open the HR Alarms modal, and set HR alarm limits to [60 - 120]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows HR alarm limits as [60 - 120], and the Cancel/Submit buttons are disabled ✅.
7. Making an unsaved change on the alarm limits screen, and changing a different alarm limits range from the Active Alarms modal: Go to the alarm limits screen (SpO2 alarm limits are [90 - 100] and HR limits are [60 - 120]), and make an unsaved change to set SpO2 alarm limits to [90 - 95]. Open the Active Alarms modal where HR is too high alarm is active. Open the HR Alarms modal, and set HR alarm limits to [60 - 110]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows HR alarm limits as [60 - 100], SpO2 is still shown as [90 - 95], and the Cancel/Submit buttons are still enabled ✅. But when I press the "Submit" button, it doesn't show the unsaved change to SpO2 alarm limits - the description is blank ❌. When I press the "Confirm" button, it does save the unsaved SpO2 alarm limit as [90 - 95].
8. Making an unsaved change on the alarm limits screen, overwriting it from the Active Alarms modal, and making another unsaved change on the alarm limits screen: Go to the alarm limits screen (SpO2 alarm limits are [90 - 95] and HR limits are [60 - 110]), and make an unsaved change to set SpO2 alarm limits to [90 - 100]. Open the Active Alarms modal where the SpO2 and HR too high alarms are active. Open the HR Alarms modal, and set SpO2 alarm limits to [90 - 100]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows SpO2 alarm limits as [90 - 100], and the Cancel/Submit buttons are disabled ✅. Now change SpO2 alarm limits to [90 - 95]. When I press the "Submit" button, it doesn't show the unsaved change to SpO2 alarm limits - the description is blank ❌. When I press the "Confirm" button, it does save the unsaved SpO2 alarm limit as [90 - 95].

I haven't looked at the code yet, but the way I expect the confirmation dialog to work is that it queries a selector to identify the list of all (relevant) differences between `AlarmLimitsRequestStandby` and `AlarmLimitsRequest` and shows it. I would expect this to be robust against the bugs identified in test cases 7 and 8 above. Since the code is behaving differently, there must be something else going on. I'll do a review of the code now.

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -292,41 +294,71 @@ export const AlarmsPage = (): JSX.Element => {
...
 
-  const alarmLimitsRequestStandby = useSelector(getAlarmLimitsRequestStandby);
+  const alarmLimitsRequestDraft = useSelector(getAlarmLimitsRequestDraft);
+  const alarmLimitsCurrent = useSelector(getAlarmLimitsCurrent);
   const dispatch = useDispatch();
   const currentMode = useSelector(getParametersRequestMode);
   const ventilating = useSelector(getParametersIsVentilating);
-  const setAlarmLimits = (data: Partial<AlarmLimitsRequest>) => {
+  const alarmLimitsUnsaved = useSelector(getAlarmLimitsUnsavedChanges);
+  const [alarmLimits] = useState((alarmLimitsCurrent as unknown) as Record<string, Range>);
```
ethanjli: Can we get rid of this `useState` and directly use the selector (typecasting it into a `Record<string, Range>` if needed)?

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -292,41 +294,71 @@ export const AlarmsPage = (): JSX.Element => {
...
       return;
     }
 
     dispatch(
-      commitRequest<AlarmLimitsRequest>(MessageType.AlarmLimitsRequest, alarmLimitsRequestStandby),
+      commitRequest<AlarmLimitsRequest>(MessageType.AlarmLimitsRequest, alarmLimitsRequestDraft),
     );
   };
   const alarmConfig = alarmConfiguration(currentMode);
   const [open, setOpen] = useState(false);
```
ethanjli: Is this actually supposed to mean `confirmOpen` and `setConfirmOpen`?

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -292,41 +294,71 @@ export const AlarmsPage = (): JSX.Element => {
...
   const handleConfirm = () => {
     setOpen(false);
     applyChanges();
+    setIsDisabled(true);
+  };
+
+  const handleDiscardConfirm = () => {
+    setDiscardOpen(false);
+    setAlarmLimitsRequestDraft(alarmLimits);
+    setIsDisabled(true);
```
ethanjli: If the `disabled` property of those buttons is actually determined by a selector (perhaps `getAlarmLimitsUnsavedChanges`, which I haven't reviewed yet) which tells whether there are any differences between `getAlarmLimitsRequest` and `getAlarmLimitsRequestDraft`, then we can get rid of the `isDisabled, setIsDisabled` useState, right? That way we can guarantee that even if changes are made elsewhere to `store.controller.alarmLimits.request`, such as from the breadcrumbs modal via the Events Log modal, the Cancel/Submit buttons will always be in the correct state, and we don't need to even think about the possibility of contention/overwrites between this `setIsDisabled` call here and the `useEffect` below synchronizing `setIsDisabled` with `alarmLimitsUnsaved`.

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -292,41 +294,71 @@ export const AlarmsPage = (): JSX.Element => {
...
   };
+
   const handleOpen = () => {
```
ethanjli: Does this actually mean `handleConfirmOpen`?

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -396,7 +440,61 @@ export const AlarmsPage = (): JSX.Element => {
                         <Typography variant="h4">Confirm New Changes?</Typography>
                       </Grid>
                     </Grid>
-                    <Grid item alignItems="center" className={classes.marginContent} />
+                    <Grid item className={classes.marginContent}>
+                      {alarmConfig.map((param) => {
+                        if (
+                          alarmLimits[param.stateKey].lower !==
```
ethanjli: Shouldn't these be comparing `alarmLimitsRequestDraft` against `alarmLimitsRequest`, not `alarmLimits`? Maybe that could explain why we're seeing buggy behavior in test cases 7 and 8.

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -396,7 +440,61 @@ export const AlarmsPage = (): JSX.Element => {
...
+                  <Grid container alignItems="center" justify="center">
+                    <Grid container alignItems="center" className={classes.marginHeader}>
+                      <Grid item xs>
+                        <Typography variant="h4">Keep Previous Values?</Typography>
+                      </Grid>
+                    </Grid>
+                    <Grid item className={classes.marginContent}>
+                      {alarmConfig.map((param) => {
+                        if (
+                          alarmLimits[param.stateKey].lower !==
```
ethanjli: (same thing applies here)

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -396,7 +440,61 @@ export const AlarmsPage = (): JSX.Element => {
...
+                            alarmLimitsDraft[param.stateKey]?.lower ||
+                          alarmLimits[param.stateKey].upper !==
+                            alarmLimitsDraft[param.stateKey]?.upper
+                        ) {
+                          return (
+                            <Typography variant="subtitle1">{`Keep ${param.label} alarm range to ${
+                              alarmLimits[param.stateKey].lower
```
ethanjli: (same thing applies here)

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -1,25 +1,29 @@
-import { AppBar, Button, Grid } from '@material-ui/core';
+import { AppBar, Button, Grid, Typography } from '@material-ui/core';
 import { makeStyles, Theme } from '@material-ui/core/styles';
 import React, { useCallback, useEffect, useState } from 'react';
 import { shallowEqual, useDispatch, useSelector } from 'react-redux';
-import { Link, useHistory, useLocation } from 'react-router-dom';
+import { useHistory, useLocation } from 'react-router-dom';
 import { getClockTime } from '../../store/app/selectors';
-import { commitRequest } from '../../store/controller/actions';
+import { commitRequest, commitStandbyRequest } from '../../store/controller/actions';
```
ethanjli: Let's rename `commitStandbyRequest` to `commitDraftRequest`

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -113,13 +149,22 @@ export const ToolBar = ({
   const history = useHistory();
   const currentMode = useSelector(getParametersRequestMode);
   const backendInitialized = useSelector(getBackendInitialized);
-  const parameterRequestStandby = useSelector(getParametersRequestStandby, shallowEqual);
-  const alarmLimitsRequestStandby = useSelector(getAlarmLimitsRequestStandby, shallowEqual);
+  const parameterRequestDraft = useSelector(getParametersRequestDraft, shallowEqual);
   const ventilating = useSelector(getParametersIsVentilating);
+  const alarmLimitsRequestDraft = useSelector(getAlarmLimitsRequestDraft);
+  const alarmLimitsCurrent = useSelector(getAlarmLimitsCurrent);
+  const alarmLimitsUnsaved = useSelector(getAlarmLimitsUnsavedChanges);
```
ethanjli: Similarly as above, this should be alarmLimitsRequestUnsaved (and maybe the selector should be named `getAlarmLimitsRequestUnsavedChanges` for clarity, though that makes it really long)

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -236,6 +291,16 @@ export const ToolBar = ({
     tools.push(<EventAlerts label={LOGS_ROUTE.label} />);
   }
 
+  const handleClose = () => {
```
ethanjli: Are these actually `handleDiscardClose` and `handleDiscardConfirm`?

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -113,13 +149,22 @@ export const ToolBar = ({
...
+  const alarmLimitsRequestDraft = useSelector(getAlarmLimitsRequestDraft);
+  const alarmLimitsCurrent = useSelector(getAlarmLimitsCurrent);
+  const alarmLimitsUnsaved = useSelector(getAlarmLimitsUnsavedChanges);
+  const alarmLimits = (alarmLimitsCurrent as unknown) as Record<string, Range>;
+  const alarmLimitsStandby = (alarmLimitsRequestDraft as unknown) as Record<string, Range>;
   const [isVentilatorOn, setIsVentilatorOn] = React.useState(ventilating);
   const [landingLabel, setLandingLabel] = useState('Loading...');
   const [label, setLabel] = useState('Start Ventilation');
   const [isDisabled, setIsDisabled] = useState(false);
+  const [open, setOpen] = useState(false);
```
ethanjli: Is this actually `[discardOpen, setDiscardOpen]`?

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -282,6 +347,42 @@ export const ToolBar = ({
...
+            <Grid container alignItems="center" className={classes.marginHeader}>
+              <Grid item xs>
+                <Typography variant="h4">Keep Previous Values?</Typography>
+              </Grid>
+            </Grid>
+            <Grid item className={classes.marginContent}>
+              {alarmConfig.map((param: AlarmConfiguration) => {
+                if (alarmLimits !== null && alarmLimitsStandby !== null) {
+                  if (
+                    alarmLimits[param.stateKey].lower !==
```
ethanjli: Similarly to above, shouldn't these compare `alarmLimitsRequestDraft` to `alarmLimitsRequest` rather than `alarmLimits`?

(relatedly: if the same code is being copy-pasted in 3 or 4 places, is there a way we can unify it into a reusable function, for maintainability?)

frontend/src/modules/displays/MultiStepWizard.tsx
```diff
@@ -363,6 +366,15 @@ const MultiStepWizard = (): JSX.Element => {
       const param = multiParams.find((param: Data) => param.stateKey === parameter.stateKey);
       if (param) param.alarmValues = [min, max];
       parameter.alarmValues = [min, max];
+      if (location.pathname === DASHBOARD_ROUTE.path) {
```
ethanjli, line 372: What's going on here, and what is it for?

frontend/src/store/controller/selectors.ts
```diff
@@ -196,9 +197,50 @@ export const getAlarmLimitsRequest = createSelector(
   getAlarmLimits,
   (alarmLimits: AlarmLimitsRequestResponse): AlarmLimitsRequest | null => alarmLimits.request,
 );
-export const getAlarmLimitsRequestStandby = createSelector(
+export const getAlarmLimitsRequestDraft = createSelector(
   getAlarmLimits,
-  (alarmLimits: AlarmLimitsRequestResponse): AlarmLimitsRequest | null => alarmLimits.standby,
+  (alarmLimits: AlarmLimitsRequestResponse): AlarmLimitsRequest | null => alarmLimits.draft,
+);
+const alarmLimitsCurrentSelector = (key: string) =>
```
ethanjli: I'd suggest renaming this to `rangeSelector` and making it similar to `numericParameterSelector` above: i.e. it takes two arguments, the first is a selector which would be either `getAlarmLimitsCurrent`, `getAlarmLimitsRequest`, or `getAlarmLimitsRequestDraft`, and the second is the key. That way we only need to define one function which is used to make 6 selectors for SpO2 and HR.

frontend/src/store/controller/selectors.ts
```diff
@@ -196,9 +197,50 @@ export const getAlarmLimitsRequest = createSelector(
...
+  createSelector(getAlarmLimitsCurrent, (alarmLimitsCurrent: AlarmLimits | null): Range | null =>
+    alarmLimitsCurrent === null
+      ? null
+      : ((alarmLimitsCurrent as unknown) as Record<string, Range>)[key],
+  );
+
+export const getAlarmLimitsCurrentSpo2 = alarmLimitsCurrentSelector('spo2');
```
ethanjli: Really small nitpick, but let's use `SpO2` rather than `Spo2`, to be consistent with the rest of the code.

frontend/src/store/controller/selectors.ts
```diff
@@ -196,9 +197,50 @@ export const getAlarmLimitsRequest = createSelector(
...
+    (alarmLimitsRequestDraft: AlarmLimitsRequest | null): Range | null =>
+      alarmLimitsRequestDraft === null
+        ? null
+        : ((alarmLimitsRequestDraft as unknown) as Record<string, Range>)[key],
+  );
+export const getAlarmLimitsStandbySpo2 = alarmLimitsStandbySelector('spo2');
+export const getAlarmLimitsStandbyHR = alarmLimitsStandbySelector('hr');
+
+export const getAlarmLimitsUnsavedChanges = createSelector(
+  getAlarmLimitsCurrentSpo2,
```
ethanjli: I believe these should be `getAlarmLimitsRequest...`

ethanjli (commented):
Here's a full code review. It mainly consists of:

- Identifying a possible of the strange bugs we see with unsaved changes: right now the code compares `alarmLimitsRequestDraft` against `alarmLimits` rather than comparing `alarmLimitsRequestDraft` against `alarmLimitsRequest`.
- Small nitpicks about naming of things, mostly to help the code express its intent more clearly.
- Identifying a few places where we might be able to remove `useState`

rohanpurohit:
> 7. Making an unsaved change on the alarm limits screen, and changing a different alarm limits range from the Active Alarms modal: Go to the alarm limits screen (SpO2 alarm limits are [90 - 100] and HR limits are [60 - 120]), and make an unsaved change to set SpO2 alarm limits to [90 - 95]. Open the Active Alarms modal where HR is too high alarm is active. Open the HR Alarms modal, and set HR alarm limits to [60 - 110]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows HR alarm limits as [60 - 100], SpO2 is still shown as [90 - 95], and the Cancel/Submit buttons are still enabled ✅. But when I press the "Submit" button, it doesn't show the unsaved change to SpO2 alarm limits - the description is blank ❌. When I press the "Confirm" button, it does save the unsaved SpO2 alarm limit as [90 - 95].
> 8. Making an unsaved change on the alarm limits screen, overwriting it from the Active Alarms modal, and making another unsaved change on the alarm limits screen: Go to the alarm limits screen (SpO2 alarm limits are [90 - 95] and HR limits are [60 - 110]), and make an unsaved change to set SpO2 alarm limits to [90 - 100]. Open the Active Alarms modal where the SpO2 and HR too high alarms are active. Open the HR Alarms modal, and set SpO2 alarm limits to [90 - 100]. When I press Submit and Confirm, and then I close the Active Alarms modal, now the alarm limits screen shows SpO2 alarm limits as [90 - 100], and the Cancel/Submit buttons are disabled ✅. Now change SpO2 alarm limits to [90 - 95]. When I press the "Submit" button, it doesn't show the unsaved change to SpO2 alarm limits - the description is blank ❌. When I press the "Confirm" button, it does save the unsaved SpO2 alarm limit as [90 - 95].
> 
> I haven't looked at the code yet, but the way I expect the confirmation dialog to work is that it queries a selector to identify the list of all (relevant) differences between `AlarmLimitsRequestStandby` and `AlarmLimitsRequest` and shows it. I would expect this to be robust against the bugs identified in test cases 7 and 8 above. Since the code is behaving differently, there must be something else going on. I'll do a review of the code now.

@ethanjli This should fix it fb3d99797065a6b7f5417f80befa7324120071af
about the confirmation dialog quering a selector, I think what we have currently is best, since we need to see which of the two `Spo2` or `HR` is changed and pass that dynamically with the stateKey, might be complicated if we try to use selector there

ethanjli:
For simplicity, I would suggest that the selector should just return a list of keys whose respective values are different between `alarmLimitsRequest` and `alarmLimitsRequestDraft`. Then you would just use that as the input for the `.map` operation, which can check the keys against `AlarmConfiguration`. This way the somewhat more complicated logic is all inside a selector (and there are several places we would reuse it), and the rendering stuff is in the component.

frontend/src/modules/displays/MultiStepWizard.tsx
rohanpurohit: so the bug that dispatch caused is:
- go to set alarms page, change any one of the alarm limits to anything, basically unsaved changes
- open alarms modal, (eg: HR too high) and change alarm limits for HR and press `cancel` and confirm
- pressing cancel will discard the changes, which would mean that the unsaved changes on the alarms page also gets discarded

frontend/src/store/controller/selectors.ts
rohanpurohit: I think the key in mcu_pb is 'spo2' right?

frontend/src/modules/alarms/AlarmsPage.tsx
rohanpurohit: yes. done

frontend/src/modules/alarms/AlarmsPage.tsx
rohanpurohit: Done.

frontend/src/modules/displays/MultiStepWizard.tsx
rohanpurohit: basically, we were dispatching only for HFNC control to show the unsaved alarm limit changes so maybe it's safe to do this check?

rohanpurohit:
Hi @ethanjli if it's possible can you review the changes to the selectors? so what I'm doing now is taking your suggestion created a RangeSelector that gives out alarmLimits/current or draft as Record<string, Range> which I can re-use in `AlarmsPage` and anywhere else, then taking those I compare it in the other selector which then returns a `string[]` (maybe this one can be improved) and finally checking if the returned string[] is empty or not and that would be the unsaved changes. Also I'm not sure about the naming should it be `getAlarmLimitsRequestUnsavedKeys` similarly for all, in some cases then it's too long.
thanks!

frontend/src/modules/alarms/AlarmsPage.tsx
rohanpurohit: yep I agree, using the selector now and removed the useEffect

frontend/src/modules/alarms/AlarmsPage.tsx
rohanpurohit: I'm resolving for now as I'm using a selector now, if there is a better way to do that I will fix it there.

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: done.

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: using getAlarmLimitsRequestUnsaved instead

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: > (relatedly: if the same code is being copy-pasted in 3 or 4 places, is there a way we can unify it into a reusable function, for maintainability?)

i think this should be fine now as I'm using a selector? I'm resolving it for now.

frontend/src/modules/displays/MultiStepWizard.tsx
ethanjli: I see, it'd be great if you could add a code comment summarizing this. What happens if we open the breadcrumbs modal from the event log while we're on the settings page, and try to do the same test?

frontend/src/store/controller/selectors.ts
```diff
@@ -201,46 +201,45 @@ export const getAlarmLimitsRequestDraft = createSelector(
...
-    alarmLimitsCurrent === null
-      ? null
-      : ((alarmLimitsCurrent as unknown) as Record<string, Range>)[key],
+type OptionalAlarmLimitsLike = AlarmLimits | AlarmLimitsRequest | null;
+type AlarmLimitsSelector = OutputSelector<
+  StoreState,
+  AlarmLimits | null,
+  (res: AlarmLimitsRequestResponse) => OptionalAlarmLimitsLike
+>;
+const rangeSelector = (alarmLimitsSelector: AlarmLimitsSelector) =>
```
ethanjli: What I had proposed was that `rangeSelector` would take two arguments, the first is a selector which would be either getAlarmLimitsCurrent, getAlarmLimitsRequest, or getAlarmLimitsRequestDraft, and the second is the key. However, if you no longer need getAlarmLimits*SpO2 and getAlarmLimits*HR, then we don't actually need `rangeSelector`. Right now all it does is cast your `OptionalAlarmLimitsLike` into a `Record<string, Range> | null`, and we don't need a selector to do that.

frontend/src/store/controller/selectors.ts
```diff
@@ -201,46 +201,45 @@ export const getAlarmLimitsRequestDraft = createSelector(
...
-export const getAlarmLimitsStandbyHR = alarmLimitsStandbySelector('hr');
-
-export const getAlarmLimitsUnsavedChanges = createSelector(
-  getAlarmLimitsCurrentSpo2,
-  getAlarmLimitsCurrentHR,
-  getAlarmLimitsStandbySpo2,
-  getAlarmLimitsStandbyHR,
+export const getAlarmLimitsUnsavedKeys = createSelector(
+  getAlarmLimitsRangeCurrent,
+  getAlarmLimitsRangeDraft,
```
ethanjli: Shouldn't you be using `getAlarmLimitsRequest`/`getAlarmLimitsRangeRequest` rather than `getAlarmLimitsCurrent`/`getAlarmLimitsRangeCurrent`? When we talk about "unsaved changes", we mean that the frontend has not yet sent the changes to the backend, by including the changes into `store.controller.alarmLimits.request`. It's a different matter whether the firmware has accepted those saved changes, by including them into `store.controller.alarmLimits.current`.

Here you might as well just pass in `getAlarmLimitsRequest`/`getAlarmLimitsRequestCurrent` and `getAlarmLimitsRequestDraft`, and cast them into `Record<string, Range>` inside the selector. I'd suggest having an `if` check at the start of the selector to return an empty array early if either `getAlarmLimitsCurrent` or `getAlarmLimitsRequestDraft` is null, so that you can take the `currentLimits !== null` and `draftLimits !== null` checks out of the `keys.filter` operation. If both are not null, then you can cast them into `Record<string, Range>`s.

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -330,13 +331,11 @@ export const AlarmsPage = (): JSX.Element => {
...
     setConfirmOpen(false);
     applyChanges();
-    setIsDisabled(true);
   };
 
   const handleDiscardConfirm = () => {
     setDiscardOpen(false);
-    setAlarmLimitsRequestDraft(alarmLimits);
-    setIsDisabled(true);
+    if (alarmLimits !== null) setAlarmLimitsRequestDraft(alarmLimits);
```
ethanjli: Shouldn't we be resetting `store.controller.alarmLimits.draft` from `store.controller.alarmLimits.request`, rather than from `store.controller.alarmLimits.current`?

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -363,12 +365,8 @@ export const ToolBar = ({
             </Grid>
             <Grid item className={classes.marginContent}>
               {alarmConfig.map((param: AlarmConfiguration) => {
-                if (alarmLimits !== null && alarmLimitsStandby !== null) {
-                  if (
-                    alarmLimits[param.stateKey].lower !==
-                      alarmLimitsStandby[param.stateKey]?.lower ||
-                    alarmLimits[param.stateKey].upper !== alarmLimitsStandby[param.stateKey]?.upper
-                  ) {
+                if (alarmLimits !== null && alarmLimitsDraft !== null) {
```
ethanjli: Is this line actually needed? In AlarmsPage.tsx it seems to have been omitted.

frontend/src/modules/alarms/AlarmsPage.tsx
```diff
@@ -478,17 +469,14 @@ export const AlarmsPage = (): JSX.Element => {
...
+                        if (alarmLimitsUnsavedKeys.includes(param.stateKey)) {
                           return (
                             <Typography variant="subtitle1">{`Keep ${param.label} alarm range to ${
-                              alarmLimits[param.stateKey].lower
+                              alarmLimits !== null && alarmLimits[param.stateKey].lower
                             } -
-                                ${alarmLimits[param.stateKey].upper}?`}</Typography>
+                                ${
+                                  alarmLimits !== null && alarmLimits[param.stateKey].upper
+                                }?`}</Typography>
```
ethanjli: Shouldn't these be using `getAlarmLimitsRequest` rather than `getAlarmLimitsCurrent`?

ethanjli (commented):
Here's my review of the changes.

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: Yes! Toolbar is actually on landing page too, and I think it would be null initially and frontend crashes saying these are null which is not the case in AlarmsPage

frontend/src/store/controller/selectors.ts
rohanpurohit: Yep i realised later that we don't really need getAlarmLimitsSpO2 or HR, I have removed the `rangeSelector `now

frontend/src/store/controller/selectors.ts
rohanpurohit: yeah I did not realize the difference until now, using getAlarmLimitsRequest now

frontend/src/modules/displays/MultiStepWizard.tsx
rohanpurohit: Sure will do that!

rohanpurohit: > I see, it'd be great if you could add a code comment summarizing this. What happens if we open the breadcrumbs modal from the event log while we're on the settings page, and try to do the same test?

Seems to be fine on light testing.

rohanpurohit:
@ethanjli made the suggested changes and renamed a few things, looks good on testing on my end.

frontend/src/modules/app/ToolBar.tsx
ethanjli: Ah, ok!

rohanpurohit:

frontend/src/modules/displays/MultiStepWizard.tsx
ethanjli: Ok. The specific outcome I'd look for in that test case is that you press cancel to discard the changes from the breadcrumbs modal while on the Settings screen; and then if you go to the Alarm Limits screen or open an Alarm Limits modal, does it load with the unsaved changes (which would be undesirable)? That's what I'd expect to happen based on my interpretation of your explanation for this code.

ethanjli (approved):

ethanjli:
https://github.com/pez-globo/pufferfish-software/pull/338#discussion_r633041093 is the only remaining unresolved discussion, but I don't consider it a blocker, so I'm approving this PR for merging.

frontend/src/modules/displays/MultiStepWizard.tsx
rohanpurohit: I think the primary issue here is unsaved changes in the alarms page and unsaved changes in breadcrumbs modal are different, or rather unsaved changes in alarms page are not propagated to breadcrumbs modal ( they probably look at current/set alarm limits), so when I say I'm discarding unsaved changes in breadcrumbs modal it means the temporary changes made there, which are currently different from the changes on alarms page, and discarding temporary changes in breadcrumbs should not discard the `different` changes on alarms page from my understanding. I'll merge for now and keep this unresolved.

ethanjli: Oh ok, so the breadcrumbs modal isn't loading/saving alarm limits changed from/to `store.controller.alarmLimits.draft`? That would explain why it behaves the way it does, then. But then how is the MultiStepWizard's top-level view (which displays all alarm limits and settings, including unsaved changes to alarm limits) showing the unsaved changes?

rohanpurohit: > Oh ok, so the breadcrumbs modal isn't loading/saving alarm limits changed from/to `store.controller.alarmLimits.draft`?

I think so.

> But then how is the MultiStepWizard's top-level view (which displays all alarm limits and settings, including unsaved changes to alarm limits) showing the unsaved changes?

Do you mean the HFNC control? if so then it uses a selector to show changes.

ethanjli: Yes, HFNC control. So here's my interpretation of what you're saying:

1. The breadcrumbs modal stores its unsaved changes into local state, not into store.controller.alarmLimits.draft (looking with a fresh mind at the code in MultiStepWizard.tsx, I see where this happens, and now I see where it gets committed to store.controller.alarmLimits when a submit is confirmed, and where it gets reset when a discard is confirmed)
2. The "HFNC Control" screen in the breadcrumbs modal shows unsaved changes for alarm limits by getting them from the selector for store.controller.alarmLimits.draft, not by checking the local state (looking at the code for MultiStepWizard.tsx, it looks like this is done from the HFNCControls component's ValueInfo components, which don't have access to the local state of the breadcrumbs modal)

It looks like the dispatch you are making here is so that the local changes get copied into store.controller.alarmLimits.draft (but only when we're on the dashboard) so that unsaved changes become visible via the selector used by ValueInfo. Now I realize that I didn't fully understand your first two comments. So if I'm understanding everything correctly, then the HFNCControls screen will not show the unsaved changes if alarm limits from the breadcrumbs modal if the user is outside the dashboard screen - this would be consistent with the comment you recently added to #308. This seems like buggy behavior. Perhaps the more robust solution is to allow specifying the source of the alarm limits as props on ValueInfo and HFNCControls, so that MultiStepWizard can pass its local state into them. This should completely remove the need for dispatching unsaved changes of alarm limits into store.controller.alarmLimits.draft. We can test whether this buggy behavior exists as I'm predicting, and if so how to fix it, as part of #338.

rohanpurohit: Yes the bug exists, i was more concerned about not discarding unsaved changes in alarms page, which was the wrong approach.

> Perhaps the more robust solution is to allow specifying the source of the alarm limits as props on ValueInfo and HFNCControls, so that MultiStepWizard can pass its local state into them. This should completely remove the need for dispatching unsaved changes of alarm limits into store.controller.alarmLimits.draft. We can test whether this buggy behavior exists as I'm predicting, and if so how to fix it, as part of #338.

Yes I agree that it's a robust approach, but I feel since we are already using selectors in ValueInfo it's still a good approach to try and fix dispatching with conditions, if we pass the local state then we might need to refactor ValueInfo to not use selectors in that case? I have opened a PR to fix the bug introduced in #338 and also the same for parameters using a dispatch and selector for ParametersRequestDraft, I have only lightly tested it so far, if it still causes unwanted regressions i will work on the approach you suggested.

PR #339, opened 2021-05-03
Disable start button in frontend until backend becomes available

### LINKED ISSUE #325: Landing Screen should only allow going to Quick Start once backend is available
Right now the landing screen allows the user to bypass the "start" button by clicking on the sidebar; we should remove the sidebar. Additionally, the "start" button always works, even if the backend is not available - in which case pressing the "start" button takes the user to the Quick Start screen and immediately triggers the "backend connection lost!" alarm. We should prevent the "start" button from allowing this. The simple approach we discussed was to disable the "start" button and make it say something like "Loading..." or "Connecting..." instead of "Start" while the backend is unavailable.

(this issue was created from https://github.com/pez-globo/pufferfish-software/pull/313#issuecomment-818406959 )

rohanpurohit:
… (#325)


ethanjli:
Right now the behavior I observe on the landing page is:

- If I start the frontend without the backend running, the landing page briefly flashes "Start" and then (within ~500 ms, based on the timeout you set) switches to "Loading"
- If I start the frontend after the backend is running, the landing page briefly flashes "Start", then changes to "Loading", then switches back to "Start"

I see this behavior on the landing page, but not on the quickstart screen.

I believe the reason this behavior happens is that you assume that the backend is connected at the start, then you wait to see if the backend is lost based on a timeout which fires "BACKEND_CONNECTION_LOST". Maybe it would be cleaner if the button started as "Loading" by default, and then a "BACKEND_CONNECTION_INITIALIZED" event were fired which changed it to "Start"?

Minor feedback: let's have the button say "Loading...", rather than just "Loading".

ethanjli:
Note related to futureproofing for the first task I'm working on this week (making the frontend initialize its store's ParametersRequest, AlarmLimitsRequest, etc., as `undefined` and wait for the backend to give it the initial values): (ignore everything in italics, we discussed a better solution in our meeting) _eventually we will want BACKEND_CONNECTION_INITIALIZED to be dispatched once the store's ParametersRequest, AlarmLimitsRequest, etc., are no longer undefined. It would be helpful if you could create an action for BACKEND_CONNECTION_INITIALIZED in such a way that I can go back in and modify its trigger conditions, e.g. using a selector._ The approach we discussed trying out is just to make a selector which decides whether the backend connection is initialized, based on the values of things in the store.

rohanpurohit:
Still working on it! has known issues, pushed to discuss approach and code review

rohanpurohit:
TODO: 
- add EventLog for backend_connection_up , also change naming for `backend_connection_down `which is currently `backend_connection_lost_code` 
- figure out a robust way to handle toolbar logic for `start/pause` button (problem: no robust way of figuring out if we are on landing page, currently we just send a boolean if start button is clicked on landing page which is `staticStart`, but we need to figure out that we are on landing page before that)
- useEffect that handles dispatching of `BACKEND_CONNECTION_DOWN `and `BACKEND_CONNECTION_UP` needs more refactoring?, does not work if it's in a single `useEffect` with an if-else block (maybe because it is only scheduled once?)
- there is some race condition on START and Pause ventilation from quickstart to dashboard (linked to toolbar logic)

@ethanjli  any thoughts overall?

rohanpurohit:
@ethanjli This PR is ready for testing and code review.

frontend/src/store/app/actions.ts
```diff
@@ -7,3 +13,11 @@ export function initialize(): AppAction {
 export function setLocale(locale: string): AppAction {
   return { type: SET_LOCALE, locale };
 }
+
+export function lostBackendConnection(backendConnection: boolean, clock: Date): AppAction {
+  return { type: BACKEND_CONNECTION_DOWN, clock, backendConnection };
```
ethanjli: Do we actually Need the "backendConnection" boolean field in these actions? What do the following actions mean?
```
{
  type: BACKEND_CONNECTION_DOWN,
  backendConnection: true
}
```
```
{
  type: BACKEND_CONNECTION_DOWN,
  backendConnection: false
}
```

(this also applies to the definitions of the `BackendConnectionDown` and `BackendConnectionUp` interfaces in `types.ts`)

frontend/src/store/app/reducers.ts
```diff
@@ -31,6 +34,10 @@ export function appReducer(state = initialState, action: AppAction): AppState {
       // TODO: Redux recommends **against** using new date inside reducers; instead
       // it should come from the code which creates the action to dispatch.
       return { ...state, backendHeartbeat: new Date() };
+    case BACKEND_CONNECTION_DOWN:
+      return { ...state, clock: new Date(), backendConnection: action.backendConnection };
```
ethanjli: Could we just set `backendConnection: false` for `BACKEND_CONNECTION_DOWN` and `backendConnection: true` for BACKEND_CONNECTION_UP`?

frontend/src/store/app/selectors.ts
```diff
@@ -31,6 +31,14 @@ export const getAlarmNotifyStatus = createSelector(
   getApp,
   (app: AppState): boolean => app.notifyAlarm,
 );
+export const getBackendHeartBeat = createSelector(
+  getApp,
+  (app: AppState): Date => app.backendHeartbeat,
+);
+export const getBackendDisconnected = createSelector(
+  getApp,
+  (app: AppState): boolean => app.backendConnection,
```
ethanjli, line 40: From naming, I would assume that `app.backendConnection` represents whether the backend is connected, while a `getBackendDisconnected` selector represents the opposite of that - i.e. shouldn't `getBackendDisconnected` return `!app.backendConnection`?

frontend/src/store/controller/selectors.ts
```diff
@@ -233,6 +234,14 @@ export const getPopupEventLog = createSelector(getController, (states: Controlle
   return states.eventLog.nextLogEvents.elements.find((el: LogEvent) => el.id === maxId);
 });
 
+// Backend Initialized
+export const getBackendInitialized = createSelector(
+  getSensorMeasurements,
+  getBackendDisconnected,
+  (sensorMeasurements: SensorMeasurements | null, backendDisconnected: boolean): boolean =>
+    sensorMeasurements !== null && backendDisconnected,
```
ethanjli: Based on function names, the logic expressed in this selector is that the backend is initialized if both of the following conditions are true:
- sensorMeasurements is not-null
- the backend is disconnected.
Based on this, and how you implemented the `getBackendDisconnected` selector, maybe you actually meant to name the selector `getBackendConnected` rather than `getBackendDisconnected`?

ethanjli: While we're at it, let's change `sensorMeasurements !== null` to `parametersRequest !== null && alarmLimitsRequest !== null` in this PR.

frontend/src/modules/app/OverlayScreen.tsx
```diff
@@ -36,34 +37,41 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  const events = useSelector(getNextLogEvents);
+  const diff = Math.abs(new Date().valueOf() - new Date(heartbeat).valueOf());
+
+  useEffect(() => {
+    if (diff < BACKEND_CONNECTION_TIMEOUT) {
+      dispatch(establishedBackendConnection(true, new Date()));
+    }
+  }, [clock, diff, dispatch, heartbeat]);
 
   useEffect(() => {
```
ethanjli: Since these just cover different branches of the same `if` check (i.e. `diff <= BACKEND_CONNECTION_TIMEOUT`), would it make sense to merge these two into the same `useEffect`? Like:
```
if (diff <= BACKEND_CONNECTION_TIMEOUT {
  ...
} else if (!lostConnectionAlarm) {
  ...
}
```

ethanjli (commented):
It works in basic testing, here's a review of the code.

rohanpurohit:

ethanjli (approved):
I tested the new changes, it works well!

PR #340, opened 2021-05-05
Use backend values to initialize frontend

### LINKED ISSUE #312: Frontend shouldn't initialize values

ethanjli:
This PR fixes #312 by making the frontend initialize all backend-derived messages in the store to `null` on startup, instead of initializing with numbers or default values. Because implementing this required changing the types produced by selectors for those messages, as well as changing the reducers for `ParametersRequest`, `AlarmLimitsRequest`, and their standby versions (mainly because those reducers previously had inconsistent implementations and were difficult for me to understand), it led to a large refactor of the store-related code in the frontend (including improvements in organization, naming), with the following main changes:

- Use selectors and action dispatch into standby requests for displaying and changing values from the quickstart and alarm limits screens. Previously the quickstart screen mainly relied on state internal to the components, while the alarm limits screen was a weird hybrid where it relied on state but was also dispatching every change into the standby message.
- Consolidate/unify reducers for all Request messages and standby Request messages into one generic reducer. Previously there were multiple copies of basically the same code, but with a few strings changed, and also they didn't behave consistently with each other - the copies of code had diverged for some reason.
- Improve type constraints & transparency of actions for committing to requests and standby requests. Previously there were two action generators, updateCommittedParameter (which generated a single `@controller/PARAMETER_COMMITTED` action) and updateCommittedState (which could generate a infinite number of possible actions with the template `@controller/{string}_COMMITTED`), except one could theoretically also use updateCommittedState in place of updateCommittedParameter, and action types were unnecessarily [stringly typed](https://wiki.c2.com/?StringlyTyped). Now we have exactly two actions: commitRequest and commitStandbyRequest, and the message type is specified as a `MessageType` rather than a string, and the action generators work with Typescript to ensure that the data provided with the action matches the type constraints for the provided message type.
- Group related parts of store/controller into objects. Previously the names of the various parts of the store were very long camelcase words (e.g. `alarmLimits`, `alarmLimitsRequest`, and `alarmLimitsRequestStandby`), now they're broken up into fragments and organized in a more logical way (e.g. `alarmLimits.current`, `alarmLimits.request`, and `alarmLimits.standby`) in a pattern established by a reorganization I did of the event log-related parts of the store in #313, which moved `nextLogEvents`, `expectedLogEvent`, and `activeLogEvents` into `eventLog`. This reorganization allows us to make generic, composable selectors instead of copying the same logic copied over and over. Previously, updating the code would've required way too much copy-pasting, and updating the code would've been error-prone (forgetting to update something in one place). Now the copy-pasting is restricted to copying a function call and just changing its arguments.
- Reorder the order of selectors, types, and reducers to be more consistent across files. This should reduce the amount of effort needed to find where any given function is in a file.
- Move `BACKEND_HEARTBEAT` & `BACKEND_CONNECTION_LOST` store logic from controller to app. This is in preparation for moving all connection-related logic into its own slice of the store.

Additionally, now the backend sends the following messages to the frontend: AlarmMuteRequest, BatteryPower, ScreenStatus, RotaryEncoder, SystemSettingRequest, and FrontendDisplaySetting. However, the backend doesn't simulate these messages or load them from the filesystem during initialization; instead, the backend simulator merely initializes some default values; the frontend does indeed receive them from the backend.


ethanjli:
@rohanpurohit This PR will almost certainly produce merge conflicts with your other open frontend PRs, we'll need to coordinate on how to merge things - i.e. for PR of yours which we merge into develop first, I'll need to make sure I understand every changed line of code, as I'll probably need to manually "replay" all of your changes on top of my changes. And for any PR of yours which we merge into develop after this PR, you'll need to understand what I changed and what assumptions you may have made which might be violated by the changes I've made in this PR.

ethanjli:
@rohanpurohit Since this PR touches a bunch of files, it'd be great if you could test all aspects of the frontend to help look for any user-facing regressions introduced by my refactoring.

rohanpurohit:
> @rohanpurohit This PR will almost certainly produce merge conflicts with your other open frontend PRs, we'll need to coordinate on how to merge things - i.e. for PR of yours which we merge into develop first, I'll need to make sure I understand every changed line of code, as I'll probably need to manually "replay" all of your changes on top of my changes. And for any PR of yours which we merge into develop after this PR, you'll need to understand what I changed and what assumptions you may have made which might be violated by the changes I've made in this PR.

yeah sounds good, I did not find time yesterday to work on my PR'S due to emergency reasons, I'm working on them now and  I think both approaches are sound, if you are done with the changes on this PR, then I can go through them to try and understand that and make changes accordingly once this PR gets merged.

rohanpurohit:
> @rohanpurohit Since this PR touches a bunch of files, it'd be great if you could test all aspects of the frontend to help look for any user-facing regressions introduced by my refactoring.

Yep, I'll definitely do that.

rohanpurohit:
so in my light to moderate testing, everything seems fine on alarm modals, dashboard, and other related features. Although mute button seems to not work anymore, triggered by stopping the backend.

ethanjli:
Great, thanks for catching that! I had forgotten to test alarm muting functionality. I'm able to reproduce the regression. I think it's fine if we leave this up to a future PR in one of the upcoming milestones, because there's some nuance we'll have to figure out there, and it will be easier to do so after #339 gets merged:
- Alarm muting state should be controlled by the firmware. The frontend should merely issue an alarm muting request to the backend (which forwards it to the firmware) - since right now our backend simulator and our frontend don't support alarm muting, it makes sense that alarm muting doesn't work after this PR (because this PR should have made alarm muting work just like any other request/response message pair)
- But when the backend is lost, the frontend needs to be able to mute the "Backend connection lost" alarm. I think ultimately we'll need an "getAlarmMuting" selector which checks both the AlarmMute response message from the backend and the backend_disconnected part of the app store, and maybe also another variable for the app store (e.g. local_alarm_mute_requested), and from those three variables decides whether to mute alarms in the frontend. This is a similar approach to what we're doing in #339. Can you create a Github issue to track this?

I will merge develop back into this PR and then merge this PR into develop as prototype-quality code which has been moderately tested, with known issues to be fixed in future PRs but no regressions identified besides alarm muting functionality. For records-keeping:

PR #348, opened 2021-05-10
Add Alarm muting functionality in firmware and backend/simulator

### LINKED ISSUE #346: Buggy mute alarm button behaviour
Currently, there is inconsistent behavior of muting the alarm when ventilation is restarted or when we navigate to a different page from the dashboard, the behavior is consistent on the dashboard.

steps to reproduce:

1. start ventilation 
2. wait for an alarm to be triggered
3. pause ventilation and start ventilation again
4. wait for alarm if not active
5. press the mute button, the mute button needs to be toggled twice to stop the alarm sound.

This behavior can also be noticed on the landing screen when the backend is lost. again it's inconsistent.

Additional context:
- only tested with backend running simulator

### LINKED ISSUE #347: Alarm muting
Right now alarm muting isn't handled by the backend or firmware. We should implement backend simulator and firmware support for it. The mute needs to automatically return after 2 minutes if not manually returned (see https://github.com/pez-globo/pufferfish-software/pull/348#issuecomment-849206700 for more discussion)

rohanpurohit:
This PR includes:
- Add AlarmMuteService in backend simulator to handle muting of alarms
- port simulator code to the firmware.
- make changes in store in firmware and backend for the same
- Allow muting backend connection lost alarm on the frontend
- fix inconsistent mute button behaviour

As stated in #346 there are few issues with mute button behavior on develop (when alarm muting worked) : 
- Inconsistent behavior with alarm sound in general, if a current alarm is muted once, any new alarm that is triggered should be muted by default as well until the timeout.
- Inconsistent behaviour when ventilation is toggled, if alarm is muted before pausing ventilation, it should remain muted once ventilation is restarted, that is not the case each time.
- There was inconsistent behavior with red border when a new alarm is triggered while an alarm was already active, to reproduce: 
mute an active alarm -> wait for a new alarm to be triggered or trigger one manually (spo2 from 100-99) -> audio shouldn't play but it does, then to mute press the button -> audio stops playing but red border is still visible, have to toggle back n forth to fix the issue! 

This PR fixes above issues:
- If an active alarm is muted, any new alarm that is triggered will be muted by default.
- fixes red border issue stated above
- now uses selector directly instead of local state `playing` to play/pause the audio
- To fix Muting the backend connection lost alarm,  check for both backend connected status from `getBackendConnected` selector and the `AlarmMuteRequest`, which is still dispatched even thought there is no backend to forward it to toggle playing the sound.


ethanjli (commented):
Generally looks good. What was the reasoning to put put the functionality for handling the AlarmMuteRequest/AlarmMute request-response service as part of the Alarms service, rather than in a separate service? Are their behaviors required to be closely coupled, or can we separate them? It feels pretty awkward to be passing ~7 parameters into the Alarms service.

rohanpurohit:
@ethanjli I remember we had discussed to keep the alarm mute service in the backend simulator similar to that in the firmware, but it seems like I'm not able to get that working. the pushed file in this branch follows the same logic of having a service like other files and that does work, But I guess mute shouldn't really depend on active service? can you point out problems if any in the first approach?
changes: 
- [alarm_mute.py](https://github.com/pez-globo/pufferfish-software/blob/44e13fa38b7d9368c53abedeff60a32402a2fe4e/backend/ventserver/simulation/alarm_mute.py)
- [simulator.py](https://github.com/pez-globo/pufferfish-software/pull/348/commits/44e13fa38b7d9368c53abedeff60a32402a2fe4e#diff-f59dd5a5c28c0f96dbb985e071a3913eec16a080cd0524161d5b11497beae2ae)

ethanjli:
Your alarm_mute.Service has an `active_alarm_ids` member which is never changed from its initial value. I would suggest making `active_alarm_ids` be an input parameter of `alarmMute.Service.transform`, rather than having it as a member. Right now I would expect the behavior to be that the backend ignores any `AlarmMuteRequest`.

rohanpurohit:
@ethanjli I haven't tested the firmware side yet, but I have updated the PR description to state the changes to Audio and alarm mute button behavior, and also muting backend connection lost alarm in frontend. I have also updated the backend simulator (for code-review), If it's possible maybe you can test these changes with the simulator? thanks!

ethanjli:
TODO from me before merging: test this branch on the Raspberry Pi with the simulator backend
TODO from me: also test this branch on my computer with the firmware

ethanjli:
Observations from testing:

- On my computer, muting behavior works as expected in the simulator and frontend (I checked both the visual border and the audio playback).
- On the Raspberry Pi, muting behavior works as expected in the simulator and frontend (I could only confirm this the visual border, because for some reason my RPi isn't playing any audio at all, even from other websites like YouTube).
- On my computer, muting behavior does not work in the firmware with the frontend. I think this is because the implementation of `AlarmMuteService` hasn't been fixed yet.
- In either this PR or a future PR (I would suggest a future PR), the AlarmMuteService in the simulator backend and in the firmware, as well as whatever mechanism allows muting in the frontend while the backend is disconnected, will need to update a countdown timer (counting down from 2 min) which gets displayed in the frontend where now it always says "2:00" , and the service implementations will need to set alarmMute.active back to false after the two minutes have elapsed.
- In a future PR, we may need to disable the mute button while the firmware is disconnected, to prevent weird behavior.

rohanpurohit:
> * On the Raspberry Pi, muting behavior works as expected in the simulator and frontend (I could only confirm this the visual border, because for some reason my RPi isn't playing any audio at all, even from other websites like YouTube).

same here. maybe we need to connect a Bluetooth speaker and test it? I don't have one currently so couldn't do that.
> * On my computer, muting behavior does not work in the firmware with the frontend. I think this is because the implementation of `AlarmMuteService` hasn't been fixed yet.

yes I have fixed it now and visually it works, also confirmed with the redux store and printing alarm_mute from the mcu.receive event. can you also test the same @ethanjli thanks!
> * In either this PR or a future PR (I would suggest a future PR), the AlarmMuteService in the simulator backend and in the firmware, as well as whatever mechanism allows muting in the frontend while the backend is disconnected, will need to update a countdown timer (counting down from 2 min) which gets displayed in the frontend where now it always says "2:00" , and the service implementations will need to set alarmMute.active back to false after the two minutes have elapsed.

oh okay! I will do this in a future PR, just to note after fixing the firmware part, it shows "0:00" instead (only when backend is connected to application, "2:00" on the simulator") haven't looked at why just yet. just writing it down here for future reference. 
> * In a future PR, we may need to disable the mute button while the firmware is disconnected, to prevent weird behavior.

Yes, I saw this in your comment on the issue. I guess we can keep that issue open and note these down there?

ethanjli:
I can also toggle alarm muting with the firmware. I suspect that the "countdown time" for the firmware's copy of the `AlarmMute` message has not been initialized to 120 seconds, so the firmware makes the frontend that the countdown is 00:00; by contrast, the simulator backend initializes `AlarmMute.remaining` as 120. I recommend that you initialize `AlarmMuteRequest` and `AlarmMute` in the `initialize_states` function in the firmware's `main.cpp`, following the same pattern used for `AlarmLimits`.

rohanpurohit:

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -0,0 +1,50 @@
...
+from ventserver.protocols.backend import states
+from ventserver.protocols.protobuf import mcu_pb
+
+def transform(
+    active_alarm_ids: mcu_pb.ActiveLogEvents,
+    request: mcu_pb.AlarmMuteRequest,
+    response: mcu_pb.AlarmMute
+) -> None:
+    """Mute the current active alarm"""
+    if active_alarm_ids is None:
```
ethanjli: `if active_alarm_ids is None` does not catch the case where active_alarm_ids is a message with field `active_alarm_ids.id == []`. Thus, this branch will not be entered in certain situations where you want to enter this branch.

Since the firmware doesn't perform this check, and since there might be some use cases where someone would want to preemptively disable alarms for 2 minutes (e.g. they're about to temporarily disconnect the patient to perform a procedure and don't want alarms during that time), maybe we should just remove this check entirely.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+#include "Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h"
+
+// AlarmMuteService
+
+namespace Pufferfish::Driver::BreathingCircuit {
+
+void AlarmMuteService::transform(
+    AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request) {
+  alarm_mute.active = alarm_mute_request.active;
```
ethanjli, line 16: Right now the simulator backend returns early if there are no active alarms, while the firmware never returns early. Whichever approach we take, I think the Alarm Mute Service should have consistent behavior between the backend and firmware.

ethanjli (commented):
The firmware seems to work fine on my end, with light testing! However, I just realized that I haven't done a full code review for the most recent version of the code with various things fixed.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
rohanpurohit: Yeah I removed the check as it was the case like you had suggested that active_alarms_ in firmware was always empty and it returned early, I'll remove it in the simulator too now !

ethanjli (approved):

PR #351, opened 2021-05-14
Initialize frontend rotary encoder state from the backend

### LINKED ISSUE #349: PR #340 caused regression which makes frontend ignore the rotary encoder
In commit 4979d001f884cbc3303a0ad1e410a15d0e0eae66, the frontend responds to the rotary encoder; in commit 7e93fa80f388ff27b28e93bf266c060081fe7f4f, it doesn't. This is probably because `rotaryEncoderReducer` does not actually handle the case when `oldState == null` and the new state provided the action is non-null.

ethanjli:
This PR fixes #349 by making the frontend's rotary encoder reducer handle the case where its state is null (which is always how the frontend starts) and it receives a RotaryEncoder message from the backend. This is a very tiny hotfix and I've tested it on my Raspberry Pi setup; @raavilagoo once you test it and confirm that the frontend responds when you turn the rotary dial, I'll merge in this PR. For records-keeping:

### DIFF
--- a/frontend/src/store/controller/reducers/backend.ts
+++ b/frontend/src/store/controller/reducers/backend.ts
@@ -158,11 +158,14 @@ export const rotaryEncoderReducer = (
       }
 
       const newState = action.state as RotaryEncoder;
+      if (state === null) {
+        // initialize the store with the backend's value
+        return { ...newState, stepDiff: 0 };
+      }
+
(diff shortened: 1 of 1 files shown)


raavilagoo:
Rotary encoder test passed on this commit. Frontend responds to Rotary dial.

PR #357, opened 2021-05-14
Manually resolve dependabot upgrades

ethanjli:
This PR addresses dependabot alerts about vulnerabilities in lodash and postcss by upgrading the packages in yarn.


ethanjli:
I've tested this on the Raspberry Pi and it works there, so I'm merging this hotfix in.

PR #360, opened 2021-05-17
frontend: show correct values in HFNCControl

rohanpurohit:
This PR fixes the remaining parts of #307  :
- Fixes bug introduced in #338 where unsaved changes from alarm limits modal only showed in the hfnc control in dashboard,
This bug was fixed by checking if multiPopUp window is `open` using the `getMultiPopupOpenState` whereas previously it was checking if we are on the dashboard, 
if it's not open the dispatch to `store.controller.alarmLimits.draft` is not made.
The same thing applies for parametersRequest, 
for alarms there is another check which checks if there are any `unsaved` changes using the `getAlarmLimitsRequestUnsaved` selector, this was done to fix the following scenario:
make some unsaved change in alarms page -> open alarms modal from multipopup and close without doing any change (before the check it would already dispatch the local state to store.controller.alarmLimits.draft, which would discard the unsaved changes) made in alarms page
- so any increment is dispatched only if those checks are passed.
- fixes HFNC control for FIO2 and Flow showing measurements instead of showing setting change.


rohanpurohit:
Closed as we'll be using a different approach

PR #370, opened 2021-06-02
Implement alarm mute timeout countdown

### LINKED ISSUE #347: Alarm muting

rohanpurohit:
This PR includes:
- adds a countdown timer in backend simulator
- adds countdown to AlarmMuteService in firmware
- added a selector to detect connection loss to firmware
- disables the mute button if connection to the firmware is lost
- timer for backend connection lost alarm on the frontend


rohanpurohit:
@ethanjli notes from testing:
- There is a slight delay on the firmware side, although I feel the countdown itself is fine but it gets updated after every 10 secs? equal to the output schedule from the firmware? should we push alarm_mute similarly to sensor_measurements?

rohanpurohit:
@ethanjli this PR is ready for testing/code review

frontend/src/store/controller/selectors.ts
```diff
@@ -276,6 +285,18 @@ export const getBackendInitialized = createSelector(
   ): boolean => parametersRequest !== null && alarmLimitsRequest !== null && backendConnected,
 );
 
+// Firmware lost
+export const getVentilatingStatus = createSelector(
```
ethanjli: This needs a more accurate name, e.g. `getVentilatingStatusChanging`. `getVentilatingStatus` sounds like it would just return `Parameters.ventilating`.

frontend/src/store/controller/selectors.ts
```diff
@@ -263,7 +265,14 @@ export const getPopupEventLog = createSelector(getController, (states: Controlle
...
   return states.eventLog.nextLogEvents.elements.find((el: LogEvent) => el.id === maxId);
 });
+// Event log code selector
+const getLogEventCode = (logEventCode: number) =>
+  createSelector(getNextLogEvents, (events: LogEvent[]): LogEvent | undefined =>
+    events.find((el: LogEvent) => (el.code as number) === logEventCode),
+  );
 
+export const getBackendLogEvent = getLogEventCode(LogEventCode.backend_connection_down);
+export const getFirmwareLogEvent = getLogEventCode(LogEventCode.mcu_connection_down);
```
ethanjli: These need more accurate names, e.g. `getBackendDisconnected` and `getFirmwareDisconnected`. Ultimately the implementations of those selectors will be changed as specified in #371.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -18,7 +18,19 @@ static const float countdown_time = 120;
 
 class AlarmMuteService {
  public:
-  static void transform(AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request);
+  void transform(
+      uint32_t current_time, AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request);
```
ethanjli: The order of method arguments used elsewhere puts all request arguments before response arguments. Let's follow that same pattern here.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -18,7 +18,19 @@ static const float countdown_time = 120;
...
+  void start_countdown(AlarmMute &alarm_mute);
```
ethanjli: This method (or any refactored version of it) should be a private method, as it is just an implementation detail for how the `transform` method behaves rather than part of the public interface of `AlarmMuteService`.

frontend/src/modules/app/EventAlerts.tsx
```diff
@@ -203,13 +205,25 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
   const popupEventLog = useSelector(getPopupEventLog, shallowEqual);
   const activeLog = useSelector(getActiveLogEventIds, shallowEqual);
   const alarmMuteActive = useSelector(getAlarmMuteActive, shallowEqual);
-  const alarmMuteStatus = useSelector(getAlarmMuteStatus, shallowEqual);
+  const alarmMuteRemaining = useSelector(getAlarmMuteRemaining, shallowEqual);
```
ethanjli: For selectors which return `number` or `boolean`, do we still need `shallowEqual` in the `useSelector` function calls?

ethanjli:
Here are my results from testing the frontend with the firmware:
- When I press the "alarm mute" button, the countdown time decrements by 1 s, at an interval of 1 s. When I modify the backend to print all `AlarmMute` messages it sends to the frontend, I see that the backend always sends at least 3 messages (usually 4 messages) for each second of the `remaining` countdown; and currently the backend's schedule of messages to send to the frontend has `AlarmMute` being sent once every 230 ms. So actually the frequent & regular updates of the countdown time with our current state synchronization schedule is correct behavior, and whatever bug you had reported previously about seeing delays and large jumps in countdown time must have been from some other bug. We should figure out what was the root cause of that behavior and how it got fixed between your first commit on this PR and your most recent commit on this PR. This also means that #372 may be less urgent.
- If I unplug the MCU and plug it back in while the alarm mute is running, the alarms stay muted and the countdown time resets to 2 min (I believe we want the alarm muting to be cancelled in the frontend, backend, and firmware when a device is disconnected, it looks like that's part of #372 so it's for a future PR to address) and the alarm unmute button is disabled (this is a bug; I don't know if it's caused by this PR or if it existed previously)

ethanjli (commented):
Here are some comments from code review of all files besides `EventAlerts`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: Fixed, it's now refactored to do `remaining = countdown - current_time`

frontend/src/modules/app/EventAlerts.tsx
rohanpurohit: No we don't, changed

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: fixed.

rohanpurohit:
I have updated the alarm mute service in backend and firmware to follow `remaining = deadline - current_time` as we had discussed in the meeting, other than that I have removed the disabling mute button feature, selector seems to be buggy, I'll work on that in the future PR when we have protobuf message for connections or before that.

rohanpurohit:
> Here are my results from testing the frontend with the firmware:
> 
> * When I press the "alarm mute" button, the countdown time decrements by 1 s, at an interval of 1 s. When I modify the backend to print all `AlarmMute` messages it sends to the frontend, I see that the backend always sends at least 3 messages (usually 4 messages) for each second of the `remaining` countdown; and currently the backend's schedule of messages to send to the frontend has `AlarmMute` being sent once every 230 ms. So actually the frequent & regular updates of the countdown time with our current state synchronization schedule is correct behavior, and whatever bug you had reported previously about seeing delays and large jumps in countdown time must have been from some other bug. We should figure out what was the root cause of that behavior and how it got fixed between your first commit on this PR and your most recent commit on this PR. This also means that #372 may be less urgent.

Note: this was due to frontend lagging most probably because of chromium developer tools open with redux tab, closing the same resulted in expected normal countdown.
> * If I unplug the MCU and plug it back in while the alarm mute is running, the alarms stay muted and the countdown time resets to 2 min (I believe we want the alarm muting to be cancelled in the frontend, backend, and firmware when a device is disconnected, it looks like that's part of #372 so it's for a future PR to address) and the alarm unmute button is disabled (this is a bug; I don't know if it's caused by this PR or if it existed previously)

@ethanjli I cross-checked `eventsLog` > `nextLogEvents` > `elements` array in redux dev tools and found that `mcu_connection_down` which is `130`, persists even after the connection to the firmware is restored, which explains why the mute button is always disabled after `starting with application -> unplugging -> and plugging back in` as the selector, I added checks for the `mcu_connection_down` code in `nextlogevents`, if this is intended or unavoidable then makes sense to use protobuf messages later, if it's an easy fix then maybe we can implement the `disable mute button on connection lost` in this PR itself, and later just update the selector implementation for protobuf message for connections. thoughts?

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -36,6 +34,36 @@ def transform(# pylint: disable=no-self-use
...
+            response.remaining = 120
+
+        response.active = request.active
+
+    def continue_countdown(
+        self, response: mcu_pb.AlarmMute
+    ) -> None:
+        """countdown for two minutes."""
+        response.remaining = int((self.deadline - self.current_time) / 1000)
+        max(0, min(response.remaining, 120))
```
ethanjli: As far as I can tell, this line doesn't do anything: `max` returns a value but you don't do anything with it.

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -36,6 +34,36 @@ def transform(# pylint: disable=no-self-use
...
+            self.initial_time = time.time() * 1000
+            response.remaining = 120
+
+        response.active = request.active
+
+    def continue_countdown(
+        self, response: mcu_pb.AlarmMute
+    ) -> None:
+        """countdown for two minutes."""
+        response.remaining = int((self.deadline - self.current_time) / 1000)
```
ethanjli: I don't see `self.deadline` being changed anywhere - it appears that `self.deadline` is always 2 minutes. If I am understanding your code correctly, maybe it would be better to rename `self.deadline` to `self.max_mute_duration` or `self.mute_max_duration` and to rename `self.initial_time` to `self.mute_start_time`: a deadline is a specific timestamp (which should always be `self.mute_start_time` + 2 minutes), not a duration

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -36,6 +34,36 @@ def transform(# pylint: disable=no-self-use
...
+
+    def transform_mute(
+        self, request: mcu_pb.AlarmMuteRequest,
+        response: mcu_pb.AlarmMute
+    ) -> None:
+        """Implement alarm muting."""
+        if response.active:
+            self.continue_countdown(response)
+        else:
+            self.initial_time = time.time() * 1000
```
ethanjli: We should not use `time.time()` in this module or other simulator modules. Instead, `current_time` should be our source of "wall clock" time.

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -36,6 +34,36 @@ def transform(# pylint: disable=no-self-use
...
         )
+
+    def update_clock(
+        self, current_time: float,
+        response: mcu_pb.AlarmMute
+    ) -> None:
+        """Update the internal state for timing."""
+        if not response.active:
+            return
+        self.current_time = current_time * 1000 - self.initial_time
```
ethanjli: Since you are setting `self.initial_time` to be the current time whenever a response is inactive, it seems like `self.current_time` is a misleading name for this variable; instead, it is more like a `self.mute_duration`, essentially a stopwatch tracking how long the alarms have been muted since the mute was started.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -14,11 +14,23 @@
 
 namespace Pufferfish::Driver::BreathingCircuit {
 
-static const float countdown_time = 120;
+static const uint64_t countdown_time = 120;
```
ethanjli: This should be named `max_mute_duration` or `mute_max_duration` or something similar. Also, we should use units of milliseconds for this constant, to be consistent with all other timestamps in the firmware.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -14,11 +14,23 @@
...
  public:
-  static void transform(AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request);
+  void transform(
+      uint32_t current_time, const AlarmMuteRequest &alarm_mute_request, AlarmMute &alarm_mute);
+
+ private:
+  void continue_countdown(AlarmMute &alarm_mute);
+  static const uint32_t sensor_update_interval = 1000;  // ms
+
+  uint32_t current_time_ = 0;   // ms
```
ethanjli: It looks like this should be renamed to `mute_duration_` or something similar?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -14,11 +14,23 @@
...
+  uint32_t previous_time_ = 0;  // ms
+  uint32_t initial_time_ = 0;   // ms
```
ethanjli: It looks like this should be renamed to `mute_start_` or something similar?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -14,11 +14,23 @@
...
 
 class AlarmMuteService {
  public:
-  static void transform(AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request);
+  void transform(
+      uint32_t current_time, const AlarmMuteRequest &alarm_mute_request, AlarmMute &alarm_mute);
+
+ private:
+  void continue_countdown(AlarmMute &alarm_mute);
+  static const uint32_t sensor_update_interval = 1000;  // ms
```
ethanjli: Let's try to remove `sensor_update_interval`, `update_needed`, and `previous_time_` if possible.

ethanjli (commented):
Here's my code review of the Alarm Mute Service implementations.

ethanjli:
Regarding the `mcu_connection_down` issue: I think it's reasonable to leave your `getFirmwareDisconnected` selector in place with a TODO comment about when it returns the wrong values (i.e. after firmware has been reconnected) and what the fix is (i.e. having a separate protobuf message), and then we can merge in current (incorrect) implementation as a placeholder in this PR.

Here are my findings from more thorough testing:
- When I try to run the frontend and application backend with the firmware, the frontend encounters some sort of fatal error when `mcu_pb.ts` attempts to serialize `message.remaining` as a uint64. I suspect this is because the frontend is leaving `message.remaining == undefined` upon startup (which is what I see in the Redux console). I think we'll need to make the application backend track `AlarmMuteRequest` as a message to save and load from the filesystem, by adding it to `protocols.backend.states.FILE_INPUT_TYPES`. We'll also need a way to initialize `AlarmMuteRequest` if the file isn't already on the filesystem; this is a bit complicated, so I've created PR #377 to fix this for you - I tried merging that branch into my local copy of this PR and found it solved the problem, so see if it also works for you.
- After fixing the above issue on my local copy of the repo, I tested the frontend and backend with the firmware, and I found that when I press the mute button the countdown always stays stuck at 02:00.
- When I start ventilating on the simulator backend and have active alarms, then mute the alarms, then disconnect the backend, the alarm mute remains, its countdown timer resets to 02:00, and then after it reaches 00:00 the alarm mute still remains but the countdown timer "circles around" back to 59:59 and starts counting down from there. If I then unmute and mute again, then the countdown time display alternates between counting down from 02:00 and counting down from 59:59. I suspect this may be related to the use of `setTimeout` in a loop which never gets cancelled for some reason. This is undesirable behavior; if it's a simple fix, let's include that in this PR.
- When I start ventilating on the simulator backend and have active alarms, then mute the alarms, and an alarm bounces between being active and inactive, once the alarm goes inactive the alarm mute countdown should reset, but it doesn't reset. Right now when the alarm goes inactive there's no way to cancel the mute, but then once the alarm becomes active again it becomes clear that the countdown has not stopped and is still running on the previous deadline. I forget if we discussed what behavior we want to have when an alarm mute is active and then all alarms become inactive. Maybe we decided that the alarm mute should remain active even when all alarms are deactivated? If so, then I think the countdown and "stop muting alarms" button should always be visible and active whenever an alarm mute is active. This may need to be handled in a separate PR.

rohanpurohit:
> Regarding the `mcu_connection_down` issue: I think it's reasonable to leave your `getFirmwareDisconnected` selector in place with a TODO comment about when it returns the wrong values (i.e. after firmware has been reconnected) and what the fix is (i.e. having a separate protobuf message), and then we can merge in current (incorrect) implementation as a placeholder in this PR.
> 
> Here are my findings from more thorough testing:
> 
> * When I try to run the frontend and application backend with the firmware, the frontend encounters some sort of fatal error when `mcu_pb.ts` attempts to serialize `message.remaining` as a uint64. I suspect this is because the frontend is leaving `message.remaining == undefined` upon startup (which is what I see in the Redux console). I think we'll need to make the application backend track `AlarmMuteRequest` as a message to save and load from the filesystem, by adding it to `protocols.backend.states.FILE_INPUT_TYPES`. We'll also need a way to initialize `AlarmMuteRequest` if the file isn't already on the filesystem; this is a bit complicated, so I've created PR #377 to fix this for you - I tried merging that branch into my local copy of this PR and found it solved the problem, so see if it also works for you.

Thanks for making that change! it works fine for me too
> * After fixing the above issue on my local copy of the repo, I tested the frontend and backend with the firmware, and I found that when I press the mute button the countdown always stays stuck at 02:00.

Yes! seems like after I changed remaining from `float` to `uint64_t`, it has stopped working. I'll fix the service and the rest of the issues. thanks

rohanpurohit:
> I forget if we discussed what behavior we want to have when an alarm mute is active and then all alarms become inactive. Maybe we decided that the alarm mute should remain active even when all alarms are deactivated? 

I don't recall discussing this particular scenario, mainly we had discussed if alarm mute is active, any new alarm that is triggered should also be muted until the countdown is over, which wasn't the case earlier and was fixed in #348

> If so, then I think the countdown and "stop muting alarms" button should always be visible and active whenever an alarm mute is active. This may need to be handled in a separate PR.

I agree, otherwise, the countdown just runs down silently, maybe we could display something like "No Alarms Active" along with the countdown

backend/ventserver/simulation/alarm_mute.py
rohanpurohit: Oh, yesss my bad! haha

backend/ventserver/simulation/alarm_mute.py
rohanpurohit: oh okay got it! using `mute_max_duration`

backend/ventserver/simulation/alarm_mute.py
rohanpurohit: ok got it! fixed.

backend/ventserver/simulation/alarm_mute.py
rohanpurohit: yeah! using `mute_start_time`

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: using `mute_max_duration `

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: using `mute_start_time_`

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: using `mute_duration_`

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: Yep removed!

ethanjli:
What I see in testing is that, even after debouncing, a value might temporarily return to within the alarm limits range for several seconds (which is enough to make it through the debouncing filter) and then go back out of range. I don't think we would want this to cancel the alarm mute early, it could be annoying. So let's keep the mute going even after all alarms are inactivated. This way we'll have a very simple description for the desired software behavior for alarm mute initiation and cancellation.

A temporary alarm mute period is cancelled and reset exactly when:
1. The user presses the unmute button from the touchscreen display or the membrane button (manual cancellation),
2. The 2-minute deadline on the temporary mute period is reached (automatic cancellation from timeout), or
3. Something has gone very wrong with the system and needs immediate attention: a piece of software has crashed or frozen, or a connection has been lost (automatic cancellation from severe system fault)

The user can press the mute button whenever both of the following requirements are met:
1. Alarms are not muted (even if there are not currently any active alarms)
2. All software and connections are running normally

Whenever alarms are muted, the user can always see the countdown to the deadline for automatic cancellation from timeout, and the user can always press the unmute button.

rohanpurohit:
@ethanjli I have resolved the issues you faced in the last test! along with that, I have added changes for the `alert` to stay active without any alarms, the part about the mute button being accessible can be done in a future PR, where we have better selectors for all connections.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
```diff
@@ -9,16 +9,26 @@
...
 #pragma once
 
-#include "Pufferfish/Application/LogEvents.h"
-#include "Pufferfish/Util/OrderedMap.h"
+#include "Pufferfish/Application/States.h"
 
 namespace Pufferfish::Driver::BreathingCircuit {
 
-static const float countdown_time = 120;
+static const uint64_t countdown_time = 120;  // s
```
ethanjli: It looks like this constant is no longer used anywhere, so we can remove it.

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -10,20 +11,18 @@
...
 @attr.s
 class AlarmMuteService:
     """Implement Alarm Mute Service"""
 
-    def transform(# pylint: disable=no-self-use
-            self, store: Mapping[
+    mute_duration: float = attr.ib(default=0)  # ms after mute_start_time
+    mute_start_time: float = \
+        attr.ib(default=time.time() * 1000)  # ms, Unix time
+    mute_max_duration: int = 120000 # ms
```
ethanjli: The convention in the firmware is that all timestamps should be `uint32_t`s in ms, while in the backend all timestamps should be `float`s in s; this is because of the types and units for the firmware's `HAL::Time.millis()`-based clock and the backend's `time.time()`-based clock. It looks like maybe the backend's breathing circuit simulator doesn't follow this convention yet; but if you can follow this convention for `AlarmMuteService`, that'd be great.

backend/ventserver/simulation/alarm_mute.py
```diff
@@ -10,20 +11,18 @@
...
-def transform(
-    request: mcu_pb.AlarmMuteRequest,
-    response: mcu_pb.AlarmMute
-) -> None:
-    """Mute the current active alarm"""
-
-    response.active = request.active
-
 @attr.s
 class AlarmMuteService:
```
ethanjli: It looks like the naming convention I've set for service classes is `Service`/`Services`, because services are accessed through the python module name (it should be `alarm_mute_service = alarm_mute.Service()`) - see https://github.com/pez-globo/pufferfish-software/blob/1f7e1926a3dd0b4656c5efcc64e6fad36cf86b52/backend/ventserver/simulator.py#L123. Let's rename this class from `AlarmMuteService` to `Service`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
```diff
@@ -11,9 +11,28 @@
...
+}
+
 void AlarmMuteService::transform(
-    AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request) {
+    uint32_t current_time, const AlarmMuteRequest &alarm_mute_request, AlarmMute &alarm_mute) {
   alarm_mute.active = alarm_mute_request.active;
+  input_clock(current_time);
+
+  if (alarm_mute.active) {
+    continue_countdown(alarm_mute);
```
ethanjli: Since `continue_countdown` only has one line in it and is only used here, it might be better to merge the contents of `continue_countdown` back into here and delete the `continue_countdown` method.

ethanjli (commented):
Here's my review of the code - only minor changes for readability of the code.

Observations from testing:
- When the simulator backend is disconnected while alarms are muted, the countdown timer resets to 02:00 and starts counting down from there, instead of continuing the countdown. The alarm mute cancels once it reaches 00:00; however, when I press the mute button again afterwards, the countdown timer still alternates between counting down from 02:00 and counting down from 59:59; and after that countdown finishes, pressing the mute button again causes the countdown timer to alternate between three countdown times. Maybe you're missing a `clearTimeout(timer)` call somewhere else? Also, if alarms are muted and the backend is disconnected and reconnected, the countdown timer resets to counting down from 02:00 after the backend is reconnected. We probably don't need to fix these buggy behaviors in this PR, since #372 will cancel the alarm mute when the backend is disconnected. However, in the PR which addresses #372, we need to test whether these behaviors still occur.
- Everything seems to work fine with the firmware. I haven't tested any behaviors when the firmware is disconnected - I don't think this PR has code related to that condition, so it will be tested in the PR for #372.

For records-keeping:

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.h
rohanpurohit: i'm using it for state_initializer https://github.com/pez-globo/pufferfish-software/pull/370/files#diff-ad9e425821d521acd9105056463dc5301d1d059e25c620755f2b15b5fa02568aR39 here, should i replace it with `mute_max_duration`, which is in `ms`

ethanjli: Ah, I see. Yes, let's replace it with `mute_max_duration` - perhaps you can make that a `static const`. If our `AlarmMute` protobuf message definition uses units of seconds and that's making the firmware code more complicated than it needs to be, maybe we should change the `AlarmMute` protobuf message to use units of milliseconds, to be consistent with the units of time everywhere else in `mcu_pb.proto`? Let's discuss in today's meeting.

rohanpurohit:
> * When the simulator backend is disconnected while alarms are muted, the countdown timer resets to 02:00 and starts counting down from there, instead of continuing the countdown. The alarm mute cancels once it reaches 00:00; however, when I press the mute button again afterwards, the countdown timer still alternates between counting down from 02:00 and counting down from 59:59; and after that countdown finishes, pressing the mute button again causes the countdown timer to alternate between three countdown times. Maybe you're missing a `clearTimeout(timer)` call somewhere else? Also, if alarms are muted and the backend is disconnected and reconnected, the countdown timer resets to counting down from 02:00 after the backend is reconnected. We probably don't need to fix these buggy behaviors in this PR, since #372 will cancel the alarm mute when the backend is disconnected. However, in the PR which addresses #372, we need to test whether these behaviors still occur.

I just pushed a change to continue countdown, and clearTimeouts where needed, although I'm not entirely sure if it has fixed all the issues stated here.

> * Everything seems to work fine with the firmware. I haven't tested any behaviors when the firmware is disconnected - I don't think this PR has code related to that condition, so it will be tested in the PR for #372.

Yes we'll just have to run the same countdown for `firmware disconnection` once we have a reliable selector for it, that's all. so yes it doesn't work currently and only runs when backend is disconnected.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
```diff
@@ -11,9 +11,24 @@
...
+  if (mute_start_time_ == 0) {
+    mute_start_time_ = current_time;
+  }
+  mute_duration_ = current_time - mute_start_time_;
+}
+
 void AlarmMuteService::transform(
-    AlarmMute &alarm_mute, const AlarmMuteRequest &alarm_mute_request) {
+    uint32_t current_time, const AlarmMuteRequest &alarm_mute_request, AlarmMute &alarm_mute) {
   alarm_mute.active = alarm_mute_request.active;
```
ethanjli, line 25: I just remembered...`AlarmMuteService` should automatically set `alarm_mute.active` to `false` once `duration >= mute_max_duration` is satisfied; even if `alarm_mute_request.active == true`, `AlarmMuteService` should only set `alarm_mute.active = alarm_mute_request.active` after a timeout, and/or only when `alarm_mute_request` has a new starting time which is past the ending time of the previous alarm mute. If this is too complicated for this PR, or if this depends on #372, let's fix this in a future PR.

Also, since `remaining` is a `uint32_t` and it's calculated as the difference between two `uint32_t`s (`self.mute_max_duration` and `self.mute_duration`), `remaining` will never be negative. You'll need to first cast the two durations into `int64_t`s or similar, then clamp the difference (rescaled to ms) to be between `0` and `120`, and then write the result to `response.remaining`.

ethanjli (commented):
One more thing I noticed, now that I see everything `AlarmMuteService::transform`. Everything looks fine in testing.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
```diff
@@ -24,19 +26,22 @@ void AlarmMuteService::transform(
   input_clock(current_time);
 
   if (alarm_mute.active) {
-    alarm_mute.remaining = (mute_max_duration - mute_duration_) / clock_scale;
+    int remaining =
```
ethanjli: In general for the firmware, we should use `int32_t` or `int64_t` rather than `int`, just to make the width of the type explicit. It should be fine to declare `remaining` as `int64_t`, to keep the same width as the right-hand side of the assignment.

ethanjli (approved):
These changes look good to me. Once you fix this one tiny thing and confirm for yourself that it compiles fine, passes clang-tidy, and works in light testing, and resolve merge conflicts from develop, I think this PR is ready to merge.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/AlarmMuteService.cpp
rohanpurohit: I think I tried int64_t first but I guess util::clamp doesn't accept int64_t as a valid type?

ethanjli: Here's the function declaration: https://github.com/pez-globo/pufferfish-software/blob/53f83d824b8db7964133b40d88b00ca4372238b9/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Ranges.h#L52

You may need to call it as `Util::clamp<int64_t>(...)`. I guess the template argument type deduction got confused if you gave it an `int64_t` and two integer literals `0` and `120` (which are probably `int32_t`s but I'm not sure); if you provided `mute_max_duration / clock_scale` (which evaluates to a `uint32_t`) instead of `120` and it wasn't working, the template argument type deduction would definitely fail because it expects all arguments to have the same type. In that case you'll have to explicitly specify the template argument type(s) for a templated function.

rohanpurohit: Ya, I did pass integer literals, yep specifying the template type works, I'll test it and maybe you could check if it's fine now?

rohanpurohit: @ethanjli seems fine on light testing. will merge once you resolve this conversation 👍🏽

PR #381, opened 2021-06-13
Open modal popup if sidebar is accessed from alarms screen

### LINKED ISSUE #350: Changing page from the sidebar on unsaved alarm limits changes should display a confirmation dialogue
When we are on the set alarms page and the alarm limits are changed but not submitted, pressing on the sidebar to change the page for instance (settings page or home) should display a confirmation dialogue.

rohanpurohit:
This PR adds functionality to open confirmation dialog on clicking the `pufferfish icon` that opens the sidebar if there are `unsaved` changes on the alarm limits screen.

### DIFF
--- a/frontend/src/modules/app/layouts/SidebarSlideRoute.tsx
+++ b/frontend/src/modules/app/layouts/SidebarSlideRoute.tsx
@@ -125,6 +125,7 @@ const FullWidthToolBar = (): JSX.Element => {
   const handleDiscardConfirm = () => {
     setAlarmLimitsRequestDraft(alarmLimitsRequest);
     setOpenModal(false);
+    setToggle(true);
   };
 
   /**
--- a/frontend/src/modules/app/layouts/SidebarSlideRoute.tsx
+++ b/frontend/src/modules/app/layouts/SidebarSlideRoute.tsx
@@ -3,16 +3,27 @@
  *
  */
 import React, { PropsWithChildren, useEffect } from 'react';
-import { Route, RouteProps } from 'react-router-dom';
+import { Route, RouteProps, useLocation } from 'react-router-dom';
 import { makeStyles, Theme } from '@material-ui/core/styles';
-import { Button, Drawer, Grid } from '@material-ui/core';
-import { useSelector } from 'react-redux';
+import { Button, Drawer, Grid, Typography } from '@material-ui/core';
+import { useDispatch, useSelector } from 'react-redux';
 import ToolBar from '../ToolBar';
 import UserActivity from '../UserActivity';
-import { SCREENSAVER_ROUTE } from '../../navigation/constants';
+import { ALARMS_ROUTE, SCREENSAVER_ROUTE } from '../../navigation/constants';
 import SidebarClickable from '../SidebarClickable';
 import OverlayScreen from '../OverlayScreen';
 import { getAlarmNotifyStatus } from '../../../store/app/selectors';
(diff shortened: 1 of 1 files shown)


rohanpurohit:
Reasons for this approach:
- settings button on sidebar is dynamically added by routes in [Navigation](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/navigation/Navigation.tsx#L87), and to accommodate changes for confirmation dialog led to a sort of static approach instead of the dynamic nature currently, also breaks quite a few things, that was something i tried sometime back, i could try again!
- if we also need a confirmation dialog on pressing the `home` button on the `sidebar` then that would mean duplicating the same code in `SidebarClickable`.

Thus, the current approach felt more feasible, we could maybe discuss whether on `confirmation` from modal popup, would we still want to toggle the sidebar? so that the user is not forced to press again.

ethanjli:
Right now the user flow is unintuitive because pressing the pufferfish icon when there are unsaved changes doesn't actually open the sidebar. I suspect that if we toggle the sidebar upon confirming to discard unsaved changes, usability will be better. However, I'm not entirely sure whether usability will be acceptable with that change. Let's merge in this PR after making the sidebar be shown after pressing the Confirm button to discard unsaved changes, and then if we get feedback about the usability later, then we can figure out what it would take to be able to show the confirmation dialog only when pressing buttons in the sidebar.

My intuition for that deeper solution is that we'd need to find a way to add a custom onClick handler callback function to each sidebar button, and a custom callback for determining whether to proceed to the route represented by the button; then the onClick handler would set something to show the modal, and the confirm button of the modal would set whether to proceed to the route. I haven't looked at the architecture of the routing or sidebar systems enough to know whether this is feasible or easy or even possible, but I suspect it would require a deeper reworking of those systems - which should only be done after we make sure all of Sudhir's replies to our questions about the code comments have been integrated into the frontend.

One interesting consequence of the approach taken by this PR is that it's not possible to lock the screen from the sidebar while there are unsaved changes to alarm limits - in that condition, the screen can only be locked with the hardware button. This is probably fine, however.

frontend/src/modules/app/layouts/SidebarSlideRoute.tsx
```diff
@@ -112,6 +160,38 @@ const FullWidthToolBar = (): JSX.Element => {
           </Drawer>
         </React.Fragment>
       </Grid>
+      <ModalPopup
```
ethanjli, line 164: Are the contents of this `ModalPopup` component copy-pasted from somewhere else? If so, we should decompose it into a separate component and/or a separate container (in the sense of a component bound to selectors using react-redux's `connect`, like what I did with the waveform plots). If we have the same few modal popups copy-pasted across the frontend codebase, maybe we should make a "frontend/src/modules/Modals" folder which contains reusable modals, e.g. for submitting/discarding unsaved changes to alarm limits, and for submitting/discarding unsaved changes to parameters.

rohanpurohit: Yes `ModalPopup` is copy-pasted multiple times, I will try to add a re-usable component in this PR.

rohanpurohit: we will do this in a future PR

ethanjli:
This works fine in light testing. For records-keeping:

rohanpurohit:

ethanjli (approved):

PR #383, opened 2021-06-14
Prevent the rotary encoder from setting min limit above max limit in the frontend

### LINKED ISSUE #386: When frontend refreshes from the quickstart screen, it fails to reestablish its connection
When the system is in ventilating mode and the frontend closes and is reopened on the landing page (which is where systemd should reopen the frontend if the frontend gets killed from freezing or dies from crashing), the frontend will automatically take the user to the ventilating dashboard. This is desirable.

When the frontend closes and is reopened (or is merely refreshed, like what `yarn start` will do when a file in the frontend is changed) on the quickstart screen, regardless of whether the system is in ventilating mode or standby mode, the frontend will fail to connect to the backend, and the backend does not receive any messages from the frontend (so it wants to kill the frontend). In the chrome devtools console, I see a protobuf error message that the `remaining` field of `AlarmMuteRequest` is undefined and thus cannot be decoded, and this unhandled exception seems to break the frontend's saga for talking to the backend. I saw this problem previously in https://github.com/pez-globo/pufferfish-software/pull/370#issuecomment-857274561 and I thought I had fixed it in #377, but apparently that was only a partial fix. When I check the redux console, I see that `store.controller.alarmMute.request` only has `active` defined, while `remaining` is not defined. I'm not sure why this issue only appears when the frontend is opened from the quickstart screen, rather than from the landing page.

This behavior can be observed on commit 2953546, and I don't remember seeing this behavior in the past; however, I have not tried running previous commits to see where the regression came from (though I strongly suspect it's from #370). This issue doesn't break the frontend for deployment, but to me it suggests some deeper problem with the frontend's and/or backend's initialization behavior for `AlarmMuteRequest`.

rohanpurohit:
This PR adds #374 fix for rotary encoder

### DIFF
--- a/frontend/src/modules/controllers/ValueClicker.tsx
+++ b/frontend/src/modules/controllers/ValueClicker.tsx
@@ -96,7 +96,8 @@ export const ValueClicker = ({
     setDisableDecrement(change <= min);
     if (change >= max) {
       return onClick(max);
-    } else if (change <= min) {
+    }
+    if (change <= min) {
       return onClick(min);
     }
     return onClick(change);
--- a/frontend/src/modules/controllers/ValueClicker.tsx
+++ b/frontend/src/modules/controllers/ValueClicker.tsx
@@ -53,8 +53,6 @@ interface Props {
   direction?: Direction;
   step?: number;
   referenceKey: string;
-  disableMax?: boolean;
-  disableMin?: boolean;
 }
 enum Type {
   INCREMENT,
@@ -77,8 +75,6 @@ export const ValueClicker = ({
   direction = 'column',
   step = 1,
   referenceKey,
-  disableMax = false,
-  disableMin = false,
 }: Props): JSX.Element => {
   const classes = useStyles();
   const [disabledInterval] = useState(100);
@@ -102,18 +98,17 @@ export const ValueClicker = ({
       return onClick(max);
     } else if (change <= min) {
       return onClick(min);
-    } else {
-      return onClick(change);
     }
+    return onClick(change);
   };
 
   /**
    * Validating & disabling button when reached min/max value
    */
   useEffect(() => {
-    setDisableIncrement(value >= max || disableMin);
-    setDisableDecrement(value <= min || disableMax);
-  }, [min, max, value, disableMin, disableMax]);
+    setDisableIncrement(value >= max);
+    setDisableDecrement(value <= min);
+  }, [min, max, value]);
 
   /**
    * Click handler for Increment button
--- a/frontend/src/modules/controllers/ValueClicker.tsx
+++ b/frontend/src/modules/controllers/ValueClicker.tsx
@@ -98,7 +98,13 @@ export const ValueClicker = ({
     const change = value + step;
     setDisableIncrement(change >= max);
     setDisableDecrement(change <= min);
-    return onClick(change);
+    if (change >= max) {
+      return onClick(max);
+    } else if (change <= min) {
+      return onClick(min);
+    } else {
+      return onClick(change);
+    }
   };
 
   /**
--- a/frontend/src/modules/alarms/AlarmsPage.tsx
+++ b/frontend/src/modules/alarms/AlarmsPage.tsx
@@ -218,6 +218,8 @@ const Alarm = ({
    */
   const [disableDecrement, setDisableDecrement] = useState(false);
   const [disableIncrement, setDisableIncrement] = useState(false);
+  const [rotaryMin, setRotaryMin] = React.useState(min);
+  const [rotaryMax, setRotaryMax] = React.useState(max);
 
   /**
    * This is a changeListener that sets disableDecrement, disableIncrement on change in RangeValues which
@@ -226,7 +228,9 @@ const Alarm = ({
   useEffect(() => {
     setDisableDecrement(rangeValues[1] <= rangeValues[0]);
     setDisableIncrement(rangeValues[0] >= rangeValues[1]);
-  }, [rangeValues]);
+    if (disableIncrement) setRotaryMin(rangeValues[0]);
+    if (disableDecrement) setRotaryMax(rangeValues[1]);
+  }, [rangeValues, disableIncrement, disableDecrement]);
 
   /**
    * Calls on initalization of the component
@@ -276,7 +280,7 @@ const Alarm = ({
                 value={rangeValues[0]}
                 onClick={(value: number) => onClick(value, SliderType.LOWER)}
                 min={min}
-                max={max}
+                max={disableIncrement ? rotaryMax : max}
                 disableMin={disableIncrement}
                 direction="column"
               />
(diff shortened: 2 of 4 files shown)


rohanpurohit:
My observations from testing:
- normal rotary encoder movement everything seems fine.
- if the rotary encoder is moved too fast, the value goes beyond the `min`, `max` but comes back to the `disabled` limit, I'm not sure if we can fix this without altering the `RotaryEncoderController`

ethanjli:
I'm able to reproduce your observations. However, I also was able to quickly turn the rotary encoder to set min > max, and the min didn't get automatically moved back down to be at max again. Only after I turned the rotary encoder again did the min get moved back down to the max value.

From skimming over the code, I see code adjusting the `min`/`max` params passed to `ValueClicker` based on a `disableIncrement`/`disableDecrement` local state. I think the brittle behavior described above is occurring because of this implementation. Do you think we can set the `min` and `max` params for `ValueClicker` without relying on `disableIncrement`/`disableDecrement`?

rohanpurohit:
@ethanjli everything seems to be working fine on my end now with the latest changes. can you confirm the same?

ethanjli:
It works. I did find a more subtle bug in testing: when adjusting the upper alarm limit, if I rotate the rotary encoder and press the decrement button at nearly the same time, I am able to force a race condition which allows the upper limit to be set to below the lower limit. I'm not always able to make this happen because it depends on the relative timing of events (when the rotary encoder message arrives at the frontend vs. when the decrement button is pressed vs. when the decrement button becomes disabled - usually I have to turn the dial and then after a very short moment then press the decrement button), but I'm able to do it often enough. I think we do need to fix the logic related to handling of increment/decrement buttons so that the min and max ValueClicker limits are always enforced on them independent of whether the buttons are disabled; we could do that in this PR (since it's so small right now) or a future PR (if that will be more work than just a small hotfix) - the proper solution for handling of increment & decrement buttons will be such that even if we never disable the buttons, pressing them when lower = upper will not allow lower > upper. Do you want to fix that in this PR or separately?

rohanpurohit:
> It works. I did find a more subtle bug in testing: when adjusting the upper alarm limit, if I rotate the rotary encoder and press the decrement button at nearly the same time, I am able to force a race condition which allows the upper limit to be set to below the lower limit. I'm not always able to make this happen because it depends on the relative timing of events (when the rotary encoder message arrives at the frontend vs. when the decrement button is pressed vs. when the decrement button becomes disabled - usually I have to turn the dial and then after a very short moment then press the decrement button), but I'm able to do it often enough. I think we do need to fix the logic related to handling of increment/decrement buttons so that the min and max ValueClicker limits are always enforced on them independent of whether the buttons are disabled; we could do that in this PR (since it's so small right now) or a future PR (if that will be more work than just a small hotfix) - the proper solution for handling of increment & decrement buttons will be such that even if we never disable the buttons, pressing them when lower = upper will not allow lower > upper. Do you want to fix that in this PR or separately?

I was able to reproduce the regression,
I think it's best if we fix it in this PR, as you said it's too small currently, and finish it by the end of 0.12 milestone

>  the proper solution for handling of increment & decrement buttons will be such that even if we never disable the buttons, pressing them when lower = upper will not allow lower > upper

I will follow this approach.

frontend/src/modules/controllers/ValueClicker.tsx
```diff
@@ -98,7 +98,13 @@ export const ValueClicker = ({
...
     setDisableIncrement(change >= max);
     setDisableDecrement(change <= min);
-    return onClick(change);
+    if (change >= max) {
+      return onClick(max);
+    } else if (change <= min) {
+      return onClick(min);
+    } else {
+      return onClick(change);
+    }
```
ethanjli:
```suggestion
    if (change >= max) {
      return onClick(max);
    }
    if (change <= min) {
      return onClick(min);
    }
    return onClick(change);
```
To address a warning raised in the Github Actions checks (not sure why the warning didn't cause the frontend to fail the eslint check action)

ethanjli (commented):
I tested this two ways:

1. On the Raspberry Pi, trying to reproduce the previous behavior. I have not been able to reproduce the behavior.
2. On my computer commenting out the `disableIncrement` and `disableDecrement` code and confirming that even when `lower == upper`, when I press the increment and decrement buttons I am never able to reach `lower > upper`. This works.

I also see that in `ValueClicker`, `AlarmsPage`, and `AlarmModal`, the `disableIncrement` and `disableDecrement` variables are now only used for setting the enable/disable state of those buttons, and not for anything else. So I think the issue is fixed.

From looking at those three files, I see that each of them makes its own local `useState` for `disableIncrement` and `disableDecrement`, and each has its own `useEffect` for setting those states; and both `AlarmsPage` and `AlarmModal` have their own sections in their returned components making grid containers of `ValueClicker` pairs for increment/decrement and slider and label display. Is there any way we can reduce the duplication of states and/or code, either in this PR or a future PR?

For records-keeping:

rohanpurohit:
> I tested this two ways:
> 
> 1. On the Raspberry Pi, trying to reproduce the previous behavior. I have not been able to reproduce the behavior.
> 2. On my computer commenting out the `disableIncrement` and `disableDecrement` code and confirming that even when `lower == upper`, when I press the increment and decrement buttons I am never able to reach `lower > upper`. This works.
> 
> I also see that in `ValueClicker`, `AlarmsPage`, and `AlarmModal`, the `disableIncrement` and `disableDecrement` variables are now only used for setting the enable/disable state of those buttons, and not for anything else. So I think the issue is fixed.
> 
> From looking at those three files, I see that each of them makes its own local `useState` for `disableIncrement` and `disableDecrement`, and each has its own `useEffect` for setting those states; and both `AlarmsPage` and `AlarmModal` have their own sections in their returned components making grid containers of `ValueClicker` pairs for increment/decrement and slider and label display. Is there any way we can reduce the duplication of states and/or code, either in this PR or a future PR?

Yeah since we are now passing `min` and `max` dynamically `disableIncrement` and `disableDecrement` are redundant, I will clean it up in this PR.

rohanpurohit:
@ethanjli since I added the fix for #386 in this PR, that would need a confirmation from you, and light testing after cleanup.
seems to be working fine on my end.

frontend/src/modules/controllers/ValueClicker.tsx
rohanpurohit: I guess it warns even if I have else if's with the same return statements.

ethanjli (approved):
I tested on the RPi and confirmed that the behavior described in #386 no longer occurs, and I also did light testing with the alarm limits adjustment to confirm that the increment/decrement buttons still seem to be disabled when appropriate. Can you also add a code comment above `commitRequest<AlarmMuteRequest>(MessageType.AlarmMuteRequest, {` explaining why we need to include the `remaining: 120,`?

Issue #372, opened 2021-06-04
Need alarm mute cancellations and session tracking to handle disconnections


ethanjli:
To have robust behavior of the alarm mute functionality, we need to appropriately handle what happens when the firmware-backend and/or backend-frontend connections are lost (either due to physical issues or due to a software crash/restart). Specifically, we would like the following behaviors:

Sequence of events:
1. Alarms are muted for 2 minutes
2. Firmware is disconnected or it restarts (which may or may not cause a disconnection event in the backend)
3. Any alarm mute should be cancelled early (firmware and backend and frontend all set AlarmMute.active to false) in the firmware and frontend and backend. The cancelled alarm mute should not be spuriously reactivated under any circumstances. For example, if the firmware restarts and it receives a stale AlarmMuteRequest with the `active` field set to `true`, it should recognize that the request is stale and ignore it instead of initiating a new alarm mute for 120 seconds.
    
Sequence of events:
1. Alarms are muted for 2 minutes
2. Backend is disconnected
3. Any alarm mute should be cancelled early (firmware and backend and frontend all set AlarmMute.active to false) in the firmware and frontend and backend.

These requested behaviors are similar to the behavior between the firmware and backend for `NextLogEvents` and `ExpectedLogEvent`. We will need to add a `session_id` field to the request/response protobuf messages, so that the firmware can tell when an inbound AlarmMuteRequest belongs to a stale session:

AlarmMute:

- bool active
- uint64 remaining (firmware always counts down from 120 sec, regardless of remaining field in AlarmMuteRequest)
- uint32 session_id (used by firmware to determine whether a request is stale; initialized from random number on startup; firmware increments this by one when the membrane button is pressed)

AlarmMuteRequest:
- bool active (frontend can use this to initiate a new alarm mute session, or to cancel the current mute session)
- uint64 remaining (ignored, only provided for symmetry with AlarmMute; maybe we can remove this field entirely?)
- uint32 session_id (when frontend wants to start a new session, it always uses the next consecutive number from the AlarmMute response's session_id)

Should the `remaining` field be uint32 or uint64? Should it be in units of ms or s? This needs to be determined. It may save us a few bytes to report it in seconds, but all other timestamps in the message protocols are in units of milliseconds. If we use units of seconds, we should rename the field to `remaining_sec` to make the unit difference very explicit. But it may be better just to use the same representation for timestamps & durations as every other protobuf message.

[...]

PR #391, opened 2021-06-17
Improve frontend code comments and refactor modules/app

### LINKED ISSUE #331: alarm limits from the set_alarms screen are set just by changing the values
Alarm limits are changed on incrementing/decrementing alarm_limits for spo2 and/or HR from the set alarm limits screen, without pressing apply changes

1. Start Ventilation
2. go to set alarms screen
3. increment, decrement spo2 or hr alarms and check event log

**Expected behavior**
alarm limits should be changed only on pressing "apply changes"

**Additional context**
backend running ventserver.simulator , maybe its just related to the frontend?
revision: f50a5e84db4307225fe73e02748b4f5b07c6fbbe

maybe a regression from 4ee0ee48d50126cbaefa3867f5128e5861664768

### LINKED ISSUE #361: Frontend Start/Pause Ventilation button behavior when MCU is lost
When MCU is lost, allow the frontend to tell the backend that it should tell the MCU to Start/Pause Ventilation once the MCU is back up. In other words, the Start/Pause Ventilation button shouldn't be automatically changed to disabled. However, whenever parameters.current.ventilating != parameters.request.ventilating, the Start/Pause ventilation button should say "Connecting..."

This issue was split off from https://github.com/pez-globo/pufferfish-software/pull/344#issuecomment-841357911

### LINKED ISSUE #390: Improve code comments and refactor/improve code in modules/app
This issue is split off from [#359 comment](https://github.com/pez-globo/pufferfish-software/pull/359#issuecomment)

In general:

- Cross-checking all code comments with your replies to the questions we added as Github comments above, to make sure they include all of the clarifying information you provided for anything which you haven't already included in the code comments. This would also include any resolutions to the unresolved comments listed above.

modules/app:
- staticStart and having a different react component for the Start button in the landing page vs. the "Start/Stop Ventilation" button in the other pages, to simplify the code: #359 (comment)
- Removing isVentilatorOn, which is dead code: #359 (comment) , #359 (comment) . This may also remove the need for a landingLabel useState? That was discussed in #359 (comment)
- refactoring the way values get committed to ...Requests when the Start Ventilation button is pressed: #359 (comment)

rohanpurohit:
This PR includes:

### Issue #390:
- cross-checked replies and comments from sudhir and updated wherever necessary, mostly additional info.
- remove `isVentilatorOn` useState, `ventilating` selector is used
- remove `landingLabel` useState 

### Issue #331:
- remove `useEffect` calling `initParameterUpdate`,currently called from `StartPauseVentilation` button's `OnClick` handler.

### Issue #361:
- Updated the `StartPauseVentilation` button's `OnClick` handler to handle these 3 scenarios [comment](https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-863733123)
- Scenario 1 & 2 work as described
- **Scenario 3 (current behavior after this PR):**
1. Ventilation is inactive
2. Firmware disconnects
3. User presses "Starts Ventilation" button
4. The button is not disabled, the button label is still `Start ventilation`
5. Firmware reconnects, ventilation is not started

### Misc Refactor/feature addition:
- added  `dispatchParameterRequest` method for better code-readability. 
- added a separate button for `LandingPage` to simplify code and to avoid `UseStates`.
- Added modal popup on pressing `Pause Ventilation` when there are unsaved changes in alarm limits
- Added a `ModalContent` component that has `discard or Keep Previous Values` content of alarm modal.
- `ModalContent` is reused in `AlarmsPage`, `SideBarSlideRoute`, `ToolBar`
- `Mute button` is disabled on connection loss to firmware
- Added TODO's from some of Sudhir's comments
- remove `eslint-disable-next-line react-hooks/exhaustive-deps`


rohanpurohit:
@ethanjli observations from testing on my end:
- On disconnecting firmware, the `start` button now shows `Connecting`, but on plugging it back in, could you describe what do we want? currently, it toggles ventilation, which I suppose is undesirable. let me know!

rohanpurohit:
@ethanjli This PR is ready for testing/code review. I have updated the PR description with the latest changes, in terms of code quality:
-  I did add a `useState` for opening the modalPopup on `Pause Ventilation` and a separate `OnConfirm` handler for the same.
- ~~Other than that a different `useState` for `disabling` the buttons as the `StartPauseVentilation` button is not really `disabled` now other than when it says `Connecting...` following the scenarios here [#361 comment](https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-863733123).~~

ethanjli:
Thanks, I will test and review. Were you able to identify the source of the behavior you reported about the backend's "firmware disconnected" alarm not causing the frontend to show it as an active alarm (with a red border)?

rohanpurohit:
> Thanks, I will test and review. Were you able to identify the source of the behavior you reported about the backend's "firmware disconnected" alarm not causing the frontend to show it as an active alarm (with a red border)?

sorry! I forgot about that, moreover so because we have disabled the mute button on firmware disconnection, but I realize we have to find the cause for that regression. I will do that now!

rohanpurohit:
@ethanjli the behavior actually makes sense, so the mute button sends a request when it's clicked to change the `active` state of `AlarmMute` message and that certainly won't happen when the firmware is disconnected, and the `red border` and `audio` for alarms looks at the `getAlarmMuteActive` selector which is `AlarmMute` message `active` field, so whatever was the `state` before firmware was disconnected `remains`, that is if `mute` was active before disconnection, there won't be a border/audio and vice-versa. for the `backend disconnection` we look at just `AlarmMuteRequest` `active` therefore we can toggle the border and audio.

ethanjli:
I see. From your explanation, I must have misunderstood what the behavior was. I had thought that you meant that when the firmware got disconnected, backend created an active alarm but that _never_ triggered the red border in the frontend (even when alarms weren't muted). Instead, it looks like the behavior is as follows:
- When alarms aren't muted, the "firmware disconnected" alarm behaves just like any other alarm, which is what we expect.
- When alarms are muted and then the "firmware disconnected" alarm is generated, the frontend don't cancel the active alarm mute. So this isn't a regression, it's just missing functionality which needs to be implemented as part of #372. As you say, the behavior makes sense for how we've currently implemented it.

This observation is useful for informing how we should implement #372, though I haven't thought about that enough to propose any concrete approaches.

ethanjli:
Observations from testing against #361: everything works as expected, except when ventilation is inactive and the firmware disconnects - the comment I made at https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-863733123 only partially specified the desirable behavior, by describing a less desirable behavior. Perhaps what we should have here is that when the backend or firmware are disconnected in standby mode, then the "Start Ventilation" is disabled and says either "Loading...", "Connecting...", or something else. Let's discuss to see if this behavior would make the most sense for usability and safety, or if we should have some other behavior.

I'm also wondering if it might make more sense to say "Loading..." instead of "Connecting...", or even to have the button label say "Starting Ventilation..." or "Pausing Ventilation..." in their corresponding situations instead of "Connecting...".

Additional observations:
- When I'm on the Alarm Limits screen during ventilation and press "Pause Ventilation", I stay on the Alarm Limits screen. When I'm on the Settings screen during ventilation and press "Pause Ventilation", I stay on the Settings screen. By contrast, when I'm on either screen during standby and press "Start Ventilation", I am taken to the dashboard. For consistency, maybe it'd make sense that pressing "Pause Ventilation" should take me back to the Quickstart screen?
- ~In `EventAlerts.tsx` I see code to disable the Alarm Mute button when the firmware is disconnected. But when I actually disconnect the firmware, I don't see this button getting changing or being disabled.~ I'm not able to reproduce this, I'll look out to see if it happens again

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -163,74 +154,69 @@ export const ToolBar = ({
   const alarmLimitsRequestDraftSelect = useSelector(getAlarmLimitsRequestDraft);
   const alarmLimitsRequestSelect = useSelector(getAlarmLimitsRequest);
   const alarmLimitsRequestUnsaved = useSelector(getAlarmLimitsRequestUnsaved);
-  const alarmLimitsUnsavedKeys = useSelector(getAlarmLimitsUnsavedKeys);
   const alarmLimitsRequest = (alarmLimitsRequestSelect as unknown) as Record<string, Range>;
   const alarmLimitsRequestDraft = (alarmLimitsRequestDraftSelect as unknown) as Record<
     string,
     Range
   >;
   const ventilatingStatus = useSelector(getVentilatingStatusChanging);
```
ethanjli: Let's rename this to `ventilatingStatusChanging` so that the name isn't misleading.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -163,74 +154,69 @@ export const ToolBar = ({
...
-    if (isVentilatorOn || staticStart) {
-      history.push(QUICKSTART_ROUTE.path);
+    if (ventilating) {
+      // if ventilating and there are unsaved alarm limit changes then open modal popup
+      if (alarmLimitsRequestUnsaved) {
+        setStartDiscardOpen(true);
+        return;
+      }
+    } else {
+      // if not ventilating and the firmware/backend is disconnected then return early
```
ethanjli: Let's also add an explanation of why we should return early when the firmware/backend is disconnected.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -243,76 +229,73 @@ export const ToolBar = ({
...
-  }, [ventilating, initParameterUpdate]);
-
   /**
-   * Update Label on Button based on ventilation status
+   * Switch to dashboard page if ventilating
    */
   useEffect(() => {
-    if (ventilating) {
+    if (storeReady && ventilating) {
       history.push(DASHBOARD_ROUTE.path);
```
ethanjli, line 270: Is this the place which takes the frontend to the dashboard once ventilation starts? If so, perhaps this is where we could take the frontend to the quickstart screen once ventilation is paused.

ethanjli: This `useEffect` could be a candidate for decomposing out to a separate (very small) component which takes care of all auto-navigation, which we would include in the main layout; its behavior is independent of the toolbar so I'm not sure the code needs to live in this file. This would be for a future refactoring PR.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -243,76 +229,73 @@ export const ToolBar = ({
...
     }
-    setIsVentilatorOn(ventilating);
-  }, [ventilating, history]);
+  }, [ventilating, history, storeReady]);
 
-  const StartPauseButton = (
+  const StartPauseVentilationButton = (
```
ethanjli, line 274: The button and its behavior is also a good candidate for decomposing out to a separate component, since it has some nontrivial behavior of its own which probably doesn't/shouldn't depend on the rest of the toolbar. This would be for a future refactoring PR.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -397,30 +396,16 @@ export const ToolBar = ({
...
-                  }
-                }
-                return <React.Fragment />;
-              })}
-            </Grid>
-          </Grid>
-        </Grid>
+        <ModalContent />
+      </ModalPopup>
+      <ModalPopup
```
ethanjli, line 412: Is there any way to just use one ModalPopup? If we're going to decompose the Start/Pause Ventilation button out anyways, then it could make sense to keep two ModalPopups (as one ModalPopup would be moved out of the Toolbar component and into the Start/Pause Ventilation component)

ethanjli (commented):
Testing results:
- When I interact with the modals and try to close them in various conditions, they all seem to behave correctly
- The alarm limits modal seems to correctly commit changed alarm limits to the store.
- When I change settings from the Settings screen, the changes seem to get dispatched correctly into the store
- I haven't noticed any problems introduced by this PR's refactoring

So the remaining changes we've discussed (as far as I can remember) are:
- Rename the `ModalContent` component to express that it's the modal content for discarding alarm limits changes
- Improve the behavior of the Start Ventilation button in standby mode when the firmware/backend disconnects, and make the button text more descriptive of the current state of the application (including a decision table, which we can use as a formal specification for verifying the frontend's behavior in unit tests, debugging, etc.)

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: > Is this the place which takes the frontend to the dashboard once ventilation starts? If so, perhaps this is where we could take the frontend to the quickstart screen once ventilation is paused.

Yes this is the useEffect that does that. yep most likely

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: So the problem here really is in the `onClick` handler and the `open` attribute which decides when the modal should open and those two behaviours need to be distinct which is why two ModalPopup

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: This probably will change considering the new scenarios that we discussed, although I'll add a comment of whatever i do in the end.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -242,15 +243,25 @@ export const ToolBar = ({
...
-   * Set Label for Start/Pause Ventilation button
+   * Enable/Disable and set label for Start/Pause Ventilation button
    */
   useEffect(() => {
-    if (ventilatingStatus) {
-      setLabel('Connecting...');
+    if (!ventilating) {
+      setIsDisabled(ventilatingStatusChanging || !firmwareConnected || !backendConnected);
+      if (!firmwareConnected || !backendConnected) {
+        setLabel(ventilatingStatusChanging ? 'Starting...' : 'Connecting...');
```
ethanjli: This logic is different from https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-867123868 , which specifies that when we're in standby and there's a disconnection, the button should show "Connecting..." regardless of the value of request.ventilating. I've been going back and forth in my head about whether we want it to behave as specified in that table, or whether we want it to behave as you've implemented (which is how you specified it in row 4 of https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-866649898). I see the advantage of your approach, in the following sequence of events, an extreme edge case we hadn't discussed:

1. The firmware & backend are connected, and the ventilator is in standby mode.
2. The user presses the "Start Ventilation" button.
3. While the message is in transit, then the button changes to say "Starting..." and is disabled
4. While the message is still in transit, the firmware/backend becomes disconnected. If we follow your implementation, the button remains the same. If we follow the decision table I wrote, then while the button stays disabled the logic will automatically change the button to say "Connecting...". Either way, the user will see an alarm indicating loss of software connection.

Your implementation makes the behavior in standby mode more symmetric with the behavior in dashboard mode. Then we can express the logic as follows:

```
if (response.ventilating) {
  setIsDisabled(...)
  setLabel(ventilatingStatusChanging ? 'Pausing...' : 'Pause Ventilation') - note that your line 259 is redundant to line 262, except that line 262 is missing a third period in the "..."
  return;
}

setIsDisabled(...)
if (ventilatingStatusChanging) {
  setLabel('Starting...')
  return;
}

setLabel(firmwareConnected && backendConnected ? 'Start Ventilation' : 'Connecting...')
```

If you agree that we should follow your implementation, can you add an updated version of the decision table, and a link back to this comment, as a new comment in #361?

ethanjli (commented):
Here's my review of the Start/Pause Ventilation button logic. We can make it a bit simpler/clearer, but after that's done I think this PR is ready to merge. I tested the button in all of the cases (including the sequence of events I listed in the comment in this review), and it seems to work fine.

For records-keeping:

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: > This logic is different from [#361 (comment)](https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-867123868) , which specifies that when we're in standby and there's a disconnection, the button should show "Connecting..." regardless of the value of request.ventilating. I've been going back and forth in my head about whether we want it to behave as specified in that table, or whether we want it to behave as you've implemented (which is how you specified it in row 4 of [#361 (comment)](https://github.com/pez-globo/pufferfish-software/issues/361#issuecomment-866649898)). I see the advantage of your approach, in the following sequence of events, an extreme edge case we hadn't discussed:

Honestly i had realised this last night and was going to fix it this morning haha!

> 1. The firmware & backend are connected, and the ventilator is in standby mode.
> 2. The user presses the "Start Ventilation" button.
> 3. While the message is in transit, then the button changes to say "Starting..." and is disabled
> 4. While the message is still in transit, the firmware/backend becomes disconnected. If we follow your implementation, the button remains the same. If we follow the decision table I wrote, then while the button stays disabled the logic will automatically change the button to say "Connecting...". Either way, the user will see an alarm indicating loss of software connection.
> Your implementation makes the behavior in standby mode more symmetric with the behavior in dashboard mode. 

But considering the case you mentioned and the code snippet that simplifies the logic, i agree that this is more symmetric to what we have in dashboard mode.


> If you agree that we should follow your implementation, can you add an updated version of the decision table, and a link back to this comment, as a new comment in #361?

Yes, Let's follow this? I'm adding the comment to the issue.

rohanpurohit:

ethanjli (approved):

PR #392, opened 2021-06-18
Generate firmware log events & alarms for connection/disconnection to backend

### LINKED ISSUE #353: Firmware should activate/generate events/alarms on startup & disconnection/reconnection
When the firmware starts up, it should create a log event indicating this; similarly, when the backend server starts up, it should create a log event indicating this. During hardware faults, would enable troubleshooting about when a loss of connection is due to loss of connection or due to a crash+restart. During normal operation, this also allows us to log when the ventilator was powered on.

When the firmware is unplugged from the computer, it should generate an active alarm for lost connection to the backend; when the firmware is plugged back in, it should deactivate that alarm and generate another (non-alarm) event for connection to the backend is up. This issue is split off from side-remark in #320. We may want to disaggregate the log event codes, so that the firmware's "lost backend connection" and "backend connection is up" events are distinct (and have different messages) from the frontend's "lost backend connection" and "backend connection is up" events. Otherwise the user will see two of each event with no explanation of why they're duplicated. The firmware's events should be triggered by whether or not its Backend driver has received a new StateSegment within some timeout - this is the same approach used by the python backend server.

ethanjli:
This PR fixes #353 by adding a message receipt timer to the `Driver::Serial::Backend::Backend` class with a timeout of 500 ms and using it to set a boolean flag in the store about whether the connection is active, and by adding a `Driver::Serial::Backend::AlarmsService` class to update the events log on rising and falling edges for that flag.

This PR also lightly refactors the various `AlarmsService` classes by taking the responsibility for updating `store.active_log_events()` out of their transform methods; instead, `main.cpp` calls `alarms_manager.transform(store.active_log_events())` once during every iteration of the main loop.

Finally, this PR changes the behavior of the Nucleo's LED LD1 so that it blinks slowly when running normally and is solid-on when an alarm is active. This allows some visibility into the state of alarms on the Nucleo board when it's not connected to the frontend.

For records-keeping:

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Application/LogEvents.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Application/LogEvents.h
@@ -25,6 +25,7 @@ class IncrementalEventsSender {
   // the provided event is not modified.
   void add_event(Element event, uint32_t &id);
   void add_event(const Element &event);
+  void add_event(LogEventCode code, LogEventType type);
 
  private:
   LogEventsSender &sender_;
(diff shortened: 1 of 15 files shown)


ethanjli:
@rohanpurohit Can you try this out on your setup to see if it works for you?

rohanpurohit (approved):
Everything looks good.

- I see a blinking light when it's running normally
- A solid green light when an alarm is active

PR #400, opened 2021-06-20
Reimplement state synchronization in backend for nestable schedules

ethanjli:
This PR prepares the backend for implementation of the Event Synchronization application-level protocol layer by refactoring the backend's implementation of the State Synchronization protocol layer (but without changing the State Synchronization protocol specification), specifically:
- By replacing the `protocols.application.states.Synchronizer` class with a more general `protocols.application.states.SequentialSender` class with more general functionality and without timing-related functionality, and a `protocols.application.states.TimedSequentialSender` class which wraps around `SequentialSender` with a constant send interval. Compare to #398, which performed the same refactor but did not provide a `TimedSequentialSender` class because the firmware only needed one root `SequentialSender` and one timer, for sending states to the Python backend; while the Python backend requires three `SequentialSenders` and three timers, for sending states to the firmware, the frontend, and the filesystem - so encapsulating timing-related functionality is useful in the Python backend.
- Otherwise, the concepts and schedule configurations for the State Synchronization implementation are the same as in #398.

I have fully removed the unit tests which were written for `states.Synchronizer`. We should hold off on writing unit tests for `protocols.application.states`, as the interfaces of the classes may undergo further changes for implementation of the Event Synchronization protocol layer.


ethanjli:
@rohanpurohit For this PR, let's merge after you test this to look for any regressions - the backend (both `ventserver.simulator` and `ventserver.application`) should still behave the same as before this PR.

For records-keeping:

rohanpurohit (approved):
Looks good! backend behaves as expected both in `simulator` and `application`

PR #401, opened 2021-06-20
Reimplement state synchronization in frontend for nestable schedules

ethanjli:
This PR prepares the frontend for implementation of the Event Synchronization application-level protocol layer by refactoring the frontend's implementation of the State Synchronization protocol layer (but without changing the State Synchronization protocol specification), specifically:
- By refactoring and generalizing `store/controller/saga.ts`'s `sendAll` generator function for a more generic State Synchronization protocol implementation
- Otherwise, the concepts and schedule configurations for the State Synchronization implementation are the same as in #398.

This PR also refactors and reorganizes the protocol implementation code into a file/folder structure which is more consistent the structure used in the backend and firmware, fully separates transport-level protocol code from application-level protocol code, and makes the protocol implementations more generic.


ethanjli:
@rohanpurohit For this PR, let's merge after you test this to look for any regressions, including performance or CPU usage regressions on the RPi with the refactored backend and the refactored frontend running together (the refactored backend has been merged from develop onto this branch).

For records-keeping:

rohanpurohit (approved):
Seems to be running fine on my end. did you feel like there was a drop in performance @ethanjli or just suspecting from the changes made?

ethanjli:
I didn't suspect a performance drop, it's just that the two ways I can think of to catch implementation errors for the output schedules are:

1. Inspecting the sequence of sent events, to see if the order of events is incorrect. I already did this and it looked fine to me.
2. Checking whether events are being sent at an incorrect rate. If the rate is too high, it would probably cause a performance drop; if the rate is too low, there would probably be too much latency from the frontend to the backend. However, we can check this more directly by just printing the current time together with the event when it's sent, to see if they have the correct timing.

The other thing is that these refactors makes the software a bit more complicated, so if that extra complexity caused a performance drop then we'd need to reevaluate; but it sounds like we're fine.

PR #402, opened 2021-06-21
Generate alarms in firmware for sensor disconnection and low battery power

rohanpurohit:
This PR adds:
- alarms for sfm drivers connection loss
- Low battery alarm for battery power less than 30%
- Critical battery alarm for battery power less than 5%


backend/ventserver/simulation/power_management.py
```diff
@@ -33,6 +33,28 @@ def transform(
...
+                code=mcu_pb.LogEventCode.battery_low,
+                event_type=mcu_pb.LogEventType.system
+            ))
+        else:
+            events_log.input(
+                alarms.AlarmDeactivationEvent(
+                    codes=[mcu_pb.LogEventCode.battery_low]
+            ))
+
+        if power_management.power_left <= 5:
```
ethanjli: Is there any particular reason why low battery is `<` while critical battery is `<=`? (same thing applies to the firmware implementation)

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/ControlLoop.h
```diff
@@ -19,9 +19,13 @@
...
 namespace Pufferfish::Driver::BreathingCircuit {
 
+static constexpr auto alarm_codes = Util::Containers::make_array<LogEventCode>(
+    LogEventCode::LogEventCode_sfm3019_air_disconnected,
+    LogEventCode::LogEventCode_sfm3019_o2_disconnected);
+
 class ControlLoop {
  public:
-  virtual void update(uint32_t current_time) = 0;
+  virtual void update(uint32_t current_time, Application::AlarmsManager &alarms_manager) = 0;
```
ethanjli: Is there any way we can avoid making the `ControlLoop` classes be dependent on (or aware of) `AlarmsManager`? For example, perhaps we could just write `ControlLoop` sensor connected/disconnected `bool`s to a struct which is provided as an output parameter, and then have a `BreathingCircuit::SensorAlarmsService` which transforms the struct into `AlarmsManger` changes. Compare this proposed approach to how the `Controller` class writes things to the `actuator_setpoints` and `actuator_vars` output parameters. (In this proposed approach, we might rename our current `BreathingCircuit::AlarmsService` to `BreathingCircuit::AirwayAlarmsService` or something; later, we would also move the SpO2 and HR alarms out of the `BreathingCircuit` driver and into the `Nonin` driver).

Reasons for this:
- `ControlLoop`'s only responsibility should be to get readings from sensors and output actuation values, not to activate/deactivate alarms.
- `ControlLoop` is going to become much more complicated in the future, so let's keep it as simple as possible.
- Activation of other alarms has already been decomposed out to other classes, so let's continue separating alarm activation/deactivation into other classes.
- By reducing interdependencies, we can simplify unit tests - so that the unit tests for `ControlLoop` are totally independent of `AlarmsManager`

Currently `ParametersService` does depend on `LogManager` as well as `AlarmsManager`; if we can find a way to decouple them, we should.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/AlarmsService.h
```diff
@@ -13,13 +13,23 @@
...
+  void transform(
+      const MCUPowerStatus &mcu_power_status, Application::AlarmsManager &alarms_manager) const;
+  static void generate_alarm(
+      const LogEventCode &alarm_code,
+      bool alarm_active,
+      Application::AlarmsManager &alarms_manager);
+
+ private:
+  const float battery_low_power = 30;
+  const float critical_battery_power = 5;
```
ethanjli: I think we can make these `static constexpr float`s.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/AlarmsService.h
```diff
@@ -13,13 +13,23 @@
...
+    LogEventCode::LogEventCode_battery_low,
+    LogEventCode::LogEventCode_critical_battery);
 
 class AlarmsService {
  public:
-  static void transform(
-      const MCUPowerStatus &mcu_power_status, Application::AlarmsManager &alarms_manager);
+  void transform(
+      const MCUPowerStatus &mcu_power_status, Application::AlarmsManager &alarms_manager) const;
+  static void generate_alarm(
```
ethanjli: If we can make this method `private`, then we should.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/AlarmsService.cpp
```diff
@@ -9,15 +9,34 @@
...
-    const MCUPowerStatus &mcu_power_status, Application::AlarmsManager &alarms_manager) {
+    const MCUPowerStatus &mcu_power_status, Application::AlarmsManager &alarms_manager) const {
+  bool charger_disconnected = false;
   if (!mcu_power_status.charging) {
-    alarms_manager.activate_alarm(
-        LogEventCode_charger_disconnected, LogEventType::LogEventType_system);
-  } else {
-    alarms_manager.deactivate_alarm(LogEventCode_charger_disconnected);
+    charger_disconnected = true;
+  }
```
ethanjli: Can this be simplified to `bool charger_disconnected = !mcu_power_status.charging`?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/AlarmsService.cpp
```diff
@@ -9,15 +9,34 @@
...
+  bool battery_low = false;
+  if (mcu_power_status.power_left < battery_low_power) {
+    battery_low = true;
   }
+  bool critical_battery = false;
+  if (mcu_power_status.power_left <= critical_battery_power) {
+    critical_battery = true;
+  }
```
ethanjli: When we are on critical battery, does that mean that both the `battery_low` and `critical_battery` alarms are simultaneously active? The logic might be clearer if you make an enum for battery level, with `critical`, `low`, and `normal` levels, and then activate/deactivate alarms based on the enum value.

proto/mcu_pb.proto
```diff
@@ -158,6 +158,10 @@ enum LogEventCode {
   backend_started = 141;
   mcu_shutdown = 142;
   backend_shutdown = 143;
+  sfm3019_air_disconnected = 144;
+  sfm3019_o2_disconnected = 145;
+  fdo2_disconnected = 146;
+  critical_battery = 147;
```
ethanjli: The battery alarm codes should be next to each other, and their names should be consistent (e.g. `battery_low` and `battery_critical` have consecutive values such as 138 and 139). Feel free to renumber the enum values as needed.

frontend/src/modules/logs/EventType.tsx
```diff
@@ -63,27 +57,20 @@ export const getEventDetails = (event: LogEvent, eventType: EventType): string =
...
-    if (event.code === LogEventCode.mcu_backend_connection_up) {
-      return 'Hardware controller connected';
-    }
-    if (event.code === LogEventCode.backend_frontend_connection_up) {
-      return 'User interface connected';
-    }
-    if (event.code === LogEventCode.charger_disconnected) {
-      return 'Battery charger is disconnected';
-    }
+    const eventDetailsMap: Record<number, string> = {
```
ethanjli: `store/controller/types.ts` uses `Map` instead of `Object` (which you've type-annotated as `Record` here) - since all keys are known before the frontend is started, and the keys are enums rather than strings, maybe it makes sense to use `Map` here too? (this also applies to `EventTypeMap` below)
https://github.com/pez-globo/pufferfish-software/blob/0695a3e034d35ef18126f10fe8acd058c480ec40/frontend/src/store/controller/types.ts#L91

The MDN Web Docs page for `Map` has a useful table comparing the tradeoffs between `Map` and `Object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

Since `Map` exposes a `has` method, we can use a more clear/explicit check for a missing event code (`if (!eventDetailsMap.has(event.code))`) to return a default value, instead of using `||`.

ethanjli (commented):
Here's my review of the code. Since the added functionality is very simple compared to the suggested changes, I'll test after changes are made.

rohanpurohit:
Observations from testing:
- Sensor disconnection/connection alarm is triggered only once and the backend/frontend needs to be restarted for the change to take place. so if there is an alarm for disconnection after removing the sensor, it's not deactivated if the sensor is plugged back in.

frontend/src/modules/logs/EventType.tsx
rohanpurohit: Switched to map, and using `eventDetailsMap.get(event.code)` since we need that anyway then handling the `undefined` by returning default value.

proto/mcu_pb.proto
rohanpurohit: yep! was thinking whether you wanted them in that order.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Power/AlarmsService.cpp
rohanpurohit: Yeah initially i implemented that way, now I'm deactivating `battery_low` if `critical_battery` is active. and used enum class

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/AlarmsService.h
rohanpurohit: removed the method now.

backend/ventserver/simulation/power_management.py
rohanpurohit: No! just missed it.

backend/ventserver/simulation/power_management.py
```diff
@@ -33,6 +33,32 @@ def transform(
...
```
ethanjli: When `power_left <-==5`, it looks like  the backend simulator is going to activate and then deactivate the `battery_low` alarm. This means that if the simulator starts up with critical battery levels, then the firmware will create a spurious `battery_low` log event.

Here are two ways you could implement this differently:
1. First do your if/else case handling to convert your numerical `power_left` into an enum representing power level bins (normal, low, critical). Afterwards, do if/else case handling to convert the power level bin to alarm activation/deactivation. This would be more similar to what the firmware does.
2. Figure out the correct if/else case handling to directly convert `power_left` into alarm activation/deactivation. For example, this could look like `if power_left <= 5, do something, else if power_left <= 30, do something else, else do something else`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/ControlLoop.h
```diff
@@ -16,6 +16,7 @@
 #include "Pufferfish/Driver/I2C/SFM3019/Sensor.h"
 #include "Pufferfish/HAL/Interfaces/PWM.h"
 #include "Pufferfish/Util/Timeouts.h"
+#include "SensorAlarmsService.h"
```
ethanjli: `ControlLoop.h` should not depend on (or be aware of) the existence of a `SensorAlarmsService.h`. Here are two ways we could remove this dependency:

1. Move the declaration of `SensorConnections` into this file. After all, it's part of the public interface for the `ControlLoop` classes.
2. Move the declaration of `SensorConnections` into a new file, e.g. `Sensors.h`, which you include from both `ControlLoop.h` and `SensorAlarmsService.h`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/AlarmsService.h
```diff
@@ -13,13 +13,21 @@
...
-    Util::Containers::make_array<LogEventCode>(LogEventCode::LogEventCode_charger_disconnected);
+static constexpr auto alarm_codes = Util::Containers::make_array<LogEventCode>(
+    LogEventCode::LogEventCode_charger_disconnected,
+    LogEventCode::LogEventCode_battery_low,
+    LogEventCode::LogEventCode_battery_critical);
 
 class AlarmsService {
  public:
-  static void transform(
+  enum class BatteryLevel { critical, low, normal };
```
ethanjli: Can this enum be made `private`?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/ControlLoop.cpp
```diff
@@ -43,6 +47,18 @@ void HFNCControlLoop::update(uint32_t current_time) {
   // TODO(lietk12): we should probably set flow to NaN otherwise, but for now we do nothing
   // so that we don't overwrite the simulated values if the sensors aren't available
 
+  if (air_status == InitializableState::ok) {
+    sensor_connections_.sfm3019_air_connected = true;
+  } else if (air_status == InitializableState::failed) {
+    sensor_connections_.sfm3019_air_connected = false;
+  }
```
ethanjli: Are there any other possible `InitializableState` values to account for? If not, could we just do `sensor_connection_.sfm3019_air_connected = air_status == InitializableState::ok`?

frontend/src/modules/logs/EventType.tsx
```diff
@@ -63,27 +57,25 @@ export const getEventDetails = (event: LogEvent, eventType: EventType): string =
...
-    }
-    if (event.code === LogEventCode.mcu_backend_connection_up) {
-      return 'Hardware controller connected';
-    }
-    if (event.code === LogEventCode.backend_frontend_connection_up) {
-      return 'User interface connected';
-    }
-    if (event.code === LogEventCode.charger_disconnected) {
-      return 'Battery charger is disconnected';
+    const eventDetailsMap = new Map<LogEventCode, string>([
```
ethanjli: If we can move our `Map`s out to file-level constants, let's do that.

ethanjli (commented):
Generally looks good, just some minor changes. Here are my comments on your testing observations:

- After we disconnect the I2C sensors, reconnecting them will not cause the firmware to recognize them correctly again; I believe the STM32's I2C HAL doesn't support that, but I might also be wrong. Thus, the "sensor disconnected" alarm should be permanent until after the firmware restarts. For the Nonin OEM sensor, we will be able to have the firmware receive data from the sensor after it's sensor is disconnected and then reconnected, because it's a UART - so for that sensor the firmware will be able to deactivate the "sensor disconnected" alarm upon reconnection.
- Currently, the backend does not save the event log to the filesystem at runtime or load it from the filesystem upon startup. This means that if you restart both the backend and the frontend, all log events which were generated by the firmware will be lost, including active alarms such as the sensor disconnect alarms. I'm not sure what you meant when you said "the backend/frontend needs to be restarted for the change to take place" - I was able to trigger the sensor disconnection alarm by unplugging the sensor, and the disconnection alarm showed up on the frontend without me having to restart the backend or frontend. The only way to hide the alarm in the frontend is to restart both the backend and the frontend (perhaps this is what you meant to say); however, as indicated by the solid-on LED in the Nucleo board, the "sensor disconnected" alarm is still active on the firmware.
- We don't have any sensor connection alarms, right? When you said "sensor disconnection/connection alarm", you meant that sensor reconnection should just deactivate the "sensor disconnected" alarm.

frontend/src/modules/logs/EventType.tsx
rohanpurohit: added constants.ts in src/modules/logs

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/BreathingCircuit/ControlLoop.cpp
rohanpurohit: > Are there any other possible `InitializableState` values to account for? 

Nope!
> If not, could we just do `sensor_connection_.sfm3019_air_connected = air_status == InitializableState::ok`?

yep done.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Power/AlarmsService.h
rohanpurohit: Fixed.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/BreathingCircuit/ControlLoop.h
rohanpurohit: Added Sensors.h and included in `ControlLoop` and `SensorAlarmsService`

frontend/src/modules/logs/EventType.tsx
ethanjli: I really like the approach of moving them into a separate file, I hadn't thought of that! It makes sense here because the `Map`s are so big.

ethanjli:
The battery alarms look good in testing, for both the firmware and the backend. Right now our implementation assumes that activation of a new alarm event will not cause alarms to be automatically unmuted - I believe that's what we decided to do, so this assumption is reasonable. If one day we decide to make generation of a new alarm event automatically unmute alarms, then when we're charging the battery from wall power and it goes from 5% to 6%, alarms will be unmuted - which is undesirable. I don't think we need to worry about this.

For how the frontend displays the sensor disconnection events, I'd suggest that the event text should be "Internal sensor failed", and then the Details text should say which sensor was lost, e.g. "O2 sensor connection lost". Otherwise, everything looks good to merge. For records-keeping: 

rohanpurohit:
Added change for label and detail

ethanjli (approved):

PR #403, opened 2021-06-21
Implement Event Synchronization protocol in backend for early delivery of changed states

ethanjli:
## State Synchronization protocol implementation extensions

This PR implements the Event Synchronization application-level protocol layer. As part of this, it adds some functionality to the State Synchronization protocol implementation:

- A `skip_unavailable` flag is added to `protocols.application.states.SequentialSender`, which is defaulted to `False` (corresponding to the previous behavior before this flag was added, which is the standard behavior specified for State Synchronization). When this flag is set to true, if the next index in the `SequentialSender`'s schedule corresponds to an output which is unavailable (i.e. `None`), the `SequentialSender` will keep advancing through the schedule until it either 1) produces an available output or 2) has gone through the entire schedule once. This functionality is used by the Event Synchronization implementation.
- A `last_index` property method is added to `SequentialSender` to expose the index corresponding to the last produced `StateSegment` - this is needed for Event Synchronization to remove that `StateSegment` from the set of sendable segments. Since this method not part of the `output` method, it obviously should only be used in synchronous environments (e.g. only within the `output` method of another Filter class). This won't be a problem for our Python backend server.

Currently, the [State Synchronization protocol specification](https://www.notion.so/pufferfish/Communication-Protocols-c6f5d4d91c9e4e178f0de883581a3142#82bd1cda5c474bfabd1c06252658a4c5) does not specify `skip_unavailable` behavior in State Synchronization scheduling; this is purely provided to enable the Event Synchronization implementation.

## Event Synchronization protocol implementation notes

This PR uses the State Synchronization protocol implementation to implement Event Synchronization, by:

- Adding a `protocols.application.events.FilteredSender` class which wraps around a `protocols.application.states.IndexedSender` object and produces `None` (i.e. segment is unavailable) for any indices not added to an `allowed_indices` set. This is used to wrap around the `protocols.backend.states.Store` to only allow producing state segments for specified message types (i.e. those which are marked as "sendable" as described by the Event Synchronization protocol specification).
- Adding a `protocols.application.events.NotificationSender` class which takes an output schedule and a `Store` object and produces "sendable" events according to the schedule.
  - `NotificationSender` is a filter; calling its `input` method with an index marks the `StateSegment` for that index as "sendable". The `NotificationSender` automatically removes the "sendable" flag for a `StateSegment` once it produces it as output.
  - `NotificationSender` also has an `output_idle` flag (by default set to `False`).
    - When this flag is `False`, if there are no available `sendable` events to output, then the `NotificationSender` produces `None`.
    - When this flag is `True`, if there are no available `sendable` events then the `NotificationSender` will produce `StateSegment`s according to a separate copy of its schedule, ignoring the `sendable` flag. This separate copy of the schedule acts independently and only when the `EventSender` is "idle", and if events become "sendable" and then are all sent (and thus are no longer "sendable"), then the idle copy of the schedule picks up where it left off before events became "sendable"
    - This flag provides the functionality allowed in the Event Synchronization specification, which states that: 'When no segments in this schedule are "sendable", the sender may send every segment in this schedule (in effect falling back to a static send schedule).'
    - Right now, the backend server sets `output_idle` to false for all of its Event Synchronizers, to help reduce the CPU usage of handling lots of messages. However, the trade-off here is that it takes longer for a newly connected client to receive all states. Depending on performance characteristics we may choose to set `output_idle` to true, or we may decide we don't need it at all and remove it entirely, which would simplify the Event Synchronization implementation. I think we might want to use `output_idle` on the UART connection (where messages might be lost) but not on the WebSocket connection (where messages are never lost); and definitely not on the filesystem output (which would cause unnecessary wear on the flash memory).
- Adding a `protocols.application.events.ChangedStateSender` class which is a `protocols.application.states.Sender` which wraps around `NotificationSender` and just keeps track of which states in `all_states` have changed since the last time `output` was called, and marks them as "sendable". `ChangedStateSender` can accept a `ResetEvent` as its `input` method, which makes the `ChangedStateSender` act as if all events have been changed; this is so that when a device changes from disconnected to connected, the `ChangedStateSender` will try to send all possible states, in case the device restarted and needs to be completely (re)initialized. This allows faster reinitialization of states after a device restarts in the average case, which is especially useful when we set `output_idle` to false.

## Revisions to Event Synchronization protocol specification

This PR accompanies a last-minute revision to the [Event Synchronization protocol specification](https://www.notion.so/pufferfish/Communication-Protocols-c6f5d4d91c9e4e178f0de883581a3142#99cbbd233f834a3087c702ebe8643b4a) to remove event acknowledgements and thus remove the need to implement any receiver-related functionality for Event Synchronization. This simplification was made because:
- It greatly simplifies the Event Synchronization protocol behavior, as well as the implementations. Acknowledgements were the most complicated part of the protocol, and removing them makes the protocol's correctness and performance (under normal as well as degraded conditions) much easier to reason about, to verify in testing, and to understand from inspecting logs of message traffic.
- By avoiding the need for receivers to send acknowledgement messages, we reduce the number of messages which need to be sent over communication links and processed by senders; or we increase the throughput of state update messages for a given send interval. This is important because we are already running at the per-core CPU limits on the RPi.
- Removing acknowledgements also means removing event notification retransmission functionality, which was the original reason for adding acknowledgements. Now that we've removed event notification retransmission, we have a clearer separation of responsibilities between the application-layer protocols: State Synchronization only deals with retransmission of events on a static schedule (to guarantee eventual delivery of the most recent value of each state variable), while Event Synchronization only deals with dynamic scheduling of state update messages (to provide earlier sending of state updates).
- We are stacking the Event Synchronization layer on top of the State Synchronization layer with the requirement that State Synchronization must also include a static schedule for sending every state which Event Synchronization might send. Thus, we can also guarantee eventual delivery of any states sent by Event Synchronization, even when a Event Synchronization message is lost in transit. Acknowledgement-based retransmission was meant to guarantee slightly earlier eventual delivery of states in the event of the loss of an event notification message in transit. However, this only provides a performance optimization when an event notification message is lost (which, at least in testing so far, is uncommon on the UART and never happens on the WebSocket), and the worst-case scenario (when all Event Synchronization retransmissions are lost in transit) is just a fall back to State Synchronization. The extra complexity of acknowledgement-based retransmission is not worth it for our use case; this is especially true on the WebSocket connection. Without acknowledgements, if a message is lost in transit then we simply wait for the State Synchronizer's static schedule (or the Event Synchronization sender's idle schedule, if it has `output_idle` enabled) to send it again.
- By removing the need for receiver-side functionality for Event Synchronization to work, we avoid adding mutual dependencies between the sender and receiver; then the behavior of the Event Synchronization protocol only depends on the sender, so we don't need to worry about emergent behaviors from the interaction between the sender and the receiver.
- It reduced the amount of work and troubleshooting I needed to do to get a working Event Synchronization implementation, so that I could achieve a proof of concept by modifying the backend and without thinking at all about the firmware or frontend.

## Testing

I've tested this PR on the RPi with the rotary encoder, and now the frontend's `ValueClicker` is much more responsive to the rotary encoder. It's still not realtime, which I suspect is because there's a minimum 30 ms delay between successive `RotaryEncoder` messages.


ethanjli:
@rohanpurohit If you could test the rotary encoder and report your observations on its new latency, what the user experience feels like, and how much of a problem the new level of latency feels like, that would be useful to guide whether we need to look into further reducing the latency.

rohanpurohit:
> @rohanpurohit If you could test the rotary encoder and report your observations on its new latency, what the user experience feels like, and how much of a problem the new level of latency feels like, that would be useful to guide whether we need to look into further reducing the latency.

@ethanjli Just tested it and also compared it with develop to get a fresh perspective, and I really like the new latency, feels `real time` compared to what we had earlier where there was a clear delay, almost like you are putting extra effort for the values to change haha! In my opinion currently, the latency is really good, I'm not sure if there is a need to reduce it further, also I don't feel like it's too fast that the user feels like he has no control. what do you think?

ethanjli:
Great! When I rotated the dial quickly from the quickstart screen, I felt there was still noticeable choppiness and also usually a significant delay between when I stopped rotating the rotary encoder and when the number finished changing on screen. I just went back and tested again, and the latency is actually much worse on the modals from the dashboard than from the quickstart screen in standby or from the alarm limits screen during ventilation - hopefully this will be resolved when we fix #404 , since this extra latency is clearly a result of the dashboard screen. The latency on the quickstart screen is much better than before, but I'm worried that it's still a usability issue. After we do some performance profiling+optimization on the frontend and backend, maybe we'll be able to reduce the min interval of the backend's state sending schedule for the frontend from 10 ms to 5 ms, but still send SensorMeasurements only once every 30 ms and send states from the main schedule only once every 30 ms.

The other initial target for performance optimization besids #404 is to use containers in the quickstart and alarm limits screens too - currently when I press the increment/decrement button, most of the quickstart screen is redrawn, which could be causing latency due to the CPU being overloaded. I expect the same thing happens when I rotate the rotary encoder. In the alarm limits screen and in the set parameter modal, the entire screen is redrawn when I press the increment/decrement button. All of these are extremely likely to make the frontend more choppy when I rotate the rotary encoder dial.

ethanjli:
Here are my observations from testing my latest commit:

- When I first start the landing page, the "Start" button says "Loading" but it's briefly green (enabled) before quickly being changed to be disabled. @rohanpurohit Do you think there's a way we can make that button start out as disabled when the landing page loads?

## Experimental measurement of startup performance from 3a89546

- On commit 3a89546, in a non-scientific test of 15 trials of closing the frontend on the RPi (by closing its tab), waiting until the simulator backend logs the loss of the frontend, and then opening the frontend again while running the backend simulator, it takes [1.8, 1.4, 2.9, 2.2, 1.9, 1.3, 2, 1.9, 2.3, 2.1, 1.3, 1.8, 1.8, 2.4, 2.2] sec between when the landing page changes the start button to say "Loading..." and when the button changes to say "Start".
- Repeating this same test on commit 95cafc8, it takes [2.8, 1.8, 1.6, 3.1, 1.6, 2, 1.4, 1.9, 2.2, 2.5, 1.5, 2.5, 2.4, 1.7, 2.9] sec between when the landing page changes the start button to say "Loading..." and when the button changes to say "Start".
- Comparing the results before and after commit 3a89546, **I cannot conclude that commit 3a89546 significantly improved the measured performance**:
hist
The mean loading time before that commit is 2.1 s, while the mean loading time after the commit is 2 s - this is probably within the range of natural variance from my reaction speed as well as sampling randomness. We might speculate that the effect of making Event Sync send all states when the frontend connects is to reduce the variance of the frontend's loading time; it appears that sending more messages does not necessarily make the frontend reach "isStoreReady" significantly faster. For example, maybe statistically the messages being sent on the event schedule aren't reaching the frontend much sooner than the messages being sent on the main schedule (I test this in the next experiment, described below); or maybe the frontend is actually unable to keep up with the 10 ms message send rate, e.g. because it's too busy parsing messages (seems unlikely to me), running the reducers (seems possible to me) or doing redraws (seems likely to me).

It takes around 17 main-schedule states * 3 message sends per main-schedule state * 10 ms per message = 0.5 s for the backend to send all states through the main schedule, so **the frontend's actual best-case measured loading time is much worse than the theoretical worst-case state-sending time from the communication protocol's send schedule**.

To better understand what's going on, I changed the backend's interval for sending messages to the frontend from 10 ms to 500 ms and then re-ran the experiment. At this speed, I expected that my reaction speed would be a much less significant source of potential bias, and that there would be no performance bottleneck for the frontend in processing messages - instead, the speed differences should all come from the message scheduling.
- On commit 95cafc8,  it takes [14.5, 17.2, 11.7, 19.4, 19.8, 11.2, 23.5, 10.2, 13.4, 14.4, 22.4, 23.5, 23, 9.5, 23.9] sec between when the landing page changes the start button to say "Loading..." and when the button changes to say "Start".
- On commit 3a89546, it takes [10.4, 14.7, 16.3, 8.0, 13.3, 15, 12.2, 10.0, 13.3, 16.3, 15.0, 14.6, 8.9, 11.7, 19.6] sec between when the landing page changes the start button to say "Loading..." and when the button changes to say "Start".
- Comparing the results before and after commit 3a89546, I can confidently say that commit 3a89546 significantly improved performance:
hist
The mean loading time before that commit is 17 s, while the mean loading time after the commit is 13 s - this is definitely beyond the variance of my reaction speed, and the histograms have clearly different shapes. The difference is big enough that even with this small sample size, Welch's T-test (for the means of two samples from two populations with unknown unequal standard deviations) rejects the null hypothesis of the means being equal, with a p-value of 0.024.

It takes around 17 main-schedule states * 3 message sends per main-schedule state * 500 ms per message = 25.5 s for the backend to send all states through the main schedule; consistent with this, the worst-case loading time I measured was 24 sec, and it was much rarer for commit 3a89546 to approach that worst-case loading time. By changing the test to measure the state-sending schedule performance with much less of a confounding effect from human reaction speed or the performance of the frontend itself, my **experimental measurements are consistent with the theoretical predictions about (quantitative) worst-case protocol performance and (qualitative) average-case performance improvements**.

To conclude, commit 3a89546 does improve the theoretical average speed at which the backend sends all the states the frontend needs for store initialization. However, this is not practically relevant when we send messages every 10 ms, as **the performance bottleneck for initializing the frontend's store (or for the frontend to do whatever else it does at that time) appears to be elsewhere in the frontend,** not over the communication protocol. I don't think we need to worry about this delay in the near future, and once we care about start-up time we should measure start-up times with nginx serving the static build of the frontend, rather than `yarn start` serving the debug version of the frontend. The highest priority for performance optimization on the frontend is its responsiveness & speed during ventilation.

rohanpurohit:
> * When I first start the landing page, the "Start" button says "Loading" but it's briefly green (enabled) before quickly being changed to be disabled. @rohanpurohit Do you think there's a way we can make that button start out as disabled when the landing page loads?
> 

Yes I'm able to reproduce this, but since in #391 i refactored the landing button i merged develop onto this branch and checked again, and it looks like it's solved? can you also confirm the same?

ethanjli:
Great, the landing page's "Start Ventilation" button is fixed now in develop! I merged develop into this branch and confirmed that it works on this branch.

ethanjli:
For records-keeping: 

ethanjli:
@rohanpurohit What level of review and/or testing do you think we should apply to this PR? Do you think we should include this in milestone v0.12, or move it to v0.13?

rohanpurohit:
> @rohanpurohit What level of review and/or testing do you think we should apply to this PR? Do you think we should include this in milestone v0.12, or move it to v0.13?

I feel like i wont be able to do moderate testing (due to prior commitments this weekend) to get a deeper understanding of the event synchronization protocol. but I did do light to moderate testing earlier to get some intuition, and since you have done a lot of analysis and testing, in particular, I don't want to be the one to block this PR, so I feel we can include this in 0.12!

ethanjli:
Since this PR and #406 involve system-level changes, let's move them to v0.13 so that we'll have some more time to test it and look for potential regressions before including it in a tagged milestone!

rohanpurohit (approved):
Tested #406 and everything looks good on my end.

PR #406, opened 2021-06-25
Implement Event Synchronization protocol in firmware for early delivery of changed states

### LINKED ISSUE #405: Postprocess nanopb output to declare everything inside a namespace
Currently nanopb, as a C library, does not try to do anything with C++ namespaces when it declares the structs for the protobuf messages. This means that when we import `mcu_pb.h` in the firmware, it pollutes the global namespace with a bunch of declarations. Then when we need to define operator overloads for these structs (such as for the firmware's implementation of the Event Synchronization protocol), we have to declare those operator overloads in the global namespace too.

Instead of polluting the global namespace this way, we should modify the `proto/generate_mcu.sh` script to add a line before line 12 to scope all the declarations inside a namespace (e.g. `Pufferfish::Application`), and another one before line 270 to close that namespace scope. We could also instead try to close that namespace at the end of the file, but nanopb does declare a `nanopb` namespace for template specializations of `MessageDescriptor` so we'll have to check if that gets broken if we try to take that approach instead.

While we're at it, we should also upgrade the files in `Core/Inc/nanopb` from 0.4.2 to 0.4.5, since we're now using nanopb generator 0.4.5 to generate `mcu_pb.h`.

ethanjli:
This PR is a mostly one-to-one direct port of the backend's implementation of the Event Synchronization protocol from #403, but with some key differences:

- Because C++17 doesn't natively provide a `==` operator for structs, I had to add the [Boost PFR library](https://github.com/boostorg/pfr) (documentation [here](https://www.boost.org/doc/libs/develop/doc/html/boost_pfr.html)) so that we can compare two instances of structs like `AlarmLimits`. It relies on template metaprogramming magic, but it only compares the literal contents of two instances of a struct and it doesn't seem to work for us in trying to compare the contents of arrays. So we have to define custom `==` operators for `NextLogEvents` and `ActiveLogEvents`. The PFR library is third-party code licensed under the Boost Software License, which is fine for our project; I'll provide some more information in the "questions for records-keeping" checklist. Due to the way nanopb works, until we address #405 we have a templated `==` operator in the global namespace, which means you might be able to erroneously use `==` and `!=` on types which shouldn't have such an operator, or else you might see cryptic compiler errors.
- While in Python we can return `None` from a sender, in C++ we have to return a `StateOutputStatus::none` error code.


ethanjli:
For records-keeping: 

3. Does this work include, or is it based on, any third-party work which you did not create? **Yes**: this PR adds the [Boost PFR library](https://github.com/boostorg/pfr/) as a header-only library included in the `Core/Inc` directory, along with its license file. The PFR library is licensed under the Boost Software License, which is a permissive license and is compatible with our project's licenses. I've recorded the use of this library, along with the license, on [our Notion page](https://www.notion.so/pufferfish/Third-Party-Works-69c7dc67979f4a91a173103329e62423) for tracking third-party works and licenses.

ethanjli:
@rohanpurohit What level of review and/or testing do you think we should apply to this PR? Do you think we should include this in milestone v0.12, or move it to v0.13?

rohanpurohit:
> @rohanpurohit What level of review and/or testing do you think we should apply to this PR? Do you think we should include this in milestone v0.12, or move it to v0.13?

I feel like I won't be able to do moderate testing (because of other commitments this weekend) to get a deep understanding, although I had done light testing earlier and also with rotary encoder to get some intuition, and since you have done a lot of analysis and testing so I don't want to block this PR for 0.12! if that sounds good to you then maybe you can go ahead with it!

rohanpurohit:
> @rohanpurohit What level of review and/or testing do you think we should apply to this PR? Do you think we should include this in milestone v0.12, or move it to v0.13?

It looks like I can only do light testing on this one! I assume if we go ahead with #403 we'll need to merge this too? I feel like maybe we can go ahead with it if it looks good to you on your end? sorry about not responding sooner.

rohanpurohit (approved):
Tested rigorously to produce regressions if any:
- tried as many features as possible to look for any regressions, did not find any!
- tested with the spo2 sensor, everything looks good on that end.
- tested various scenarios in terms of ventilation that I could think of, triggering alarms, events, trying to force a race condition (connection/disconnection), etc. Everything looks great there as well.

Observations; 
- In terms of speed/latency everything feels snappier.
- for instance alarm muting (essentially waiting for a response from MCU, now seems faster than before).
- rotary encoder response (as earlier stated) seems better to me overall.

Apologies for not testing sooner.

ethanjli:
Great, thanks! I had forgotten to test the alarm muting to get a sense of the latency improvements for that.

PR #407, opened 2021-07-01
Fix Dashboard spurious redraws

### LINKED ISSUE #397: Refactor code in frontend modules/dashboard

### modules/dashboard:

- Removing the updateModalStatus function and replacing it with setOpen in ValueInfo and LargeValueInfo: #359 (comment) and #359 (comment)
- Marking ValueInfoProps for deprecation in ValueInfo and LargeValueInfo: #359 (comment)
- Removing isLive from ValueInfo and LargeValueInfo: #359 (comment)
- Refactoring ValueInfo and LargeValueInfo by unifying shared logic, while still providing different components to be used for different purposes, or parameterizing a single component or making its style size-responsive (e.g. for narrower grid boxes, wider grid boxes, one- vs. two-column-wide grid goxes, huge grid boxes with big font), and moving them to the modules/views folder: #359 (comment) and #359 (comment)
- Changing the ValueInfo component with a "main container" and two "sub-containers" into a "ControlValuesGroup" component which can specify the dimensions and relative layout of its constituent "ControlValuesDisplay" components (and maybe ControlValuesDisplay maybe should be renamed to ValueInfo?): #359 (comment)

### LINKED ISSUE #404: The frontend's dashboard has too many spurious redraws
When I open the frontend in Google Chrome and set the Preact Developer Tools extension to highlight all component redraws, I see that all the `LargeValueInfo` components are constantly being redrawn, even when the data hasn't changed. This is because of the way we implemented the HFNC dashboard layout as returning a bunch of `LargeValueInfo` components, rather than returning containers (i.e. `LargeValueInfo` components which were connected to selectors using `react-redux`'s `connect`). I'm not sure this is the main reason the frontend uses so much single-core CPU on the RPi (often above 100%), but in the past we found that this was definitely a major contributing factor in wasting CPU. We need to fix this.

rohanpurohit:
### Issue 404: redraws

(Files changed tab in github is slightly misleading: `ValueInfo` is now reduced more or less to how `LargeValueInfo` looked like)
- `LargeValueInfo` was a copy-paste of `ValueInfo` therefore is now removed/unified with ValueInfo.
- `ControlValuesDisplay` in `ValueInfo` is modified to accommodate for Large Value components on the dashboard.
- `ValueInfo` is basically reduced to just one component used previously as `mainContainer`, code for `subContainer1` and `subContainer2` is removed, The sub-containers in the current code base was only used in `PressureControlMainView` which in itself is `unused`, we can modify `ValueInfo` later and potentially differently to account for those changes (if we need them).
- The above change also removes an `eslint-disable react/prop-types` in ValueInfo where we were passing 3 props in one object.

Selectors:
- A selector is added to return AlarmLimits based on the `stateKey` as an array of numbers, since `ValueInfo` prop `alarmLimits` is of type `number[]`

HFNCGraphView.tsx, MultiStepWizard.tsx:
- Changes here are simply how the props should be passed to the `ValueInfo` component which is now `as-is` instead of in a `mainContainer` object

### issue 397 - Frontend refactor for dashboard
- `ValueInfo` and `ControlInfo` are now moved to `dashboard/components` as they are not containers using the `connect` from react-redux.
- `ValueInfo` is now simplified
- remove `updateModalStatus`  function in `ValueInfo` and `ControlInfo`
- some parts of the refactor are not covered, mainly providing an alternate component for subcontainers with different layouts (in my opinion, can be done when we need them i.e: different modes).


frontend/src/modules/dashboard/components/ValueInfo.tsx
```diff
@@ -1,17 +1,22 @@
...
+ * Each Value Info must have 1 main Container & optional 2 subContainer
+ * This matches with Dashboard  Display Value Layout design
  *
  */
-import { Grid, makeStyles, Typography } from '@material-ui/core';
+import { Grid, makeStyles, Theme, Typography } from '@material-ui/core';
 import React, { useEffect, useState } from 'react';
 import { shallowEqual, useSelector } from 'react-redux';
-import { getAlarmLimitsCurrent } from '../../../store/controller/selectors';
+import { getAlarmLimitsRequest } from '../../../store/controller/selectors';
```
ethanjli, line 13: Any particular reason why you chose to show the requested alarm limits rather than the alarm limits which the firmware reports that it's using for its alarms?

frontend/src/modules/dashboard/components/ValueInfo.tsx
```diff
@@ -294,32 +348,39 @@ const ControlValuesDisplay = ({
...
+ * @component Component for showing information based on layout configured.
  *
- * Uses the [[Props]] interface
+ * Uses the [[ValueInfoProps]] interface
  *
  * @returns {JSX.Element}
  *
  */
+// TODO: we should delete this component if it's not being used in any current code;
+// its structure is weird and its proptypes doesn't pass the linter
```
ethanjli: Is this TODO comment resolved now? It looks like you made `ValueInfo` have a reasonable structure and reasonable proptypes now.

frontend/src/modules/dashboard/views/HFNCGraphView.tsx
```diff
@@ -159,34 +159,13 @@ const HFNCMainView = (): JSX.Element => {
...
           <Grid container item justify="center" alignItems="stretch">
-            <ValueInfo
-              mainContainer={{
-                selector: getSmoothedSpO2,
-                label: 'SpO2',
-                stateKey: 'spo2',
-                units: PERCENT,
-              }}
-            />
+            <ValueInfo selector={getSmoothedSpO2} label="SpO2" stateKey="spo2" units={PERCENT} />
```
ethanjli, line 162: I really like this approach of parameterizing the `ValueInfo` component on the selector as an argument, this way we don't have to create a bunch of files representing each combination of `ValueInfo` and selector.

frontend/src/modules/dashboard/views/HFNCMainView.tsx
```diff
@@ -57,6 +61,8 @@ const useStyles = makeStyles((theme: Theme) => ({
  */
 const HFNCMainView = (): JSX.Element => {
   const classes = useStyles();
+  const spo2AlarmLimits = useSelector(getSpO2AlarmLimitsCurrent, shallowEqual);
+  const hrAlarmLimits = useSelector(getHRAlarmLimitsCurrent, shallowEqual);
```
ethanjli, line 65: Would it make sense to just provide the selector directly to `ValueInfo` the way you do for the value itself, or do we have to use `useSelector` here?

ethanjli (commented):
Comments:
- Thanks for unifying ValueInfo and LargeValueInfo!
- When we add other modes which will have groups of ValueInfo elements, we can just create a new ValueInfoGroup component which takes three ValueInfo-based containers and lays them out however things need to be specified - instead of having a "main container" and "subcontainers".
- It looks like the react components now all take selectors as their inputs, rather than (for example) `ControlValuesDisplay` taking a number as its input and `ValueInfo` doing something to `connect(...)` the selector it's given with the `ControlValuesDisplay` component, i.e. making a container. I assume you took this approach because `ControlValueDisplay` uses `ValueSelectorDisplay`, which already takes a selector and itself acts like a container. This seems like a reasonable approach to me.

Observations from testing:

- The Preact highlights behavior on the dashboard looks great!
- Performance before this branch (i.e. on the v0.12.0 tag): while the frontend is on the dashboard and I'm running the backend simulator on the RPi, the main Chromium process's CPU usage on a single core hovers around 105% to 110% most of the time, but I've seen peak usage as high as 120%. Performance on the latest commit of this branch: CPU usage hovers around 100% to 105%, though it occasionally goes just below 100% or up to 100% and it sometimes peaks around 115%. It seems like we still need to do more performance optimization which isn't related to redraws, but this PR definitely reduced CPU usage.
- After I merge develop into this branch on my RPi's local copy of the repo (i.e. to get the backend's Event Synchronization implementation, which is currently configured so that the backend doesn't send duplicate messages to the frontend as frequently) and measure CPU usage again, CPU usage on this branch does not seem to improve. We will need to do some performance profiling to identify the next bottleneck on CPU performance.

ethanjli:
For records-keeping: 

frontend/src/modules/dashboard/components/ValueInfo.tsx
rohanpurohit: So this is the misleading part i described in PR description, since I restructured `ValueInfo` and moved it to components, it thinks I changed `LargeValueInfo`, but `ValueInfo` always had `requested` alarm limits see [ValueInfo L13 on develop](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/dashboard/containers/ValueInfo.tsx#L13), this was done because when we are passing AlarmLimits from `HFNCControl` in multistep popup, initial values are not set, so we show the requested one's until they are set to some value we care. 
whereas on the Dashboard we still pass the current values through the prop [HfncMainView](https://github.com/pez-globo/pufferfish-software/pull/407/files#diff-14f0ce1c307005375c5d869583340a77eaa8a39d29141c2e3aef874dd8bad816R64)

rohanpurohit: @ethanjli so could you quickly take a look at HFNCControl in multi step popup, now it's similar to components on dashboard, I did not change that yet (go back to original padding) i feel we should? Let me know your thoughts.

frontend/src/modules/dashboard/components/ValueInfo.tsx
rohanpurohit: Yes this is resolved now.

frontend/src/modules/dashboard/views/HFNCMainView.tsx
rohanpurohit: That would mean changing the types in props, and since we currently use it in HFNCControl in MultiStepWizard too, I don't want to add another layer of complexity!

ethanjli: Ok, we'll save this for a future refactor.

rohanpurohit:
Once you test Top-level breadcrumb modal it should be good to merge @ethanjli 

ethanjli (approved):
Looks good to me in basic testing!

PR #410, opened 2021-07-07
Implement Event Synchronization protocol in frontend for early delivery of changed states

ethanjli:
This PR is a port of the backend's implementation of the Event Synchronization protocol from #403, but with some key differences:

- Because everything is driven by redux-saga, we are using ES6 Generators and regular functions for the protocol implementation. The various units of functionality (e.g. sequential sending, indexed sending, notification sending, changed state sending, etc.) have a mostly one-to-one correspondence in the concepts and semantics, and they are composed in a similar way: while the backend implementation composes classes as object members, the frontend implementation composes Generators as local variables. However, it should be feasible to implement the State and Event Synchronization protocols using closures (to provide public and private methods) instead of Generators; I am using Generators just to be consistent with the redux-saga related code.
- To avoid yielding redux-saga effects from within the State Synchronization & Event Synchronization senders (which I tried before refactoring away in fab04c2 and made the code way too complicated and difficult to understand), I make a `Map` of the state segments from the store and an `IndexedSender` adapter around it, and then I copy all relevant state segments from the store by yielding redux-saga `select` effects from the Backend state sender (which integrates the State Synchronization & Event Synchronization senders) each time I need to output anything from the senders. This is not exactly efficient, but it's efficient for us because the frontend send schedule is set to almost always output from the Event Synchronization sender, which needs to check the values of all relevant state segments from the store every single time.
- The ChangedStateSender implementation more directly checks whether a new connection has been established (by getting a last connection established timestamp as an input parameter to the Generator), because otherwise I haven't thought of a clean and correct way to pass that information in.

This PR also refactors the `saga.ts` file to move all backend protocol implementation code out of it and into `store/controller/protocols/backend/backend.ts`; then `saga.ts` becomes similar to the `ventserver.integration._trio` module in the Python backend. In the future when we reorganize the files in `store`, we may want to move some files and directories up a level, out of `store/controller`.

This PR also fixes a potential bug in the backend's `ChangedStateSender` which was masked due to the way we initialize states in the backend and our `output_idle` configuration in the backend: if a connection was just established but the backend had `None` as one of its sendable states and that state always remained at `None`, then the `NotificationSender` would fall back to the idle sender. This bug appeared in the frontend as a breaking bug before I fixed it there: the frontend has no "idle sender" functionality, and so it would simply get stuck and stop outputting anything sometime after a connection was jut established. Instead, upon establishment of a new connection the backend and frontend's `ChangedStateSender` first should confirm that each state it wants to mark as sendable is not `None`/`null`, before instructing the `NotificationSender` to mark that state as sendable.


ethanjli:
Performance measurement data (collected on an RPi running the simulator backend):

- On the base `develop` branch (d71f11a), single-core CPU usage of the most CPU-intensive chromium process ranges from 100 % - 105% most of the time, occasionally reaching 110-120%. The python process stays aroudn 100% single-core CPu usage.
- On cca642b, chromium single-core CPU usage ranges from 100-105% most of the time, occasionally reaching 110-120%. The python process stays around 100% single-core CPU usage.

Thus, this PR doesn't seem to cause a major performance change in either direction.

ethanjli:
Performance measurement data (collected on an RPi running the simulator backend):

- On fc221cc, chromium single-core CPU usage ranges from 100-105% most of the time, occasionally reaching 110-120%. The python process stays around 100% single-core CPU usage.

Thus, the rewrite in 2bbd755 doesn't seem to cause a major performance change in either direction.

ethanjli:
@rohanpurohit Could you review this by testing the frontend running together with the application backend and the firmware and looking for any behavior/performance regressions, as well as latency differences (if any) in how quickly frontend user inputs (e.g. mute/unmute) are handled?

rohanpurohit:
observations from testing (backend running application, MCU on develop)
- compared to the performance on develop branch (https://github.com/pez-globo/pufferfish-software/commit/d71f11acfe92357142e66592ae671c153d40aa78) I feel like there is a visible `performance drop.
`

Consider inputs like:
- mute/unmute (compared to develop, the response is delayed in terms of changing of the red border and other visual cues), I'm almost waiting for it to change, whereas on develop I can see myself changing continuously without a delay.
- transition between start/pause ventilation (dashboard to quickstart), try starting and pausing ventilation multiple times, seems a lot snappier on develop, in terms of the page rendering completely before we can make further inputs.
- open `events log` modal, continuously change between active alarms and events log by pressing the button on top. it is a lot faster on develop, the pages switching itself, and even the button changing text, whereas here I see a significant delay on this branch.
- the backend killed the frontend multiple times, felt like it happened more when I changed too many things at a time, as in a lot of user inputs (if that makes sense) not sure if you faced this on your side. let me know and I could test again to confirm.
I did wait for the frontend to start before i started the backend (I remember facing this once) but everything is fine on develop so I'm not really sure.

ethanjli:
I hadn't tested the responsiveness/latency when trying to interact with the frontend, and in fact I was mostly testing from the landing page (while looking at `console.log` messages emitted from my saga code). Here's what I see on d71f11a when trying to interact with the frontend while it talks to the application backend:
 - There are noticeable delays between when I press unmute/unmute and when the red border changes. In a random sample, there's a latency of [1.3, 1.8, 1.3, 1.5, 1.8] sec between when I press the button and when the red border changes.
 - There are long delays between when I press Start/Pause (ventilation) and when the screen finishes redrawing everything (e.g. the red border disappears momentarily and then comes back when I Pause ventilation). In a random sample, there's a latency of [2.1, 2.5, 2.5, 3.1, 2] sec.
 - There are short but noticeable delays between when I press the Events Log button from the Active Alarms modal and when the modal itself finishes redrawing the table. In a random sample, there's a latency of [1.1, 1, 1.2, 1.1, 1.1] sec, with ~35-40 events in the events log. There's a much shorter delay between when I press the Active Alarms button from the Events Log modal and when the modal itself finishes redrawing the table (with 3-5 events in it).
 - When I quickly and repeatedly switch between the "HFNC Control" tab and the "SpO2 Alarms" tab in the breadcrumbs modal, the backend never kills the frontend.

And here's what I see on fc221cc when testing with the application backend:
 - I also see noticeably longer delays between when I press unmute/unmute and when the red border changes. In a random sample, there's a latency of [2.2, 2.3, 1.6, 1.8, 1.4, 1.9] sec between when I press the button and when the red border changes.
 - The first time I launched the frontend and went to the quickstart screen, the backend killed the frontend; right now the backend has a heartbeat timeout of 2 sec on messages from the frontend, while the frontend is configured to send a non-null state at least once every 500 ms, and to check all selectors for state changes every 20 ms. I also saw this happen when I opened the Events Log for the first time. This problem made it impossible for me to collect more latency measurements. So at a high level I'm able to reproduce your findings.

So in 69fcd5f now I've changed the frontend to send a non-null state at least once every 100 ms, and to check all selectors for state changes every 50 ms (compare to the previous configuration to just send the next state every 50 ms). Here's what I see on that commit when trying to interact with the frontend while it talks to the application backend:
 - There are still noticeably worse delays between when I press unmute/unmute and when the red border changes. In a random sample, there's a latency of [1.5, 1.5, 1.9, 2.1, 2.7] sec between when I press the button and when the red border changes.
 - There are similar or shorter delays between when I press Start/Pause (ventilation) and when the screen finishes redrawing everything (e.g. the red border disappears momentarily and then comes back when I Pause ventilation). In a random sample, there's a latency of [2.2, 1.8, 1.8, 1.8, 2.2] sec.
 - There are similar delays between when I press the Events Log button from the Active Alarms modal and when the modal itself finishes redrawing the table. In a random sample, there's a latency of [1, 1.3, 1.2, 1, 0.9] sec, with ~35-40 events in the events log.
 - When I quickly and repeatedly switch between the "HFNC Control" tab and the "SpO2 Alarms" tab in the breadcrumbs modal, the backend never kills the frontend. I haven't seen the backend kill the frontend in the conditions which occurred when I tested the frontend in fc221cc.

@rohanpurohit Can you try to repeat these measurements on 69fcd5f and report what you get? I will also be curious to see how develop branch performance compares to this PR once we merge #412 into develop and into this branch, as that action flooding appears to be eating up CPU all the time, which might have unintuitive interactions with the performance effects of this PR. If we can't get performance to be acceptable, I may need to dig deeper into how I can use shallow/reference equality (`===`) on objects returned by my `select(selector)` redux-saga calls - i.e. taking advantage of the same performance optimization used by react-redux and reselect. Right now for some reason the objects returned by those selectors are never shallowly equal.

rohanpurohit:
Observations from testing on RPI running application backend in all cases:

### Scenario 1: on the develop branch https://github.com/pez-globo/pufferfish-software/commit/d71f11acfe92357142e66592ae671c153d40aa78

- When I press mute/unmute button and when the red border changes, In a random sample, there is a latency of 
[1.35, 1.84, 2.02, 1.26, 1.68, 1.55, 1.43. 1.32, 1.48, 1.07] sec between when I press the button and when the red border changes
- When I press the start/pause ventilation button, In a random sample, there is a latency of [1.50, 1.69, 1.99, 1.74, 1.14, 1.17, 2.06, 1.69, 1.95, 1.30] sec between when I press the button and everything is redrawn.
- On pressing the button in events log modal to switch between active alarms screen and events log modal with events log of around 80, In a random sample, there is a latency of  [1.04, 1.23, 1.02, 0.87, 1.07, 1.02, 0.61, 0.77. 0.68, 1.01, 1.04]


### Scenario 2: on commit https://github.com/pez-globo/pufferfish-software/commit/69fcd5fa1f91064907129750568352eedc155714 

- When I press mute/unmute button and when the red border changes, In a random sample, there is a latency of 
[1.52, 2.34, 1.31, 1.83, 1.27, 1.46, 1.83, 1.26, 1.39, 1.80] sec between when I press the button and when the red border changes
- When I press the start/pause ventilation button, In a random sample, there is a latency of [2.73, 2.31, 2.85, 2.50, 2.47, 1.65, 2.33, 1.43, 1.66, 1.43] sec between when I press the button and everything is redrawn.
- On pressing the button in events log modal to switch between active alarms screen and events log modal with events log of around 80, In a random sample, there is a latency of  [0.95, 0.84, 0.92, 0.85, 0.68, 0.75, 0.90, 0.82, 0.83, 0.90]
- The backend does not kill the frontend anymore in any scenario.


### Scenario 3: on branch frontend-event-sync merged with #412 

- When I press mute/unmute button and when the red border changes, In a random sample, there is a latency of 
[1.25, 1.08, 1.07, 1.53, 1.05, 0.85, 1.19, 0.71, 1.40, 1.08] sec between when I press the button and when the red border changes
- When I press the start/pause ventilation button, In a random sample, there is a latency of [1.28, 1.37, 1.85, 1.58, 2.14, 1.53, 3.15, 1.46, 1.51, 1.52] sec between when I press the button and everything is redrawn.
- On pressing the button in events log modal to switch between active alarms screen and events log modal with events log of around 80, In a random sample, there is a latency of  [0.66, 1.25, 0.88, 0.90, 0.81, 0.78, 0.79, 0.68, 0.80, 0.84]


### In conclusion:
average times for all 3 scenarios:

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/Rohan/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/Rohan/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#0563C1" vlink="#954F72">



Scenario | mute/unmute | start/pause redraw | events log modal
-- | -- | -- | --
on develop | 1.266s | 1.623s | 1.036s
on 69fcd5f | 1.601s | 2.136s | 0.844s
frontend-event-sync merged   with #412 | 1.121s | 1.739s | 0.839s



</body>

</html>

@ethanjli Therefore I feel there isn't a significant delay anymore on https://github.com/pez-globo/pufferfish-software/commit/69fcd5fa1f91064907129750568352eedc155714.  Let me know what you feel about the findings. thanks!

frontend/src/store/app/selectors.ts
```diff
@@ -40,6 +40,10 @@ export const getBackendConnected = createSelector(
   getApp,
   (app: AppState): boolean => app.backendConnection,
 );
+export const getLastBackendConnectionTime = createSelector(
```
rohanpurohit, line 43: I haven't gone through the entire PR, but I'm thinking maybe we can use this in the `useEffect` in `OverlayScreen` or if we replace that with the saga in the future to limit the dispatch of `BACKEND_CONNECTION_UP` action?

ethanjli: That's an interesting idea! `lastBackendConnectionTime` changes exactly when a `BACKEND_CONNECTION_UP` action changes `backendConnection` from false to true, so whether it limits dispatch of the `BACKEND_CONNECTION_UP` action will depend on the structure of our conditionals in that `useEffect`.

ethanjli:
Qualitatively comparing your measurements with mine, it seems that we get similar latencies on d71f11a (which was a commit on develop before #412), except you get slightly shorter latencies on the start/pause ventilation button. It looks like we get similar latencies on 69fcd5f (which is the most recent commit on this PR), except you get slightly shorter latencies on the active alarms/events log modal switching. The similarity of our measurements is good news.

Now that we've merged #412 into develop, here are my latency measurements of the frontend running on the RPi with the simulator backend, after both have been running for a while (> 5 minutes):

On develop after #412 (8e804ac51fdee2dc251e6ec9c1a83f2bff39eb1c):
- Latency between pressing the mute/unmute alarms button and seeing a change in the red border: [1.8, 1.7, 1.9, 1.6, 1.8, 1.4, 2.0, 1.3, 2.1, 1.3] sec.
- Latency between pressing the start/pause ventilation button and seeing a completion in the redraw of everything (including, on the quickstart screen while a power alarm is active, a blink of the red border): [2.0, 1.6, 2.0, 2.2, 2.0, 2.3, 2.1, 1.9, 2.4, 1.9] sec.
- Latency between pressing the button to toggle show events log modal from the active alarms modal and seeing the table be fully redrawn: [1.1, 0.8, 1.1, 0.8, 1.1, 0.9, 1.0, 1.0, 0.9, 1.2] sec., with 35-40 items in the event log

These latencies look significantly longer than the measurements for d71f11a (i.e. develop before #412), but probably the test conditions are different in a meaningful way, e.g. the temperature in my room, running the simulator vs. application backend, etc. CPU usage of the chromium process seems to stay between 75% and 90% most of the time, with rare spikes to 110% and occasional jumps to 100%. For now maybe we'll need to be careful about doing paired tests where when we compare two conditions we have to test one condition immediately after the other.

On this PR after #412 (93bd122):
- Latency between pressing the mute/unmute alarms button and seeing a change in the red border: [1.3, 1.9, 2.0, 1.7, 1.4, 2.0, 1.3, 1.6, 1.6, 1.8] sec. This seems pretty similar to 83804ac.
- Latency between pressing the start/pause ventilation button and seeing a completion in the redraw of everything: [2.5, 1.8, 2.0, 1.7, 2.3, 1.7, 2.4, 1.8, 2.4, 2.0] sec. This seems pretty similar to 8e804ac, though here it's clearer that there's some other problem with the quickstart screen which makes the red border blink and thus makes it take longer to finish redrawing.
- Latency between pressing the button to toggle show events log modal from the active alarms modal and seeing the table be fully redrawn: [1.1, 0.8, 1.2, 1.2, 0.9, 1.0, 1.1, 1.0, 0.8, 0.7] sec., with 35-40 items in the event log. This seems to be basically the same latency as in 8e804ac.

So from eyeballing my quantitative measurements, it seems to me that this PR does not significantly improve or worsen the latency vs. develop after #412; but my measurements seem pretty different from yours. Can you try to compare 8e804ac51fdee2dc251e6ec9c1a83f2bff39eb1c and 93bd122 to report if it feels like there's a qualitative difference in responsiveness/latency? If it doesn't feel less responsive, maybe we can merge this PR into either v0.13 or v0.14, and then in future PRs we can keep working on frontend performance optimizations. If future profiling reports a bottleneck in the event synchronization Generator function, then we can look into the shallow/reference equality thing as a potential bottleneck to optimize.

rohanpurohit (approved):
I compared https://github.com/pez-globo/pufferfish-software/commit/8e804ac51fdee2dc251e6ec9c1a83f2bff39eb1c and https://github.com/pez-globo/pufferfish-software/commit/93bd1225d83caa3e647b41293afcb035c7d20bd0, and I feel there isn't a significant drop or gain in performance so definitely `it does not feel less responsive`. I believe we can move this to 0.13, and keep improving performance on the frontend in future PR's.

Issue #411, opened 2021-07-07
Redux is spammed with `BACKEND_CONNECTION_UP` actions


ethanjli:
Right now the `HeartbeatBackendListener` component in `OverlayScreen.tsx` dispatches an `establishedBackendConnection` action on every single message the frontend receives from the backend. With the current messaging settings, that means that the frontend is usually having to process an extra (redundant) action up to once every 10 ms. This is probably not the main performance bottleneck in our frontend, but it adds a ton of extra events into the Redux inspector (which makes it harder to catch other actions). The fix should be relatively straightforward to implement: only dispatch the `establishedBackendConnection` action if the backend is not already connected (as reported by the `getBackendConnected` selector).

It might be worth getting CPU usage measurements before and after this is fixed, just to get a sense of whether this issue has any measurable impact on CPU usage.

ethanjli:
There also seems to be some kind of correctness problem with the triggering of `establishedBackendConnection` action dispatches. When I modify the backend's frontend send schedule to only send a state to the frontend once every 5 seconds, and I uncomment the backend's `print` code to print every single event sent by the backend to the frontend, I see that there's a flood of `BACKEND_CONNECTION_UP` actions (spaced ~5 ms apart) triggered after every single event sent by the backend. There should be no more than one such action triggered after every event from the backend. What I see now reminds me of the behavior we saw in #363. @rohanpurohit Can you see if you can reproduce those test conditions and this observation on your end? I first observed this on 2bbd755, but I would expect that it should also be reproducible on the develop branch (e.g. d71f11a).

Given the timing of this flood of actions, it seems plausible to me that this could be making the frontend's CPU usage measurably higher than it's supposed to be, so we should definitely measure CPU usage before and after this is fixed. This flooding problem should be fixed _before_ we suppress `establishedBackendConnection` dispatching when the backend is already connected (as specified in the main issue description for #411), as that suppression will make it harder to verify the correctness of the fix to this flooding problem.

rohanpurohit:
[...]

PR #412, opened 2021-07-08
Fix width of alert button in the toolbar and dispatch of establishedBackendConnection

### LINKED ISSUE #387: Frontend topbar needs to be made more compact

rohanpurohit:
This PR fixes #387 and #411 
- I'm setting a larger width so that slightly longer text doesn't wrap to a new line which makes the button change its height

### DIFF
--- a/frontend/src/modules/app/EventAlerts.tsx
+++ b/frontend/src/modules/app/EventAlerts.tsx
@@ -45,10 +45,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     justifyContent: 'space-between',
     alignItems: 'center',
   },
-  alertMargin: {
-    marginLeft: theme.spacing(0.5),
-    marginRight: theme.spacing(0.5),
-  },
   marginRight: {
     marginRight: theme.spacing(1),
   },
@@ -96,7 +92,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
   },
   alertButtonSpan: {
-    width: '250px',
+    width: '270px',
     borderRadius: 5,
     marginRight: '10px',
     backgroundColor: '#FF0000',
@@ -106,8 +102,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
   },
   alertButtonSpanMuted: {
-    justifyContent: 'flex-end',
-    width: '250px',
+    width: '270px',
     borderRadius: 5,
     marginRight: '10px',
     backgroundColor: '#FF0000',
@@ -124,11 +119,13 @@ const useStyles = makeStyles((theme: Theme) => ({
     padding: '6px 10px',
   },
   timer: {
-    padding: '6px 16px',
+    justifyContent: 'flex-end',
+    padding: '6px 20px',
     backgroundColor: '#88211b',
     borderRadius: '0px 4px 4px 0px',
   },
   alertTimer: {
+    width: '270px',
     margin: '0px 12px',
     backgroundColor: '#0053b336',
     '&:hover': {
@@ -301,7 +298,7 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
       setIsMuted(!alarmMuteRequestActive);
       // Start the timer
       const timer = setTimeout(() => {
-        setRemaining(remaining - 1);
+        setRemaining(() => remaining - 1);
         if (remaining <= 0) {
           clearTimeout(timer);
         }
@@ -436,9 +433,7 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
           className={isMuted ? classes.alertButtonSpan : classes.alertButtonSpanMuted}
           onClick={onActiveAlarmClick}
         >
-          <span className={isMuted ? `${classes.alertMargin}` : ''} style={{ padding: '6px 16px' }}>
-            {alert.label}
-          </span>
+          <span style={{ padding: '6px 16px', width: 200 }}>{alert.label}</span>
           {alertCount > 1 && (
             <div
               className={classes.iconBadge}
@@ -448,7 +443,7 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
             </div>
           )}
           {!isMuted && countdownTimer !== undefined && (
-            <div className={classes.timer} style={{ marginLeft: 10 }}>
+            <div className={classes.timer} style={{ right: 'auto' }}>
               {new Date(countdownTimer * 1000).toISOString().substr(14, 5)}
             </div>
           )}
--- a/frontend/src/modules/app/EventAlerts.tsx
+++ b/frontend/src/modules/app/EventAlerts.tsx
@@ -92,17 +92,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
   },
   alertButtonSpan: {
-    width: '270px',
-    borderRadius: 5,
-    marginRight: '10px',
-    backgroundColor: '#FF0000',
-    color: '#fff',
-    '&:hover': {
-      backgroundColor: '#FF0000',
-    },
-  },
-  alertButtonSpanMuted: {
-    width: '270px',
+    width: '300px',
     borderRadius: 5,
     marginRight: '10px',
     backgroundColor: '#FF0000',
@@ -125,7 +115,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     borderRadius: '0px 4px 4px 0px',
   },
   alertTimer: {
-    width: '270px',
+    width: '300px',
     margin: '0px 12px',
     backgroundColor: '#0053b336',
     '&:hover': {
@@ -430,10 +420,10 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
           style={{ marginLeft: 10, marginRight: 10, margin: '0px 10px', padding: 0 }}
           variant="contained"
           color="primary"
-          className={isMuted ? classes.alertButtonSpan : classes.alertButtonSpanMuted}
+          className={classes.alertButtonSpan}
           onClick={onActiveAlarmClick}
         >
-          <span style={{ padding: '6px 16px', width: 200 }}>{alert.label}</span>
+          <span style={{ padding: '6px 16px', width: 250 }}>{alert.label}</span>
           {alertCount > 1 && (
             <div
               className={classes.iconBadge}
@@ -451,7 +441,7 @@ export const EventAlerts = ({ label }: Props): JSX.Element => {
       </Grid>
       <Grid hidden={alertCount > 0}>
         <Button
-          style={{ marginLeft: 5 }}
+          style={{ marginLeft: 10, marginRight: 10 }}
           variant="contained"
           color="primary"
           className={classes.alertButton}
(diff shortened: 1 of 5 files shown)


rohanpurohit:
Perfomance testing on RPI (running the application backend)
- single-core CPU usage for the chrome process ranges from 75-88%, averaging around 84, never reaching 100% or beyond
- The python process ranges from 93-98 averages around 94, always below 100%

Alert Button:
- Definitely looked better on chrome with the resolution of 1280x800, although on RPI it's below the mute button, not sure if we want that ( edit: saw that it's the same on develop), maybe we need to reduce the width further?
- since we want the `timer` at the end of the button, I used `justifyContent: 'flex-end'` for the button CSS, mostly looked fine on chrome on my local pc (1280x800 resolution) except for labels like `Low Battery` since the text is too short, spacing is not symmetrical (more space on the left side) because of the `flex-end` attribute. maybe there is a better way to do this?

Should we consider the visual aspects of the UI on RPI (I'm using vnc client with a different resolution maybe that's a factor too?) or simply the resolution of `1280x800` on chrome should be enough?

ethanjli:
Correctness regression related to #411 (in the future, PRs like this consisting of very small entirely separate fixes should be split into smaller hotfix PRs so that we don't have to wait for all of them to be fully ready before merging any one of them, and so that we can have more descriptive branch names):

- Now when I kill the backend, the frontend fails to detect that the backend connection is lost. This is because of where you placed your `if (backendConnected) { return; }`.
- Since the `useEffect` is now depending on `backendConnected`, which is also modified by the `BACKEND_CONNECTION_DOWN` and `BACKEND_CONNECTION_UP` actions, we should be careful to guard against any circular dependencies which might be coincidentally masked by the code right now - at a minimum, there should be an inline code comment explaining the ways in which a circular dependency may be introduced if someone isn't careful in the future. I would also suggest still checking the potential impact of the changes I suggested in https://github.com/pez-globo/pufferfish-software/issues/411#issuecomment-876695814 (you submitted this PR before I made that comment).

Frontend layout:

- The alarm message display box has a different width when there's no active alarm vs. when there's an active alarm; it should instead have the same width in both conditions.
- The centering of the alarm message does look weird when there's an alarm mute countdown. It sounds like `flex-end` is not the best tool for this. The approach I initially had in mind was to manually change the width of the span which contains the `{alert.label}` to one of two constant values, depending on `isMuted`; and then we should set a fixed width for the `alertTimer` div. If you tried this, what ended up happening?
- I'm not sure what you mean by "although on RPI it's below the mute button, not sure if we want that ( edit: saw that it's the same on develop), maybe we need to reduce the width further?". Here's what I see on my RPi:



CPU performance measurement:

- For some reason I'm always getting that the python process (either simulator or application) averages 100% CPU, occasionally going up to 102%.
- The chromium process CPU usage has improved significantly for me: now it averages around 89%, and peaks at 100% (with very rare peaks at 112%).

rohanpurohit:
> * Since the `useEffect` is now depending on `backendConnected`, which is also modified by the `BACKEND_CONNECTION_DOWN` and `BACKEND_CONNECTION_UP` actions, we should be careful to guard against any circular dependencies which might be coincidentally masked by the code right now - at a minimum, there should be an inline code comment explaining the ways in which a circular dependency may be introduced if someone isn't careful in the future. I would also suggest still checking the potential impact of the changes I suggested in [Redux is spammed with `BACKEND_CONNECTION_UP` actions #411 (comment)](https://github.com/pez-globo/pufferfish-software/issues/411#issuecomment-876695814) (you submitted this PR before I made that comment).

Yes, I'm really concerned about using this, is there a way we could only dispatch the action once without relying on `backendConnected`?

rohanpurohit:
Latest changes:

useEffect:
- The latest changes are not final and I'm just trying to get a sense of the performance implications of this, what I'm doing right now will certainly introduce a delay in the detection of `connection loss`

actions:
- removed the reducer for both backend connection up and down updating the `clock`

alert button:
- fixed width of alarm message display box in all scenarios.
- made some improvements also increased the width to accommodate 'software connectivity lost' message, anything lower than this makes the button change its height (as it wraps the text in a new line).

Perfomance testing on RPI (running application backend)
- Python process is more or less the same averaging around 95%
- chromium process:
1. when idle on dashboard ranges around 64-70% with momentary spikes of up to 85% and more
2. while transitioning between pages and user inputs peaks around 89% and more in some cases.

ethanjli:
In light testing, the visual changes look good to me! I think they're ready to merge. I haven't tested anything related to the useEffect, since you plan to make more changes.

> Yes, I'm really concerned about using this, is there a way we could only dispatch the action once without relying on `backendConnected`?

I can't think of any way to do this. Maybe the best we can do is to figure out a formal/informal proof/argument about the dependency relationships around `backendConnected`. As a start, we know that the `useEffect` will run any time `backendConnected` changes. If we can establish a logical relationship between the latest state of `backendConnected`, the occurrence of changes to its state, and the occurrence of `BACKEND_CONNECTION_UP` and `BACKEND_CONNECTION_DOWN` action dispatches, that would be a formal specification we can look at to see whether our design could cause a circular dependency under certain edge cases. Since this logic is mostly about conditionals, maybe you could try to put together a decision table specifying the logical relationship between the input dependencies of the useEffect and the resulting action dispatches?

rohanpurohit:
> As a start, we know that the `useEffect` will run any time `backendConnected` changes. 

It is clearly a circular dependency as the `useEffect` itself is dispatching the actions that change `backendConnected`!

> If we can establish a logical relationship between the latest state of `backendConnected`, the occurrence of changes to its state, and the occurrence of `BACKEND_CONNECTION_UP` and `BACKEND_CONNECTION_DOWN` action dispatches, that would be a formal specification we can look at to see whether our design could cause a circular dependency under certain edge cases. Since this logic is mostly about conditionals, maybe you could try to put together a decision table specifying the logical relationship between the input dependencies of the useEffect and the resulting action dispatches?

Hmm seems complicated to come up with a table, I'll do my best!

> I haven't tested anything related to the useEffect, since you plan to make more changes.

what are your thoughts about the current implementation? since we don't really want to keep `getClock` I'm thinking either we could defer it for later or come up with something that's acceptable as prototype level code

ethanjli:
> It is clearly a circular dependency as the `useEffect` itself is dispatching the actions that change `backendConnected`!

Right, at the level of syntax there has to be a circular dependency because the useEffect under certain circumstances will dispatch actions which change `backendConnected`. We just have to make sure that our conditionals will break any potential positive feedback loops where dispatching an action causes the useEffect to be run more than once as a result. Negative feedback loops are fine - i.e. the circular dependency is used to suppress the dispatching of actions in a dispatch action -> re-run useEffect -> dispatch action -> re-run useEffect -> ... loop. My intuition is that once we write out the decision table we'll be able to prove that we don't have any positive feedback loops.

> what are your thoughts about the current implementation? since we don't really want to keep `getClock` I'm thinking either we could defer it for later or come up with something that's acceptable as prototype level code

It looks good to me - if you're happy with it, I'll try running it on my RPi tomorrow morning (go ahead and add your responses to the 3 questions for records-keeping and that way I can approve immediately if everything looks good to me in testing). You could probably refactor:
```
    } else {
      if (backendConnected) return;
```
as:

```
    } else if (!backendConnected) {
```

rohanpurohit:
> > It is clearly a circular dependency as the `useEffect` itself is dispatching the actions that change `backendConnected`!
> 
> Right, at the level of syntax there has to be a circular dependency because the useEffect under certain circumstances will dispatch actions which change `backendConnected`. We just have to make sure that our conditionals will break any potential positive feedback loops where dispatching an action causes the useEffect to be run more than once as a result. Negative feedback loops are fine - i.e. the circular dependency is used to suppress the dispatching of actions in a dispatch action -> re-run useEffect -> dispatch action -> re-run useEffect -> ... loop. My intuition is that once we write out the decision table we'll be able to prove that we don't have any positive feedback loops.
> 

Oh okay great! so is the PR blocked by decision table or we can write it out later?
> > what are your thoughts about the current implementation? since we don't really want to keep `getClock` I'm thinking either we could defer it for later or come up with something that's acceptable as prototype level code
> 
> It looks good to me - if you're happy with it, I'll try running it on my RPi tomorrow morning. You could probably refactor:

so there is a clear performance hit when we are using `getClock`!
performance on https://github.com/pez-globo/pufferfish-software/pull/412/commits/224fcc13a326a8aaa7010c8e15d3ff2558b0e683:
- [#412 comment](https://github.com/pez-globo/pufferfish-software/pull/412#issuecomment-877581029)

performance on https://github.com/pez-globo/pufferfish-software/pull/412/commits/9c38c6aaf2da827b9fa093166170a4fd31c5564f:
- mirrors what we tested initially [comment](https://github.com/pez-globo/pufferfish-software/pull/412#issuecomment-876578511) & [comment](https://github.com/pez-globo/pufferfish-software/pull/412#issuecomment-876843486)

for records-keeping.

ethanjli (approved):
> Oh okay great! so is the PR blocked by decision table or we can write it out later?

Since now it seems clear that for #415 we should remove the `useEffect` and replace it with a saga, I guess we ultimately won't actually need a decision table proving that `useEffect` won't cause a positive feedback loop in the dispatching of an action. As long as we empirically don't observe action flooding in this PR, I think it's good enough to merge as a temporary fix.

> so there is a clear performance hit when we are using `getClock`!

I measured the CPU usage on a6c3513 (running on the RPi together with the application backend). Initially, the backend averages around 90-95% CPU usage, while Chromium averages around 80-90% CPU (occasionally peaking at 100%). After running both processes for a while, the backend averages 100% CPU usage (plus or minus 3% most of the time), while Chromium averages around 85-90% CPU usage, occasionally peaking at 105%. There's probably some thermal throttling happening, so we should make sure that we only measure CPU usage after the RPi has been running a frontend and backend together for a while.

This is a major improvement over develop, and (from looking at the redux debug console) it fixes #411 . So let's merge this in for v0.13, and we'll do another iteration for v0.14.

My guess is that that the performance hit you saw is not directly from `getClock`, but rather from the fact that when we add `getClock` back then we are actually running the `useEffect` and thus dispatching a bunch of `BACKEND_HEARTBEAT` actions (in fact, once every 10 ms), which is probably eating up a lot of CPU. My intuition is that the refactor for #415 may also remove the `BACKEND_HEARTBEAT` action, since it would be simpler to just track that as a local variable in the saga.

PR #413, opened 2021-07-08
Refactor frontend modules/display and screensaver

### LINKED ISSUE #399: Refactor code in modules/displays and screensaver

### modules/display:

- Deciding for ourselves what belongs in modules/displays, and moving files in and out as needed: #359 (comment)
- Turning bannerType into an Enum: #359 (comment)
- Renaming the Data type internal to the MultiStepWizard to something more descriptive: #359 (comment)
- Replacing committedSetting with setValueActual, if possible: #359 (comment)
- Changing the toggleBetween field from an array of an unnamed object type to an array of ToggleSetting objects, where the ----ToggleSetting is defined as an interface with label and value fields: #359 (comment)
- Renaming ValueDisplay and ValueSelectorDisplay to something more descriptive/specific: #359 (comment) and #359 (comment)

### moduels/screensaver:

- Rename "Control Wrapper" to something more descriptive.

rohanpurohit:
This PR includes:

- Turning bannerType into an Enum
- Renaming the Data type internal to the MultiStepWizard to `InternalState`
- Renaming setValue and setValueActual to `ValueDraft` and `ValueCurrent` respectively 
- using mcu_pb interface `Range` instead of `number[]`
- getting rid of functions in MultiStepWizard accessing store directly as they weren't react functional components, now selectors and hooks are used instead.
- `MultiStepWizard`, `ModalPopup`, `DiscardAlarmLimitsContent` and `EventAlerts` are now moved to `modals` folder.
- `ToogleValue` is now moved to `controllers` as its an interactive component
- DashBoardPage is cleaned up as we only have `HFNCMainView` for now.
- `ValuesPage` and corresponding `info` components are removed (reason: they are deprecated, and restrict us from re-structuring other components like ValueModal and ValueDisplay)
- alarmLimitsSelector is updated to return `Range` instead of `number[]`


frontend/src/modules/controllers/AlarmModal.tsx
```diff
@@ -124,7 +124,7 @@ export const AlarmModal = ({
   step,
   contentOnly = false,
   labelHeading = false,
-  alarmRangeValues = [],
+  alarmRangeValues = { lower: 0, upper: 0 },
```
ethanjli: Maybe it would make sense to initialize `alarmRangeValues` as `null`?

ethanjli (commented):
Here are my comments from looking at code changes. I will report results of light testing separately, after changes in this review and the next one which I've just posted below.

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -40,36 +42,36 @@ import store from '../../store';
...
   stateKey: string;
   label: string;
   units: string;
   isAlarmEnabled: boolean;
   isSetvalEnabled: boolean;
   committedSetting?: number | null;
-  alarmValues: number[];
-  setValue: number;
+  alarmValues: Range;
+  ValueDraft: number;
```
ethanjli: Shouldn't the `V` be lowercase here?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -40,36 +42,36 @@ import store from '../../store';
...
   minValue?: number | null;
   maxValue?: number | null;
   alarmLimitMin?: number | null;
   alarmLimitMax?: number | null;
-  alarmValuesActual: number[];
-  setValueActual: number;
+  alarmValuesActual: Range;
+  ValueCurrent: number;
```
ethanjli: Similarly, the `V` should be lowercase here too, right?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -40,36 +42,36 @@ import store from '../../store';
...
-  setValue: number;
+  alarmValues: Range;
+  ValueDraft: number;
   minValue?: number | null;
   maxValue?: number | null;
   alarmLimitMin?: number | null;
   alarmLimitMax?: number | null;
-  alarmValuesActual: number[];
-  setValueActual: number;
+  alarmValuesActual: Range;
```
ethanjli: Should this just be `alarmLimitsCurrent`?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -40,36 +42,36 @@ import store from '../../store';
...
+interface InternalState {
   stateKey: string;
   label: string;
   units: string;
   isAlarmEnabled: boolean;
   isSetvalEnabled: boolean;
   committedSetting?: number | null;
-  alarmValues: number[];
-  setValue: number;
+  alarmValues: Range;
```
ethanjli: Should this actually be `alarmLimitsDraft`? Or does it serve a slightly different purpose?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -405,16 +294,67 @@ const MultiStepWizard = (): JSX.Element => {
...
    */
-  const [multiParams, setMultiParams] = React.useState<Data[]>([]);
+  const [multiParams, setMultiParams] = React.useState<InternalState[]>([]);
   /**
    * State to manage if button submitted is disabled
    */
   const [isSubmitDisabled, setIsSubmitDisabled] = React.useState(false);
 
+  const alarmValuesSpO2 = useSelector(getSpO2AlarmLimitsRequest, shallowEqual);
+  const alarmValuesHR = useSelector(getHRAlarmLimitsRequest, shallowEqual);
```
ethanjli: Whenever this file calls something `alarmValues`, does it actually mean `alarmLimits`?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -405,16 +294,67 @@ const MultiStepWizard = (): JSX.Element => {
...
+        ],
+        [
+          'fio2',
+          createInternalState(
+            'FiO2',
+            'fio2',
+            PERCENT,
+            true,
+            false,
+            { lower: 0, upper: 0 },
```
ethanjli: Would it be more appropriate to have `null` be the initial/default value here, instead of `{ lower: 0, upper: 0 }` which could be an actual combination of alarm limits values specifiable by the user?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -484,9 +424,11 @@ const MultiStepWizard = (): JSX.Element => {
     if (parameter) {
       setTabIndex(1);
       if (!multiParams.length) {
-        setMultiParams([parameter as Data]);
+        setMultiParams([parameter as InternalState]);
```
ethanjli: Is `multiParams`/`setMultiParams` just `internalState`, or is it also holding additional data?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -502,11 +444,13 @@ const MultiStepWizard = (): JSX.Element => {
    *
    * Updates the new Set value into multiParams `Data` object
    */
-  const doSetValue = (setting: number) => {
+  const doValueDraft = (setting: number) => {
```
ethanjli: What does "do value draft" mean?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -529,9 +473,11 @@ const MultiStepWizard = (): JSX.Element => {
    */
   const doSetAlarmValues = (min: number, max: number) => {
```
ethanjli: Can we rename this function? I assume `setAlarmValues` was already taken, but maybe we can find a clearer name.

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -549,11 +495,11 @@ const MultiStepWizard = (): JSX.Element => {
...
    */
   const getAlarmValues = (stateKey: string) => {
-    const param = multiParams.find((param: Data) => param.stateKey === stateKey);
+    const param = multiParams.find((param: InternalState) => param.stateKey === stateKey);
     if (param) {
-      if (param.alarmValues.length) return param.alarmValues;
+      if (param.alarmValues) return param.alarmValues;
     }
-    return [];
+    return { lower: 0, upper: 0 };
```
ethanjli: I think this is another example where `null` is better as a sentinel value indicating that alarm limits are missing.

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -827,20 +768,19 @@ const MultiStepWizard = (): JSX.Element => {
...
                 } else if (
-                  param.alarmValues.length &&
-                  (param.alarmValues[0] !== param.alarmValuesActual[0] ||
-                    param.alarmValues[1] !== param.alarmValuesActual[1])
+                  param.alarmValues.lower !== param.alarmValuesActual.lower ||
+                  param.alarmValues.upper !== param.alarmValuesActual.upper
                 ) {
                   return (
-                    <Typography variant="subtitle1">{`Change ${param.label} alarm range to ${param.alarmValues[0]} - ${param.alarmValues[1]}?`}</Typography>
+                    <Typography variant="subtitle1">{`Change ${param.label} alarm range to ${param.alarmValues.lower} - ${param.alarmValues.upper}?`}</Typography>
```
ethanjli: While we're at it, let's change `alarm range` to `alarm limits` in this text, and let's use the same format for displaying the range here as what we use in the event log (it involves square brackets, and it should include units). It would be great if we could display the old range, too: e.g. Change HR alarm limits from [0 BPM - 200 BPM] to [50 BPM - 100 BPM]?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -827,20 +768,19 @@ const MultiStepWizard = (): JSX.Element => {
...
             </Grid>
             <Grid item className={classes.marginContent}>
-              {multiParams.map((param: Data) => {
+              {multiParams.map((param: InternalState) => {
                 if (param.isSetvalEnabled) {
-                  if (param.setValue !== param.setValueActual) {
+                  if (param.ValueDraft !== param.ValueCurrent) {
                     return (
-                      <Typography variant="subtitle1">{`Change ${param.label} to ${param.setValue} ${param.units}?`}</Typography>
+                      <Typography variant="subtitle1">{`Change ${param.label} to ${param.ValueDraft} ${param.units}?`}</Typography>
```
ethanjli: It would be great if we could display the old value, too: e.g. Change FiO2 from 21% to 50%?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -865,20 +805,19 @@ const MultiStepWizard = (): JSX.Element => {
...
             </Grid>
             <Grid item className={classes.marginContent}>
-              {multiParams.map((param: Data) => {
+              {multiParams.map((param: InternalState) => {
                 if (param.isSetvalEnabled) {
-                  if (param.setValue !== param.setValueActual) {
+                  if (param.ValueDraft !== param.ValueCurrent) {
                     return (
-                      <Typography variant="subtitle1">{`Keep ${param.label} to ${param.setValueActual} ${param.units}?`}</Typography>
+                      <Typography variant="subtitle1">{`Keep ${param.label} to ${param.ValueCurrent} ${param.units}?`}</Typography>
```
ethanjli: Let's see what usability is like if we also display the change we're discarding: e.g. Keep FiO2 at 21% instead of changing to 50%?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -865,20 +805,19 @@ const MultiStepWizard = (): JSX.Element => {
...
                 } else if (
-                  param.alarmValues.length &&
-                  (param.alarmValues[0] !== param.alarmValuesActual[0] ||
-                    param.alarmValues[1] !== param.alarmValuesActual[1])
+                  param.alarmValues.lower !== param.alarmValuesActual.lower ||
+                  param.alarmValues.upper !== param.alarmValuesActual.upper
                 ) {
                   return (
-                    <Typography variant="subtitle1">{`Keep ${param.label} alarm range to ${param.alarmValuesActual[0]} - ${param.alarmValuesActual[1]}?`}</Typography>
+                    <Typography variant="subtitle1">{`Keep ${param.label} alarm range to ${param.alarmValuesActual.lower} - ${param.alarmValuesActual.upper}?`}</Typography>
```
ethanjli: Let's see what usability is like if we also display the change we're discarding here, too: e.g. Keep HR alarm limits at [0 BPM - 200 BPM], instead of changing to [50 BPM - 100 BPM]?

ethanjli (commented):
Oops, I had missed your changes in MultiStepWizard because github had hidden the diff and I had assumed it was just a rename of the file before looking at your PR description!

rohanpurohit:
Need to work more on this! not ready for testing

rohanpurohit:
I feel like I'm changing way too many files and should stop before I make too many changes as it gets harder to review! @ethanjli 
It looks like I have resolved most of the GitHub review comments but I haven't resolved them in this PR so that you can refer to them while reviewing! but there's a chance I might have missed something.

In terms of TO-DO's in issue #399, things that are unresolved are:
`Renaming ValueDisplay and ValueSelectorDisplay to something more descriptive/specific`.

As it turns out only one component currently is using `ValueDisplay` which is [ValueController](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/controllers/ValueController.tsx), which is re-used across a lot of `active` components in our frontend, which are:
1. [QuickStartPage](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/quickstart/QuickStartPage.tsx), these are the two ValueClickers for `FiO2` and `Flow` on the Quickstart screen, turns out they were indeed using `ValueController` and not just `ValueClicker`, it was just imported as ValueClicker which is misleading! fixed that in https://github.com/pez-globo/pufferfish-software/pull/413/commits/dcc0ab9548f52606afe725143af74a3d77d6021a and some other changes to quickstart page(next point)
- changed the switch block for ventilation mode returning JSX to a Map, the `defaultContent` Element that's returned when `Ventilation mode is null` or the Map returns `undefined` is currently `Never` because mode is selected as `HFNC` before we reach the QuickStart Page, thus what I did is in the `ValueClicker` I'm setting the `onClick` callback to null, we need to fix these when we have different modes.
- The reason for that is related to #419 and this [TODO](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/quickstart/QuickStartPage.tsx#L132)(overall really weird structure of the `setValue` method) I'll probably get rid of that when I'm fixing #419 
2. the second one is in [DisplayTab](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/settings/containers/DisplayTab.tsx) these are `brightness`, `hour`, `day`, `month` etc

So the point is initially we had decided if you recall that we'll get rid of `ValueDisplay` and stick to `ValueSelectorDisplay` because we thought the latter was used widely, now as it turns out `ValueDisplay` is used far more frequently in our active code
- so I thought about removing `ValueSelectorDisplay` instead (reasons were: it's just a simple react fragment with value) which I tried here https://github.com/pez-globo/pufferfish-software/pull/413/commits/c9e3d09cadb46312c3ed45e23c2f127a27e03550
Problems i faced with that approach were: we lose all the redraw fixes we did in #407 as the whole Info component would render instead of just the value part (not exactly like the issue we had, its just one component that changes redraws continuously) therefore I reverted that now!
- and the argument for having `ValueDisplay` around is 1. has a different look 2. not all values that its used for has a `selector` currently and cant be replaced by `ValueSelectorDisplay` entirely.
So, In conclusion, I'm leaning towards keeping them both! and if so maybe we can rename them? or not?

Testing: I have done mild testing with the changes, I'll do more tomorrow!

frontend/src/modules/dashboard/components/ValueInfo.tsx
```diff
@@ -240,13 +240,8 @@ const ControlValuesDisplay = ({
...
       ? undefined
       : ((alarmLimitsRequest as unknown) as Record<string, Range>)[stateKey];
-  const rangeValues = range === undefined ? { lower: '--', upper: '--' } : range;
-  const alarmTemp = alarmLimits === undefined ? { lower: '--', upper: '--' } : alarmLimits;
-  const alarmLimitsRange = alarmTemp as number[];
-  const { lower, upper } =
-    alarmLimitsRange?.length === 0
-      ? rangeValues
-      : { lower: alarmLimitsRange[0], upper: alarmLimitsRange[1] };
+  const rangeValues = range === undefined ? Range.fromJSON({ lower: '--', upper: '--' }) : range;
```
ethanjli: I believe `Range`'s fields are supposed to be typed as `number`s, so `'--'` shouldn't be a valid value to provide for constructing a `Range`. I'm surprised the original code passed TypeScript. What would happen if you used `NaN` instead of `'--'` here?

frontend/src/modules/modals/MultiStepWizard.tsx
```diff
@@ -3,73 +3,85 @@
...
-interface Data {
+interface InternalState {
   stateKey: string;
   label: string;
   units: string;
-  isAlarmEnabled: boolean;
-  isSetvalEnabled: boolean;
+  isAlarmLimitsEnabled: boolean;
+  isParamEnabled: boolean;
   committedSetting?: number | null;
```
ethanjli: What does `committedSetting` describe? Is it `committedParam`, or does it also apply to alarm limits?

ethanjli (commented):
Here's my review. Everything looks good to me.

Yes, it makes sense to save the issue of names for `ValueDisplay` and `ValueSelectorDisplay` for a future PR. Maybe you can split this off from #399 into a separate Github issue?

Do you have any insight into whether we actually need to import `ValueClicker` in some components and `ValueController` in others, or whether we can import only one of them for all components? I forget if we had already come to a conclusion on this. Maybe the answer will change after #419 is addressed.

If we get rid of `ValueSelectorDisplay`, I suspect we'll need to make explicit containers using `react-redux`'s `connect`. It might be worth exploring that in a separate branch, at least to see what kinds of complexity that approach would introduce. This would seem to me to be the cleaner (and perhaps more idiomatic React-Redux) approach, though I've probably forgotten to consider some important factors, which is why I'd like to see what happens if we try to use containers. I think if we can get rid of `ValueSelectorDisplay` while solving the redraw problem a different way, it will make the frontend's division of responsibilities a bit simpler and easier to understand - if so, that's a big win. As for the visual appearance, maybe `ValueDisplay` should be modified so that its style can be controlled via props.

I will do some testing of this PR tomorrow, before our call.

rohanpurohit:
> Here's my review. Everything looks good to me.
> 
> Yes, it makes sense to save the issue of names for `ValueDisplay` and `ValueSelectorDisplay` for a future PR. Maybe you can split this off from #399 into a separate Github issue?

Yes I can do that!
>  Do you have any insight into whether we actually need to import `ValueClicker` in some components and `ValueController` in others, or whether we can import only one of them for all components? I forget if we had already come to a conclusion on this. Maybe the answer will change after #419 is addressed.

I believe we haven't discussed this! 
So basically `ValueClicker` is just the two buttons for Increment/Decrement (see [here](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/controllers/ValueClicker.tsx#L182)) and we play around with the visual for `Value`, so it does make sense to keep ValueClicker separate (and it anyway has complicated logic related to how it handles the value that's passed)
whereas `ValueController` is a simple component with `ValueDisplay` and `ValueClicker` together.
if you take a look at `AlarmsPage` we use the `ValueClicker` component as it is, and the visual for `value` is defined locally, instead of using `ValueDisplay` or `ValueController` to replace both! see [(AlarmsPage L246](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/alarms/AlarmsPage.tsx#L246))
So having the `Value` separated from `ValueClicker` allows us to do that!

> As for the visual appearance, maybe `ValueDisplay` should be modified so that its style can be controlled via props.

I guess we can definitely do that, and sort of unify how `Value` and `ValueClickers` are used together across the frontend.

frontend/src/modules/modals/MultiStepWizard.tsx
rohanpurohit: To my understanding, it's the `committedParam` value we pass to the child component `SetValueContent`  which is the modal to set `FiO2` and `Flow`, see [here](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/controllers/ValueModal.tsx#L100), so `No` it does not apply to alarm limits, so yes maybe I'll change it to `committedParam`

frontend/src/modules/dashboard/components/ValueInfo.tsx
rohanpurohit: looks like `fromJson` takes `object: any` as parameter, so yes `NaN` also compiles.

ethanjli: Any particular reason you chose to initialize with `0` rather than `NaN` in abe6334?

ethanjli:
> So basically `ValueClicker` is just the two buttons for Increment/Decrement (see [here](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/controllers/ValueClicker.tsx#L182)) and we play around with the visual for `Value`, so it does make sense to keep ValueClicker separate (and it anyway has complicated logic related to how it handles the value that's passed)
> whereas `ValueController` is a simple component with `ValueDisplay` and `ValueClicker` together.
> if you take a look at `AlarmsPage` we use the `ValueClicker` component as it is, and the visual for `value` is defined locally, instead of using `ValueDisplay` or `ValueController` to replace both! see [(AlarmsPage L246](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/alarms/AlarmsPage.tsx#L246))
> So having the `Value` separated from `ValueClicker` allows us to do that!

Ah ok, that really helps! Let's modify the code comments to include this nuance, including the relationship between the two components and where it would be better to use a `ValueController` instead of directly using a `ValueClicker`. Let's also make sure that we only use `ValueClicker` in places where there isn't a corresponding `ValueDisplay` to go along with it (e.g. we can use it in the alarms page). Maybe it would make sense to rename `ValueController` to `ValueSpinner`, in the sense of https://en.wikipedia.org/wiki/Spinner_(computing) - I think "Spinner" is more descriptive than "Controller" about what `ValueController` actually looks like or behaves like.

In a future PR, we could consider passing a component into the props of `ValueController` which could be a `ValueDisplay` or any custom component to render a single value (such as what is used on `AlarmsPage`); in that case, then it should be possible to use `ValueController` on `AlarmsPage`, and maybe then the only file in our app which needs to import `ValueSpinner` will be `ValueController`.

ethanjli:
There's a regression in the "HFNC Control" tab of the breadcrumbs modal in the alarm limits displayed for FiO2 and Flow: they should be displaying the actual alarm limits (i.e. +/- 2 from the current parameter value). However, on this PR they are just shown as [0 - 100]. The previous implementation was also buggy: after changing a parameter setting and committing it, opening the breadcrumbs modal again should reload the alarm limits to the new values which the firmware or simulator backend automatically set based on the parameter value, but instead the alarm limits never change after the frontend initially loads. Since the previous implementation was also buggy, I think we can fix these problems in a future PR, if you create a separate Github issue to track this problem.

I haven't noticed any other problems in this PR when trying to do "normal" user actions on the frontend.

frontend/src/modules/dashboard/components/ValueInfo.tsx
rohanpurohit: Oh no reason as such! changed it to `NaN`

rohanpurohit:
> > So basically `ValueClicker` is just the two buttons for Increment/Decrement (see [here](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/controllers/ValueClicker.tsx#L182)) and we play around with the visual for `Value`, so it does make sense to keep ValueClicker separate (and it anyway has complicated logic related to how it handles the value that's passed)
> > whereas `ValueController` is a simple component with `ValueDisplay` and `ValueClicker` together.
> > if you take a look at `AlarmsPage` we use the `ValueClicker` component as it is, and the visual for `value` is defined locally, instead of using `ValueDisplay` or `ValueController` to replace both! see [(AlarmsPage L246](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/modules/alarms/AlarmsPage.tsx#L246))
> > So having the `Value` separated from `ValueClicker` allows us to do that!
> 
> Ah ok, that really helps! Let's modify the code comments to include this nuance, including the relationship between the two components and where it would be better to use a `ValueController` instead of directly using a `ValueClicker`. Let's also make sure that we only use `ValueClicker` in places where there isn't a corresponding `ValueDisplay` to go along with it (e.g. we can use it in the alarms page). Maybe it would make sense to rename `ValueController` to `ValueSpinner`, in the sense of https://en.wikipedia.org/wiki/Spinner_(computing) - I think "Spinner" is more descriptive than "Controller" about what `ValueController` actually looks like or behaves like.
> 

renamed ValueController to ValueSpinner and added code comments

> In a future PR, we could consider passing a component into the props of `ValueController` which could be a `ValueDisplay` or any custom component to render a single value (such as what is used on `AlarmsPage`); in that case, then it should be possible to use `ValueController` on `AlarmsPage`, and maybe then the only file in our app which needs to import `ValueSpinner` will be `ValueController`.

that's a great idea! yes, we should.

> There's a regression in the "HFNC Control" tab of the breadcrumbs modal in the alarm limits displayed for FiO2 and Flow: they should be displaying the actual alarm limits (i.e. +/- 2 from the current parameter value). However, on this PR they are just shown as [0 - 100]. The previous implementation was also buggy: after changing a parameter setting and committing it, opening the breadcrumbs modal again should reload the alarm limits to the new values which the firmware or simulator backend automatically set based on the parameter value, but instead the alarm limits never change after the frontend initially loads. Since the previous implementation was also buggy, I think we can fix these problems in a future PR, if you create a separate Github issue to track this problem.
> 

Oh yeah, that is a bug, I think it should be fixed now in the latest commit, if it's still buggy I'll open a separate issue.

rohanpurohit:

rohanpurohit:
> There's a regression in the "HFNC Control" tab of the breadcrumbs modal in the alarm limits displayed for FiO2 and Flow: they should be displaying the actual alarm limits (i.e. +/- 2 from the current parameter value). However, on this PR they are just shown as [0 - 100]. The previous implementation was also buggy: after changing a parameter setting and committing it, opening the breadcrumbs modal again should reload the alarm limits to the new values which the firmware or simulator backend automatically set based on the parameter value, but instead the alarm limits never change after the frontend initially loads. Since the previous implementation was also buggy, I think we can fix these problems in a future PR, if you create a separate Github issue to track this problem.
> 

I believe it's completely fixed now on https://github.com/pez-globo/pufferfish-software/pull/413/commits/4dedf8dca06a3c639b5ca5e5236eb9785082cb43, if you could just confirm that then I won't create a separate issue for this! @ethanjli

ethanjli:
Yup, it looks like it's been solved!

ethanjli (approved):

PR #414, opened 2021-07-08
Refactor frontend modules/controller

### LINKED ISSUE #396: Refactor code in frontend modules/controllers

### modules/controllers:

- Marking disableAlarmButton for removal as deprecated code (for now we can just comment it out): #359 (comment)
- Refactoring AlarmModal to have one component which returns just the content inside the modal, and one component which returns the content wrapped inside an AlarmModal popop (which might not even be needed?)? #359 (comment)
- Using a Range object (interface with lower and upper) instead of a variable-length array for alarm limits? #359 (comment)
- Unifying the logic shared between a click handler function and a useEffect in ValueClicker? #359 (comment)
- Marking ValueModal and ControlInfo for deprecation, and marking HFNCGraphView as dead code which will be rehabilitated in the future (i.e. once we get trends plotting, which may be rather far in the future): #359 (comment) and #359 (comment) and #359 (comment)

### LINKED ISSUE #429: Display alarm limits for FiO2 and flow rate consistently with draft settings in HFNC Controls of breadcrumbs modal
As of 4dedf8d on #413 (see https://github.com/pez-globo/pufferfish-software/pull/413#issuecomment-883081048), we display the current alarm limits for FiO2 and flow rate, which fixed a previous bug described in https://github.com/pez-globo/pufferfish-software/pull/413#issuecomment-883001336.

It would be nice if we could create `alarmLimitsDraftFiO2` and `alarmLimitsDraftFlow` selectors which returned `{ lower: parametersDraftFiO2 - 2, upper: parametersDraftFiO2 + 2 }` and `{ lower: parametersDraftFlow - 2, upper: parametersDraftFlow + 2 }`, respectively, and then used them in the breadcrumbs modal. I'm not sure if we'd also need to create `alarmLimitsRequestFiO2` and `alarmLimitsRequestFlow` selectors for the breadcrumbs modal, but we could do a similar thing there too.

rohanpurohit:
This PR includes
- Removed disableAlarmButton.
- Refactored AlarmModal to have one component which returns just the content inside the modal, and one component which returns the content wrapped inside an AlarmModal popop.
- Using a Range object (interface with lower and upper) instead of a variable-length array for alarm limits (This is done in #413 as it was linked to changes to multistepwizard).
- Marked ValueModal and ControlInfo for deprecation.
- Fixed showing draft alarm Limits for Fio2 and flow in HFNCControls or top-level breadcrumbs modal


frontend/src/modules/controllers/AlarmModal.tsx
```diff
@@ -70,12 +70,9 @@ export interface AlarmAdjustProps {
  * @prop {string} units Alarm paramater unit measurement to display
  * @prop {number} committedMin Lower Set Alarm Range Value
  * @prop {number} committedMax Upper Set Alarm Range Value
- * TODO: it was used in ValueModa in values page, can be removed
- * @prop {boolean} disableAlarmButton Toggle to show/hide alarm button
  * @prop {function} updateModalStatus Callback to send current modal open/close status
  * @prop {function} onModalClose Callback after Modal close
- * TODO: rename to 'getCommittedRange' as it just returns the committed values.
- * @prop {function} requestCommitRange Callback on updating the Alarm range values
+ * @prop {function} getCommittedRange Callback on updating the Alarm range values
```
ethanjli: It looks like the function interface for `getCommittedRange` means that it doesn't actually return anything; the `useEffect` which uses it just runs `getCommittedRange` without assigning any return value from that function into anything. It seems like `getCommittedRange` is also a misleading name, or else maybe we can refactor the way we split up code into different functions so that `getCommittedRange` returns something which is then used by the `useEffect` which calls it? I'm not sure, we should discuss our options.

frontend/src/modules/controllers/AlarmModal.tsx
```diff
@@ -110,30 +105,19 @@ interface Props {
  *
  * @returns JSX.Element
  */
-export const AlarmModal = ({
+export const AlarmModalContent = ({
```
ethanjli, line 111: The jsdoc code comment for this `AlarmModalContent` component should be updated, and a code comment should be added for the new `AlarmModal` component; together, the two jsdoc code comments should explain what is the division of responsibilities between the two components.

frontend/src/modules/controllers/ValueModal.tsx
```diff
@@ -105,167 +67,7 @@ interface ContentProps {
...
- *
- * @component A container for setting the value of modal settings in multistep popup.
- * This is a modified version of ValueModal
- *
- * Uses the [[ContentProps]] interface
- * @returns JSX.Element
- */
-
-export const SetValueContent = ({
+const ValueModal = ({
```
ethanjli: This component still needs a jsdoc code comment explaining it. Maybe you could move the jsdoc comment from the top of the file down to here, and then make the code comment here more detailed while making the code comment at the top of the file more general?

frontend/src/modules/controllers/ValueModal.tsx
```diff
@@ -1,15 +1,19 @@
 /**
- * @summary Re-usable wrapper components to display & control value
- * Currently unused, setValueContent is used in MultiStepPopup
+ * ValueModal
  *
+ * @component A container for setting the value of modal settings in multistep popup.
+ * This is a modified version of ValueModal
```
ethanjli: We no longer need this line, right?

frontend/src/modules/controllers/ValueModal.tsx
```diff
@@ -1,15 +1,19 @@
 /**
- * @summary Re-usable wrapper components to display & control value
- * Currently unused, setValueContent is used in MultiStepPopup
+ * ValueModal
  *
+ * @component A container for setting the value of modal settings in multistep popup.
```
ethanjli: Can we make this description clearer and more specific and also provide more context? e.g. it's only the content shown inside a modal for adjusting the values of parameters in the wizard, but the content for adjusting the values of alarm limits is in AlarmModal.tsx, etc.

frontend/src/modules/displays/MultiStepWizard.tsx
```diff
@@ -762,13 +761,11 @@ const MultiStepWizard = (): JSX.Element => {
                 />
               ) : (
                 parameter && (
-                  <AlarmModal
-                    openModal={open && parameter.stateKey === stateKey}
+                  <AlarmModalContent
```
ethanjli: Is there a specific reason we have `ValueModal` and `AlarmModalContent`, rather than both being `*Modal` or both being `*ModalContent`?

ethanjli (commented):
In light testing of the multi-step wizard for adjusting parameters and alarm limits, everything seems to work correctly.

Here are my comments from looking at the changed code - this is all about naming and documentation, so if no other changes come up then we can merge after addressing these.

frontend/src/modules/controllers/AlarmModal.tsx
ethanjli: Looks like `handleCommitRange` or `handleRangeCommit` or `handleCommittedRange` is more appropriate. This is used by MultiStepWizard, which provides `doSetAlarmValues` as the callback function for this.

rohanpurohit: using `handleCommittedRange`

rohanpurohit:
Updated JSdoc comments and fixed #429!

frontend/src/store/controller/selectors.ts
```diff
@@ -267,7 +267,16 @@ export const getSpO2AlarmLimitsRequest = alarmLimitsSelector(getAlarmLimitsReque
 export const getHRAlarmLimitsRequest = alarmLimitsSelector(getAlarmLimitsRequest, 'hr');
 export const getFiO2AlarmLimitsCurrent = alarmLimitsSelector(getAlarmLimitsCurrent, 'fio2');
 export const getFlowAlarmLimitsCurrent = alarmLimitsSelector(getAlarmLimitsCurrent, 'flow');
-
+const alarmLimitsRequestDraftSelector = (selector: SelectorType) =>
```
ethanjli: This name is a bit too general, it implies that this function would be useful in other situations where it might not be relevant. Maybe `automaticAlarmLimitsRangeSelector` is more specific, and we can parameterize the range width (instead of hard-coding 2 into `alarmLimitsRequestDraftSelector`, we hard-code it in the function calls defining `getFiO2AlarmLimitsDraft` and `getFlowAlarmLimitsDraft`

ethanjli (approved):
Everything looks good in code review and testing. Here's just one small rename/interface change of a selector-making function!

rohanpurohit:

PR #416, opened 2021-07-09
Wrap protobuf struct declarations in Pufferfish::Application namespace

### LINKED ISSUE #405: Postprocess nanopb output to declare everything inside a namespace

ethanjli:
This PR fixes #405, including moving the templated `==` and `!=` operators in `Application/States.h` into the `Pufferfish::Application` namespace and upgrading nanopb from v0.4.1 to v0.4.5 (see nanopb's changelog [here](https://github.com/nanopb/nanopb/blob/master/CHANGELOG.txt)).

For records-keeping: 

3. Does this work include, or is it based on, any third-party work which you did not create? Yes: this PR upgrades nanopb from v0.4.1 to v0.4.5; nanopb's license hasn't changed, and licensing compliance for nanopb is already recorded and tracked in our internal system.


rohanpurohit (approved):
Everything looks good in light testing!

PR #417, opened 2021-07-12
Remove clock from Redux store

### LINKED ISSUE #411: Redux is spammed with `BACKEND_CONNECTION_UP` actions

### LINKED ISSUE #415: Remove `CLOCK_UPDATED` action from the frontend?
The initial motivation for the `CLOCK_UPDATED` action was for `store.app.clock` (and thus the `getClockTime` selector) to be a single-source-of-truth of the system time, making it easier to specify arbitrary times when writing unit tests by modifying the clock in the store. However, in practice we've just been using `new Date()` everywhere. If we use `new Date(Date.now())` instead of `new Date()` whenever we need the current time (which we could do in a future refactor), it should be possible to mock out `Date.now()` in jest.

I think it might be useful to remove the `CLOCK_UPDATED` action and the `getClock` and `getClockTime` selectors:
- because the `CLOCK_UPDATED` action also spams the Redux debug console (once every second) in a way that makes the Redux debug console harder to use (actions scroll by quickly) or less useful (actions get discarded quickly)
- and because `HeartbeatBackendListener` shows a possible alternative of just instantiating `new Date()`s to get the current time from inside a React component

It looks like we are only using `getClock` in one place, to get the system time for the settings screen (I'm ignoring the import of `getClock` in `OverlayScreen.tsx`, because that file never actually tries to get a value from that selector; refer to https://github.com/pez-globo/pufferfish-software/issues/411#issuecomment-876695814). We are only using `getClockTime` to get the system time for the clock in the topbar. We're probably already using `new Date()` everywhere else.

In places where we need to have long-running control logic based on the system time, it might make sense to move that logic into redux-saga sagas. In particular, the heartbeat timeout logic should probably be a saga instead, since (like the other saga-related code in our frontend) it's more about the underlying connection behind the store. Restructuring the heartbeat logic this way would also make our frontend implementation more similar to the backend and firmware implementations, where heartbeat timeouts are detected by the driver for the communication protocol, rather than by application-side modules. Our saga would simply dispatch a `backendConnectionLost` action to be handled by react components and/or by reducers - for example, the event log reducer could simply respond to the `backendConnectionLost` action by deciding whether to create a new ephemeral log event about loss of the connection. This should probably remove the need for an invisible `HeartbeatBackendListener` component in an `OverlayScreen.tsx` file.

rohanpurohit:
This PR:
- move `app` backend actions/reducers/selectors to `connection`
- removed `CLOCK_UPDATED` action
- removed `backendHeartbeat` action (only used in HeartbeatBackendListener component)
- removed `HeartbeatBackendListener` component from `OverlayScreen`
- naturally removed  `getClock` and `getClockTime` selectors
- fix the toolbar clock after removing `getClock`, an useEffect with a interval of 1 sec using `new Date()`
- fix display tab in settings after removing `getClockTime`

add temporary saga code to detect backend connection loss.

TODO (ethan/rohan):
- finish saga part for the backend connection (it works! but I'm sure there is a better way to do it)
~fix unit tests after removing `getClock`~
-~add the ability to dispatch event log for software connectivity loss.~


rohanpurohit:
I did not really give much thought to my code for saga! it does work but I'm sure there is a better way to do things! @ethanjli maybe you can co-author that part??

Testing:
- performance on chromium process has improved a lot.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -75,7 +75,34 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    const clockTimer = setInterval(() => {
+      setClockTime(
+        new Date().toLocaleTimeString([], {
+          hour: '2-digit',
+          minute: '2-digit',
+          second: '2-digit',
+          hour12: true,
+        }),
+      );
+    }, 1000);
```
ethanjli: If performance stays the same, the wall clock display will be more synchronized with actual wall clock time (incl. when the seconds digits change) if we use an interval of 500, 250, or 100 Ms.

frontend/src/modules/app/ToolBar.tsx
```diff
@@ -75,7 +75,34 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    return () => {
```
ethanjli, line 102: This is really interesting, (when) is useEffect able to take advantage of the function you return? If so, I believe we may be able to use this in a few other places in the frontend too.

frontend/src/store/app/actions.ts
```diff
@@ -1,8 +1,8 @@
+import { PBMessage } from '../controller/types';
```
ethanjli: app/types shouldn't depend on the existence of controller/types, and ideally it wouldn't depend on the existence of P message.

frontend/src/store/app/actions.ts
```diff
@@ -15,14 +15,10 @@ export function setLocale(locale: string): AppAction {
   return { type: SET_LOCALE, locale };
 }
 
-export function lostBackendConnection(): AppAction {
-  return { type: BACKEND_CONNECTION_DOWN };
+export function lostBackendConnection<T extends PBMessage>(updates: Partial<T>): AppAction {
```
ethanjli: Is there a way we can just define a separate type for this action?

frontend/src/store/app/types.ts
```diff
@@ -38,21 +34,14 @@ interface SetLocaleAction {
...
   status: boolean;
 }
 
-interface UpdateBackendHeartbeat {
-  type: typeof BACKEND_HEARTBEAT;
-}
-
 interface BackendConnectionDown {
   type: typeof BACKEND_CONNECTION_DOWN;
+  update: Record<string, unknown>;
```
ethanjli: What is update supposed to contain?

frontend/src/store/controller/saga.ts
```diff
@@ -49,6 +55,17 @@ function* serviceConnection() {
   yield fork(sendAll, sock);
   yield take(connectionChannel);
   receiveChannel.close();
+  const backendConnected = yield select(getBackendConnected);
+  if (backendConnected) {
+    yield put({
```
ethanjli: Ideally we would instead put a lostBackendConnection action (which would be symmetric with establishedBackendConnection) and then we would modify the eventLogReducer to handle changes to the event log in response to those two actions.

ethanjli (commented):
This is a great start! Here are some initial comments from just looking at the code. We'll also need to add the heartbeat-related functionality back as a saga. I'm not sure if we should also dispatch establishedBackendConnection and lostBackendConnection actions when we detect loss/establishment of websockets - let's check what the backend server does.

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: So this is sort of cleanup, return is more like when the component is unmounted we should clear the timer, I believe in React class components we have a constructor by a similar name, so return does the same thing.

frontend/src/store/app/types.ts
rohanpurohit: This is the part where we pass LogEventType. code, and time! looking at your other comment that `we should be doing this in the reducer` then this can be removed.

frontend/src/store/app/actions.ts
rohanpurohit: If we can do it in the reducer then i guess we don't need this!

frontend/src/modules/app/ToolBar.tsx
rohanpurohit: oh right! i will change that

rohanpurohit:
> We'll also need to add the heartbeat-related functionality back as a saga

I don't think I completely understood this, just add back what I deleted, or add a generator function that dispatches the action in `saga.ts`? btw I deleted that thinking even that spams a lot and doesn't look like we are using it anywhere, but I'm guessing we might need it just in case or for future features?

rohanpurohit:
saw this [TODO ](https://github.com/pez-globo/pufferfish-software/blob/develop/frontend/src/store/app/types.ts#L22) and moved backend stuff from `app` to `connection`

frontend/src/store/app/types.ts
rohanpurohit: deleted it therefore resolving

frontend/src/modules/app/ToolBar.tsx
ethanjli: I see - that would be good to keep in mind anywhere else we allocate resources (such as `setInterval`s) inside `useEffect`s.

frontend/src/store/connection/selectors.ts
```diff
@@ -0,0 +1,22 @@
...
+import { StoreState } from '../types';
+import { ConnectionState } from './types';
+
+// Connection
+
+export const getConnection = (store: StoreState): ConnectionState => store.connection;
+
+// Backend
+
+export const getbackendHeartBeat = createSelector(
```
ethanjli: Let's take this opportunity to fix the capitalization in this selector name, to `getBackendHeartbeat`

frontend/src/store/connection/reducers.ts
```diff
@@ -0,0 +1,36 @@
...
+  backendConnection: false,
+  lastBackendConnectionTime: null,
+};
+
+export function connectionReducer(state = initialState, action: ConnectionAction): ConnectionState {
+  switch (action.type) {
+    case BACKEND_CONNECTION_DOWN:
+      return { ...state, backendConnection: false };
+    case BACKEND_HEARTBEAT:
+      // TODO: Redux recommends **against** using new date inside reducers; instead
```
ethanjli: Let's move this TODO comment to line 29 in the `BACKEND_CONNECTION_UP` case. Even better is if we could create the `new Date()` in `action.ts`'s `establishedBackendConnection()` action creator like what `receivedBackendHeartbeat` does right now.

frontend/src/store/connection/actions.ts
```diff
@@ -0,0 +1,18 @@
...
+export function lostBackendConnection(): ConnectionAction {
+  return { type: BACKEND_CONNECTION_DOWN };
+}
+
+export function establishedBackendConnection(): ConnectionAction {
+  return { type: BACKEND_CONNECTION_UP };
+}
+
+export function receivedBackendHeartbeat(): ConnectionAction {
+  return { type: BACKEND_HEARTBEAT, backendHeartbeat: new Date() };
```
ethanjli: It looks like we no longer need to create a `new Date()` in this action creator.

ethanjli (commented):
> I don't think I completely understood this, just add back what I deleted, or add a generator function that dispatches the action in `saga.ts`? btw I deleted that thinking even that spams a lot and doesn't look like we are using it anywhere, but I'm guessing we might need it just in case or for future features?

What I mean is that if the frontend hasn't received a new message from the backend in the past ~3 seconds (i.e. a heartbeat from the backend), then it should treat the backend as having gone down (e.g. maybe the backend froze or got stuck in an infinite loop), including creating an alarm about loss of connection to the backend. If we implement this logic in a saga, then we don't need to dispatch BACKEND_HEARTBEAT actions, we can just directly dispatch BACKEND_CONNECTION_DOWN when the backend has gone silent.

rohanpurohit:
@ethanjli can you take a look at the latest changes for saga? so basically I'm checking if we have a `STATE_UPDATED` in a time period of 3 seconds, if we don't have that then we dispatch `BACKEND_CONNECTION_DOWN`, since we have this I'm assuming we can now remove the BACKEND_HEARTBEAT actions?

frontend/src/store/connection/selectors.ts
rohanpurohit: I guess I misinterpreted the previous to-do where you probably meant `B` in HeartBeat should not be capitalized. will change if we still need backendHeartbeat and need to add it back

frontend/src/store/controller/connection/backend.ts
```diff
@@ -0,0 +1,31 @@
+import { CallEffect, delay, put, PutEffect, select, SelectEffect } from 'redux-saga/effects';
```
rohanpurohit: If the saga's itself are ok, maybe I was thinking I should move this to `src/store/connection/saga.ts` instead?

ethanjli: I will try to prototype a somewhat different approach to implementing the sagas so that it matches the behavior in the backend more closely, but yes eventually we should be moving several things from src/store/controller into src/store/connection.

ethanjli: The solution I ended up with looks roughly similar to your approach, except I make the race between heartbeat update and heartbeat timeout more explicit, using redux-saga's [race](https://redux-saga.js.org/docs/advanced/RacingEffects) rather than an implicit race on `takeLatest`, and I make the heartbeat update more directly linked to receiving a State Segment. I suspect your approach would also work without your `while (true)` loops in `store/controller/connection/backend.ts`, because usually receiving a State Segment should cancel the `while (true)` loops triggered by the previous State Segment (though I'm not sure about edge cases) - that loop makes the intent of the implementation harder for me to follow.

In 12dec45 I also added commented-out code in the Python backend so that it sends states to the frontend for 10 s, then stops sending states for 10 s, then sends states again for 10 s, etc.; this makes it possible to test the heartbeat timeout detection/recovery behavior in the frontend.

frontend/src/store/controller/reducers/backend.ts
```diff
@@ -128,7 +135,12 @@ export const eventLogReducer = (
     }
     case BACKEND_CONNECTION_DOWN: {
       // Make an ephemeral frontend-only event
-      const logEvent = (action.update as unknown) as LogEvent;
+      const update = {
+        code: LogEventCode.frontend_backend_connection_down,
+        type: LogEventType.system,
+        time: new Date().getTime(),
+      };
+      const logEvent = (update as unknown) as LogEvent;
```
ethanjli: Can we get rid of this typecast now that we're explicitly constructing `update`?

frontend/src/store/controller/reducers/backend.ts
```diff
@@ -128,7 +135,12 @@ export const eventLogReducer = (
     }
     case BACKEND_CONNECTION_DOWN: {
```
ethanjli, line 135: It may be helpful to add a comment that this reducer will behave in unwanted ways (e.g. creating multiple ephemeral events) if `BACKEND_CONNECTION_DOWN` is dispatched multiple times after the connection goes down.

An alternative approach could be to look up events inside `state.ephemeralLogEvents` to see if there's already a `frontend_backend_connection_down` event, but perhaps that adds complexity which isn't worth it for now. You could also mention this other feasible approach in the code comment, for future reference.

ethanjli (commented):
Remaining tasks:

- [x] @rohanpurohit If you can check that my approach also behaves correctly on your end, we can proceed with the next tasks. It would also be helpful if you could review my changes to see if they make sense or if there's anything questionable (e.g. edge-case behavior) you can think of.
- [x] Then we can move a bunch of code from `store/controller` into `store/connection`. In particular, I think probably all of the saga-related code belongs in `store/connection`. I can do this.
- [x] We should get some CPU usage measurements on develop vs. this PR
- [x] We should check user-perceived latency on develop vs. this PR

For records-keeping: 

@rohanpurohit If you could also record your responses to those three questions for your contributions, then we'll be ready to merge after completing those tasks

ethanjli:
CPU usage measurements after running the frontend dashboard with the simulator backend on the RPi for at least 10 minutes:
- On develop (85ae493), CPU usage of the chromium process is usually between 90% and 100%, occasionally reaching 85% and rarely reaching 80% or 105%
- On this PR (12dec45), CPU usage of the chromium process is usually between 65% and 75%, often reaching 60%, occasionally reaching 85%, rarely reaching 90% or 55%.

If end-user latency looks good to you, I think this concludes the performance optimization on the frontend for running in the dashboard; in future commits we just need to check against performance regressions. I think it's still worth fixing the issues of massive redraws for the alarm border, modals, etc., in order to improve responsiveness to user actions. Regardless, for HFNC mode we need to fix the issue of the red border flashing on the quickstart screen (and being slow in doing that) when we stop ventilation.

CPU usage in the Python backend server is still very high, and based on what we've seen in this PR I suspect the culprit is that the backend is polling on clock updates (like what we removed with redux + sagas in this PR) every millisecond (!!!). The simple solution would be to accept higher clock jitter, e.g. by polling only every 2 ms or every 5 ms; the more efficient, more accurate, but potentially much more complicated solution, would be to have all timing-related filters (state synchronizers, backend protocol, server protocol) report the next clock updates they need, and then ventserver.integration._trio.process_clock would sleep for variable durations and wake for the deadline of each requested clock update.

frontend/src/store/controller/connection/backend.ts
rohanpurohit: Oh awesome! so we only use `race` if the frontend is frozen or there is a delay of sorts and we dispatch lost backend action on loss of WebSocket. works as expected on my end!

rohanpurohit:
Looks good to me in terms of latency! 

ethanjli:
Moving the connection-specific code from `store/controller` into `store/connection` turned into a giant renaming operation because then it also made sense to move the protobuf files (and some associated type definitions) from `store/connection/proto` to `store/proto`, so a lot of frontend files have been touched. I've tried to restrain myself by not doing more moving of code or renaming of things beyond that, since those things should really be saved for a future PR. @rohanpurohit Can you try to go to all of the various frontend screens and press all of the various buttons to ensure that the frontend doesn't break (e.g. with console.log errors or with typescript/eslint/etc. complaints in `yarn start`) on these latest changes? Once it looks good to you, you can go ahead and merge this PR.

ethanjli (approved):

rohanpurohit:
> Moving the connection-specific code from `store/controller` into `store/connection` turned into a giant renaming operation because then it also made sense to move the protobuf files (and some associated type definitions) from `store/connection/proto` to `store/proto`, so a lot of frontend files have been touched. I've tried to restrain myself by not doing more moving of code or renaming of things beyond that, since those things should really be saved for a future PR. @rohanpurohit Can you try to go to all of the various frontend screens and press all of the various buttons to ensure that the frontend doesn't break (e.g. with console.log errors or with typescript/eslint/etc. complaints in `yarn start`) on these latest changes? Once it looks good to you, you can go ahead and merge this PR.

mostly tried everything! looks good to me. merging

PR #418, opened 2021-07-14
Move PWM's `set_duty_cycle_raw` from `public` to `protected`

### LINKED ISSUE #149: Simplify hardware PWM interface
Right now the hardware PWM interface exposes a `setDutyCycleRaw` method which is a bit more complicated/error-prone to use, all just to provide a way to avoid doing any floating-point operations (I suspect this may be premature optimization, since our bottleneck should be I2C/SPI sensors rather than CPU operations). Either we should remove that method from the public interface, or we should define some pre-set resolution for the PWM (e.g. `uint8_t` duty cycle).

ethanjli:
This PR fixes #149.

### DIFF
--- a/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/HAL/Interfaces/PWM.h
+++ b/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/HAL/Interfaces/PWM.h
@@ -42,13 +42,10 @@ class PWM {
   PWMStatus set_duty_cycle(float duty);
 
   /**
-   * Set a duty cycle of PWM, can be done when PWM is active
-   * this function does NOT start the PWM output.
-   * This is faster than setDutyCycle() as no floating point calculation is done
-   * @param duty    an integer between 0 and getMaxDutyCycle() (inclusive) for
-   * the desired duty cycle
+   * Returns the maximum duty cycle that can be set with setDutyCycleRaw()
+   * @return the maximum duty cycle
    */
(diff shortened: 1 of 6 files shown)


ethanjli:
For records-keeping: 

@rohanpurohit If you have access to a ventilator unit where the Nucleo and prototype electronics board have been integrated together with the flow sensors and the solenoid valves, would you be able to test this PR to confirm that the valves are closed upon program startup (even if the valves were previously opened before the Nucleo was forcibly restarted, e.g. by pressing its reset button)? The only place where we used `set_duty_cycle_raw` was at startup.

rohanpurohit:
> @rohanpurohit If you have access to a ventilator unit where the Nucleo and prototype electronics board have been integrated together with the flow sensors and the solenoid valves, would you be able to test this PR to confirm that the valves are closed upon program startup (even if the valves were previously opened before the Nucleo was forcibly restarted, e.g. by pressing its reset button)? The only place where we used `set_duty_cycle_raw` was at startup.

We were supposed to bring a working ventilator unit from one of the hospitals, the `other unit` we have has the full feature power board and the new PCB I presume and there are no connections made on that one, so I think we can test it on a ventilator unit once Renji collects the one from the hospital, so I guess its kinda blocked from my end, maybe you could go ahead with it if you have tested it on your end?

ethanjli:
Currently I don't actually have a way of testing this with the hardware integrated. Since this change is basically static (i.e. any issues which occur should probably be caught by the compiler), let's merge in this fix for now and then we can file a new issue if a regression is found later.

PR #421, opened 2021-07-15
Refactor public interfaces of some low-level driver code for the Nonin sensor

rohanpurohit:
This PR includes:
- remove enums that were not used
- Reorder enums, consts to Types.h
- Rename `SignalAmplitude` to `SignalQuality`
- Rename `Pleth` to `PlethPulseAmplitudes`
- Rename `PacketMeasurements` to `Sample`
- Rename `StatusByteStruct` to `Status`
- Reorder `StatusByteStruct` fields as per datasheet
- Reorder fields in Sample as per datasheet (pg 6, Serial Data format (2) table)
- Rename `heart_rate` to `hr`
- Rename `spo2_d_beat` to `spo2_b_b`
- Rename `spo2_d_fast` to `spo2_fast` (as per datasheet)
- Rename `nonin_oem_revision` to `firmware_revision`
- Remove unnecessary code comments
- Use FrameBufferStatus for FrameBuffer instead of BufferStatus
- Rename `get_6bit_data` to `get_spo2_data` as it essentialy only gets SpO2 data from the packet
- Rename `get_9bit_data` to `get_hr_data`
- Rename `mask_6bit` to `spo2_missing_data` which is equal to `0x7F` or 127, this is an indicator for missing SpO2 data, we just ignore it if we get these bytes
- Rename `mask_9bit` to `hr_missing_data` which is `0x01FF` or 511 this is an indicator for missing HR data, we just ignore it if we get these bytes
- Rename `FrameInputStatus` `available` enum to `output_ready`
- Rename `FrameInputStatus` `framing_error` enum to `checksum_failed`
- Rename `PacketStatus` `framing_error` enum to `checksum_failed`
- Rename `PacketStatus` `missed_data` enum to `frame_loss`
- Rename `PacketInputStatus` `missed_data` enum to `frame_loss`
- Rename `PacketStatus` `available` enum to `ok`

TODO: check for other changes


ethanjli:
Partial review of the existing Nonin OEM III code:

- Types.h:
    - Should we move enum class statuses for Receiver classes specific to those classes (e.g. "FrameInputStatus") into those classes (e.g. "FrameReceiver::InputStatus")?
    - enum class PacketStatus:
        - What does "not_available" mean? Presumably any value other than "available" should imply "not available".
        - What's the difference between "framing_error" and "missed_data"?
    - enum class SignalAmplitude:
        - Should this be renamed to SignalQuality?
    - struct StatusByteStruct:
        - Should this be renamed to FrameStatus or even just Status?
    - using StatusByteError:
        - Should this be renamed to FrameStatuses or even just Statuses?
        - Is this just dead code?
    - using Pleth:
        - Should this be renamed to PlethPulseAmplitudes?
    - struct PacketMeasurements:
        - Should this be renamed to Sample?
        - Should the fields in here be reordered to match the order of items on page 8 of the datasheet? (e.g. hr_d should be after spo2)
        - Should "heart_rate" in the field names be renamed to "hr" for consistency with the datasheet's names?
        - uint8_t spo2_d_beat:
            - Should this be renamed to spo2_b_b for consistency with the datasheet's name?
        - uint8_t nonin_oem_revision:
            - Should this be renamed to firmware_revision to match the description in the datasheet?
            - Should our Sensor class check that this value should be within some allowed range?
        - The datasheet says that spo2 = 127 and hr = 511 as sentinel values when those quantities cannot be computed. How are we reporting missing data here? Should we add bool fields has_spo2 and has_hr, or is that already reported in some other way?
        - std::array<bool, packet_size> ...:
            - Should we make a `using Flags = std::array<bool, packet_size>` and then just use that for all these declarations?
    - static const uint8_t mask_*:
        - Since these are single-byte masks, we should use binary notation (and we can use the digit grouping symbol `'` for readability)
- FrameBuffer.h
    - Frame frame_buffer_:
        - The doxygen comment for the `output` method has an unclear description which doesn't describe what it actually does
        - Should this be renamed to `frame_`?
    - uint8_t received_length:
        - This should be a `size_t` since it's used for array indexing (and then we can remove a static_cast in the cpp file)
- FrameBuffer.cpp
    - There are too many useless comments (e.g. `/* Increment the frame buffer index */`), it adds too much noise to code comments.
    - `received_length_++`:
        - In general we should use the prefix `++` and `--` rather than the postfix `++` and `--`: e.g. `++received_length_`
    - `BufferStatus input()`, `BufferStatus output()`:
        - The way these methods use `BufferStatus::partial` abuses the semantics for `BufferStatus`, where `partial` means "only a partial buffer operation was possible" - this is a safety problem. `FrameBuffer` should instead have something more like `ok` and `waiting`, and maybe `empty` and full` if needed, or else we should redesign the interface and separation of responsibilities between `FrameBuffer` and `BufferReceiver` to remove the need for a `partial` or `waiting` status code from `FrameBuffer`. Or we could just have `ok` and `waiting`, and also `input_overwritten` like in `ChunkSplitter`.
    - shift_left:
        - We should instead do `if (received_length_ == 0) return;`

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Types.h
```diff
@@ -0,0 +1,104 @@
...
+                     /// quality
+  yellow_perfusion,  /// YPRF: Yellow Perfusion – Amplitude representation of
+                     /// medium signal quality
+  green_perfusion    /// Green Perfusion – Amplitude representation of high signal
+                     /// quality
+};
+
+/* Frame */
+const uint8_t frame_max_size = 5;
+using Frame = std::array<uint8_t, frame_max_size>;
```
ethanjli: It makes sense to me that `frame_max_size` and `packet_size` belong in Types.h, because they're configuration parameters which we took directly from the datasheet. However, maybe the type definitions of `Frame` and `Packet` can just live in the FrameReceiver.h/PacketReceiver.h/etc. files, since the specific choice of data structure is an implementation detail which doesn't come from the datasheet and isn't needed by the outside world (which only needs to use `Sample).

By this logic, we should still define `Flags`, `PlethPulseAmplitudes`, etc., in Types.h, because they're part of `Sample` which is used by the outside world.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Types.h
```diff
@@ -0,0 +1,104 @@
...
+  red_perfusion,     /// Red Perfusion – Amplitude representation of low signal
+                     /// quality
+  yellow_perfusion,  /// YPRF: Yellow Perfusion – Amplitude representation of
+                     /// medium signal quality
+  green_perfusion    /// Green Perfusion – Amplitude representation of high signal
+                     /// quality
+};
+
+/* Frame */
+const uint8_t frame_max_size = 5;
```
ethanjli, line 41: Let's also add a code comment describing which Serial format we're using from the Nonin OEM III datasheet (I think it's format 2) for our low-level driver, including the URL of the datasheet.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Types.h
```diff
@@ -0,0 +1,104 @@
...
+
+// Status Mask
+static const uint8_t start_of_frame = 0b1000'0000;
+static const uint8_t snsd = 0b0100'0000;
+static const uint8_t artf = 0b0010'0000;
+static const uint8_t oot = 0b0001'0000;
+static const uint8_t snsa = 0b1000;
+static const uint8_t yprf = 0b0110;
+static const uint8_t rprf = 0b0100;
+static const uint8_t gprf = 0b0010;
```
ethanjli: Let's wrap these definitions in an internal namespace such as `StatusMasks`. Since these are implementation details which just come from the datasheet, it's fine for these definitions to be in `Driver::Serial::Nonin::StatusMasks` rather than being exported as part of the "public" interface of the file, which is `Driver::Serial::Nonin`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Types.h
```diff
@@ -0,0 +1,104 @@
...
+static const size_t spo2_index = 2;
+static const size_t firmware_revision_index = 3;
+static const size_t spo2_d_index = 8;
+static const size_t spo2_fast_index = 9;
+static const size_t spo2_b_b_index = 10;
+static const size_t e_hr_index = 13;
+static const size_t e_spo2_index = 15;
+static const size_t e_spo2_d_index = 16;
+static const size_t hr_d_index = 19;
+static const size_t e_hr_d_index = 21;
```
ethanjli: Let's wrap these definitions in an internal namespace such as `ByteIndex`. This way we can also remove the `_index` postfix from each variable name.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameBuffer.h
```diff
@@ -27,15 +27,15 @@
...
 
-namespace Pufferfish {
-namespace Driver {
-namespace Serial {
-namespace Nonin {
+enum class FrameBufferStatus {
+  ok = 0,
+  waiting,
+  full,
+};
```
ethanjli, line 41: Here is a comparison of the return code semantics of our two buffer classes:
- `Util::Containers::RingBuffer::push` returns `full` if there is no more space available to push bytes, so that the `push` could not successfully complete. Similarly, `Driver::Serial::Nonin::FrameBuffer::input` returns `full` if the input if there is no more space available to push bytes, so that the `input` operation could not successfully complete - which is what we want.
- `RingBuffer::push` returns `ok` if there more space available to push bytes. We should change `FrameBuffer::input` to have the same semantics; right now it returns `waiting` instead.
- `RingBuffer::push` doesn't have any concept of an output being ready or of the caller needing to stop adding bytes in order to process the ones already in the bufer. So `FrameBuffer::input` should have a new return code with a different name to indicate this; right now it returns `ok` instead, which conflicts with the semantics of `RingBuffer::push`'s `ok` return code.
- `RingBuffer::pop` returns `empty` if there are no bytes available to pop, so that the `pop` could not successfully complete. `FrameBuffer::output` should return some error code if there is not a complete buffer to inspect, so that the `output` could not successfully complete; right now it returns `waiting`, which is fine.
- `RingBuffer::pop` returns `ok` if there is a byte available to push. `RingBuffer::output` should return `ok` if there's a complete frame to inspect; right now it returns `ok`, which is what we want.

Either we should have two separate enums for `InputStatus` and `OutputStatus` (which is what I've done in the `Protocols::Transport` classes) or we should have a single enum for `FrameBufferStatus` but be more descriptive in the `@return` code comment about exactly which error codes it might return and what the semantics of it are (which is what I've done in RingBuffer.h, see [this example](https://github.com/pez-globo/pufferfish-software/blob/d386a033885c6d89196643199d5f374c170ebfe7/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Containers/RingBuffer.h#L79)).

Since this renaming is slightly more complicated (i.e. we're changing the meaning of `ok`, which is already being used elsewhere in the code), we could save it for the next PR - it's up to you.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Device.cpp
```diff
@@ -23,65 +23,41 @@
...
+    case PacketInputStatus::output_ready:
       break;
   }
 
-  /* On packet input available invoke output method to read sensor measurements
-   */
-  if (packet_receiver_.output(sensor_measurements) !=
-      PacketReceiver::PacketOutputStatus::available) {
-    /* Return sensor status is waiting to receive more bytes of data */
+  if (packet_receiver_.output(sensor_measurements) != PacketOutputStatus::ok) {
```
ethanjli: This made me nervous because I initially thought there were error codes in `PacketOutputStatus` and we were just ignoring them without a `TODO` comment; it turns out that right now `PacketOutputStatus` only has `ok` and `waiting`, so it doesn't have error codes.

If `clang-tidy` allows it, let's convert this to a two-case `switch` statement handling `ok` and `waiting`, so that we should get a clang-tidy warning about an unhandled enum in the `switch` if someone adds another code (e.g. an error code which needs to be handled) to `PacketOutputStatus` in the future.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameReceiver.h
```diff
@@ -22,34 +22,29 @@
...
-namespace Pufferfish {
-namespace Driver {
-namespace Serial {
-namespace Nonin {
+namespace Pufferfish::Driver::Serial::Nonin {
+
+/* FrameReceiver input status return values */
+enum class FrameInputStatus {
+  ok = 0,           /// Input is ready to receive new bytes of sensor data
+  checksum_failed,  /// Error in checksum or status byte or in byte 1 of a frame
```
ethanjli, line 32: [probably for next PR] We should probably split this up into two error codes: `invalid_header`/`invalid_start_of_frame` (i.e. `new_frame[0] != 0x01 || (new_frame[1] & mask_start_of_frame) != mask_start_of_frame` in `validate_frame`) and `invalid_checksum`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -29,11 +29,8 @@ namespace Pufferfish::Driver::Serial::Nonin {
  * Called based on start_of_frame_status_ private variable
  */
 bool validate_start_of_frame(const Frame &new_frame) {
```
ethanjli, line 31: [probably for the next PR] From looking at the code, this looks like it should actually be `validate_start_of_packet`? It's weird to me that `FrameReceiver` expects to know anything about the structure of packets, such as whether a frame is at the start of a packet.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -46,87 +43,64 @@ bool validate_start_of_frame(const Frame &new_frame) {
  * validateFrame function is called to validated the every frame for Status byte
  * and Checksum.
  */
-FrameReceiver::FrameInputStatus validate_frame(const Frame &new_frame) {
+FrameInputStatus validate_frame(const Frame &new_frame) {
```
ethanjli, line 46: [probably for next PR] Right now `validate_frame` has an unnecessary nested `if` statement structure. We should probably just split up `validate_frame` into two functions `validate_frame_header` and `validate_frame_checksum` and use them directly in `update_frame_buffer`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameReceiver.h
```diff
@@ -89,9 +81,6 @@ class FrameReceiver {
```
ethanjli, line 81: [for the next PR] The code comment for `start_of_frame_status` is confusing to me because it's talking about packets (and also the descriptions were already run-on sentences with a structure I don't understand, before clang-format merged separate lines for `true` and `false`); but FrameReceiver.h should have no awareness of anything about packets. Maybe we can also make the variable slightly shorter and clearer by naming it `at_start_of_frame` or something.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -46,87 +43,64 @@ bool validate_start_of_frame(const Frame &new_frame) {
...
     start_of_frame_status_ = false;
   }
+
   return input_status_;
 }
 
-FrameReceiver::FrameInputStatus FrameReceiver::input(const uint8_t new_byte) {
-  /* Update the frame buffer with new byte received */
+FrameInputStatus FrameReceiver::input(const uint8_t new_byte) {
   input_status_ = this->update_frame_buffer(new_byte);
```
ethanjli, line 87: [for next PR] I don't think we need to use `this->`, and in general we don't use `this->` in our codebase.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -30,16 +30,16 @@ InitializableState Sensor::output(float &spo2, float &hr) {
     return InitializableState::failed;
   }
 
-  if (device_.output(measurements_) == Device::PacketStatus::available) {
+  if (device_.output(measurements_) == PacketStatus::ok) {
```
ethanjli, line 33: Do we need to add a TODO comment about handling the case where the `Device::output` method doesn't return `ok`? Right now it seems that `Sensor::output` returns `ok` regardless of what happens. Maybe just what we need to change in order to add POST checks for the Nonin sensor? (there's probably other things we'll need to change, too).

This is probably code I wrote in a hurry. If I had used a return-early structure (i.e. `status = device_.output(...); if (status !== ok) {return an error code or do nothing but have a TODO comment for dealing with this} `), this would have been easier to catch and understand.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/PacketReceiver.h
```diff
@@ -25,76 +25,26 @@
...
-  std::array<bool, packet_size> sensor_disconnect;
-  std::array<bool, packet_size> artifact;
-  std::array<bool, packet_size> out_of_track;
-  std::array<bool, packet_size> sensor_alarm;
-  std::array<SignalAmplitude, packet_size> signal_perfusion;
-};
-
-static const uint8_t mask_6bit = 0x7F;
+using PacketOutputStatus = Pufferfish::Driver::Serial::Nonin::FrameOutputStatus;
+static const uint8_t spo2_missing_data = 0x7F;
```
ethanjli: Is `spo2_missing_data` a mask? If so, should `spo2_missing_data` be moved to be with the other masks in Types.h? Also, what does "spo2 missing data" mean? Would it be clear if we just named it to "spo2 missing"? I think "spo2 missing data" is too ambiguous because it could either be saying "There is something called SpO2 and it needs some data which is currently missing", "We are missing the data for SpO2", or "Here is some data describing why SpO2 is missing".

If `spo2_missing_data` is a non-mask value defined in the datasheet, perhaps we should move it to `Types.h` (maybe into a `SpecialConstants` or `ErrorConstants` or `Flags` internal namespace) and use the datasheet's format for representing the value, e.g. `static const uint8_t spo2_missing = 127; // Normally we'd use hex, but the datasheet lists it as 127 in decimal so we do the same thing here`.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/PacketReceiver.h
```diff
@@ -103,39 +53,22 @@ inline uint16_t get_6bit_data(uint8_t byte_data) {
  * @param  lsbByte - MSB to extract 5 bits
  * @return Calculated 9 bit data from MSB and LSB
  */
-inline uint16_t get_9bit_data(uint8_t msb_byte, uint8_t lsb_byte) {
+inline uint16_t get_hr_data(uint8_t msb_byte, uint8_t lsb_byte) {
   /* Pack 2 bits of MSB and 6 bits of LSB for 9 bits of heart rate data  */
   static const uint16_t msb_shift = 7;
   static const uint16_t mask_msb = 0x18;
-  static const uint16_t mask_9bit = 0x01FF;
+  static const uint16_t hr_missing_data = 0x01FF;
```
ethanjli: If `hr_missing_data` is actually a value defined in the datasheet, perhaps we should move into Types.h together with the "SpO2 missing" flag

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/PacketReceiver.cpp
```diff
@@ -33,214 +33,93 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
+  if ((byte_value & yprf) == yprf) {
+    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::yellow_perfusion;
+  } else if ((byte_value & rprf) == rprf) {
+    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::red_perfusion;
+  } else if ((byte_value & gprf) == gprf) {
+    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::green_perfusion;
   } else {
-    sensor_measurements.signal_perfusion[frame_index] = SignalAmplitude::no_perfusion;
+    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::no_perfusion;
   }
```
ethanjli: [for a future PR] To avoid copying `sensor_measurements.signal_perfusion[frame_index]` so many times, maybe we could first get a non-const ref as `SignalPerfusion &signal_perfusion = sensor_measurements.signal_perfusion[frame_index]` and then just assign to `signal_perfusion`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/PacketReceiver.cpp
```diff
@@ -33,214 +33,93 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-  /* Update the PLETH and Status byte errors into the packet measurements */
-  for (frame_index = 0; frame_index < packet_size; frame_index++) {
+
+  // SpO2 data
+  sensor_measurements.spo2 = get_spo2_data(packet_data[spo2_index][byte3]);
+  sensor_measurements.spo2_d = get_spo2_data(packet_data[spo2_d_index][byte3]);
+  sensor_measurements.spo2_fast = get_spo2_data(packet_data[spo2_fast_index][byte3]);
+  sensor_measurements.spo2_b_b = get_spo2_data(packet_data[spo2_b_b_index][byte3]);
+  sensor_measurements.e_spo2 = get_spo2_data(packet_data[e_spo2_index][byte3]);
+  sensor_measurements.e_spo2_d = get_spo2_data(packet_data[e_spo2_d_index][byte3]);
```
ethanjli: [for a future PR] I wonder what `byte3 means` and why we use it so many times. Maybe we can find a more descriptive name, e.g. have a `ByteIndices` namespace in Types.h with constants for `header_byte`, `status_byte`, `pleth_byte`, `payload_byte`/`data_byte`, and `checksum_byte`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/PacketReceiver.cpp
```diff
@@ -33,214 +33,93 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
+  for (size_t frame_index = 0; frame_index < packet_size; ++frame_index) {
     sensor_measurements.packet_pleth[frame_index] = packet_data[frame_index][byte2];
     read_status_byte(sensor_measurements, frame_index, packet_data[frame_index][byte1]);
   }
 }
 
-PacketReceiver::PacketInputStatus PacketReceiver::input(const Frame &frame) {
-  /* Check the frame received is first frame in the packet SYNC bit is 1 */
+PacketInputStatus PacketReceiver::input(const Frame &frame) {
   if ((frame[1] & 0x01U) == 0x01) {
```
ethanjli, line 85: Since this `if` statement is changing a lot of internal state (which is where our codebase is most error-prone), we should use descriptive naming and code comments to help communicate what's going on.

What does the expression being tested by the `if` statement actually mean? We can answer this question in the code by having a descriptive function name to replace `(frame[1] & 0x01U) == 0x01`.

What does this `if` statement try to achieve when the condition is met? We can answer this question in the code by adding a code comment which summarizes everything inside, as a replacement for the line-by-line code comments which take too much work to try to compile into a summary-level understanding.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/PacketReceiver.cpp
```diff
@@ -33,214 +33,93 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-    input_status_ = PacketInputStatus::missed_data;
+    input_status_ = PacketInputStatus::frame_loss;
     return input_status_;
   }
 
-  /* Update the frame received to packet */
   packet_data_[received_length_] = frame;
-  /* Increment the frame index */
-  received_length_ = received_length_ + 1;
+  received_length_ += 1;
```
ethanjli: This could be shortened to `++received_length`.

ethanjli (commented):
I reviewed all sections of code touched by these changes; all of your changes look good. Most of these are suggestions for the next Nonin driver refactoring PR. Since there is some code changed which would not be trivially identified by compiler errors, later today I'll test the updated driver code with my Nonin sensor just to confirm that there aren't any obvious regressions in normal operation. Error-handling is complicated and needs more refactoring and possibly some deeper changes, so I won't test that in this PR.

All of the suggested changes for this PR in this review should be minor changes (anything which a compiler can't immediately and obviously catch as an error should be saved for a future PR; and anything about the quality of files besides FrameBuffer.h and Types.h is technically outside the scope of this PR), so if you add your responses to the records-keeping questions for open-source licensing, then I'll approve this PR for merging after very light testing.

rohanpurohit:

ethanjli (approved):
I just tested the firmware together with my Nonin sensor, and everything seems to work fine in the normal case!

PR #425, opened 2021-07-19
Unit test for Util

raavilagoo:
1.EnumMap
2.EnumSet
3.OrderedMap
4.RingBuffer


ethanjli:
Example of the overall structure for RingBuffer tests
```
newest: ^
oldest: *
Scenario (non-volatile ringbuffer)
Given a ringbuffer with capacity 5 to which 4 bytes of data were pushed and then 3 bytes of data was popped
// internal state should look like [] [] [*a] [^b] []
    When: pop
        // then internal state should look like [] [] [] [*^b] []
        Then: (peeks)
    When: push
        // then internal state should look like [] [] [] [*b] [^c]
Given: (... for [] [] [o] [] [n])
    When: pop
        Then: (... for [] [] [] [o] [n])

Scenario (volatile ringbuffer)
Given a volatile ringbuffer...
```

raavilagoo:
Hi @ethanjli  , I wanted some clarification with respect to constructing a vector with initializer list , 
I have declared vector as this, PF::Util::Containers::Vector<uint8_t, 4> vector1{10, 20, 30}. If I check the values at every index, I get  0. Could you elaborate on this? 
Similarly with EnumMap, OrderedMap and EnumSet.

ethanjli:
I tried declaring some container instances at the top of the `main` function in main.cpp and inspecting them with the STM32 debugger. Here are my results

- `PF::Util::Containers::Vector<uint8_t, 5> vec{0, 1, 2, 3, 4};` causes `vec` to be initialized with an internal buffer which is all zeroes. So I'm able to reproduce the problem you found with how I expected the initializer list to work vs. how it actually works.
- `PF::Util::Containers::Vector<uint8_t, 5> vec({0, 1, 2, 3, 4});` has the same problem.
- `std::initializer_list<uint8_t> init{0, 1, 2, 3, 4}; PF::Util::Containers::Vector<uint8_t, 5> vec(init);`  also has the same problem.
- When I use the debugger to step through the `Vector` constructor defined on the initializer list, what I see is that the first `push_back` method called in the constructor returns early with `IndexStatus::out_of_boudns`, because `size_ == array_size`. In this case, `size_` was already initialized as 5, because the constructor pre-initialized size_ to 5. This is clearly a bug because I was careless in implementing the constructor and because I never tested my code.

By removing the `size_(init.size())` member initialization from the `Vector(std::initializer_list<Element> init)` constructor, I was able to fix the bug and get the correct result for `PF::Util::Containers::Vector<uint8_t, 5> vec{0, 1, 2, 3, 4};`. Can you include this fix into this PR?

Also, we should make sure that you will have easy access to an STM32H743 Nucleo board so that you'll be able to upload the firmware with some test code added to the top of the main function, just to be able to take advantage of the debugger for troubleshooting these kinds of issues in lower-level code even without a fully working ventilator setup. @renjipanicker Is this already possible / how easy would it be to make this possible, so that both Rohan and Raavi have easy access to Nucleo boards for troubleshooting software-only issues using the STM32Cube IDE's debugger?

raavilagoo:
Hi @ethanjli, this is ready for review.

ethanjli (commented):
Notes from live review in meeting:

RingBuffer tests:
- Modify RingBuffer so that the template arg is capacity, and buffer_ is declared as buffer_[capacity + 1]
- Update uses of RingBuffer and `using` derived types for this change
- Unit test is still designed for [] [] [] [] [], and we say capacity is 4 in our THEN
- Prediction: we'll have 5 * 4 - 5 cases (we subtract 5 because newest can be oldest - 1 but cannot be incremented into oldest (because newest == oldest means empty))
- Order & format of GIVENs: push n bytes, pop [0, 1, .., n] bytes; push n + 1 bytes, pop [0, 1, ..., n + 1] bytes, etc.

Vector:
- Wording for initializer list tests

Other Containers:
- Wording in WHENs ("is called on" is not needed)
- GIVENs should specify template parameters including Value types
- Containers for const/non-const ref tests should use reference_wrapper types
- Initializer list scenario:
    - GIVEN a container constructed with an initializer list with keys 1, 2, 3, and 4
        - WHEN we call the available method
            - THEN ...
        - WHEN we call the has method for keys 1, 2, 3, and 4
            - THEN ...
        - WHEN we call the items method (only for OrderedMap)
            - THEN ...
        - WHEN we iterate through all items using the range-based for loop
            - THEN ...
        - WHEN we call input method
            - THEN ...
- THENs for indexing into containers:
    - THEN the has method returns true for keys 1, 2, 3, 4, and false for all other keys in the range [-10000, 100000]
    - THEN indexing into the containers for keys 1, 2, 3, 4 returns the values 10, 11, 12, 13, respectively
- Scenario for mutating the fields of struct values in containers (incl. vectors)
- Erase scenario:
    - Can use "items" instead of "keys" or "key-value pairs" to refer to key-value pairs
    - GIVEN should specify the values of the keys in the map
    - WHENs should specify the values of the keys being erased
    - Erase on capacity and capacity + 1 should specify values, but also explain that they correspond to capacity and capacity + 1; test should be constructed so that erase method returns ok. Should have one GIVEN for the capacity where the keys are erasable, and another GIVEN for the same capacity where they aren't erasable
- Other wording

raavilagoo:
@ethanjli  This is ready for review.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Containers/Vector.h
```diff
@@ -34,6 +34,7 @@ class Vector {
   // initializer list (e.g. Vector<uint8_t, 5> vec{1, 2, 3})
   // Note: this makes copies of the values!
   Vector(std::initializer_list<Element> init) {
```
ethanjli, line 38: TBD: does `Vector(std::initializer_list<Element> init) : size_(0) {` work and pass cppcheck?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -1,303 +1,355 @@
...
-      THEN('The avaliable method reports that 1 key-value pair are avaliable') 
-      THEN('The full method reports that the EnumMap is not completely filled') 
+      THEN('The avaliable method reports that 1 item is avaliable') 
+      THEN('The full method reports that the EnumMap is not full') 
       THEN('The empty method reports that the EnumMap is non-empty') 
-      THEN('The has method returns true for 3 keys') 
+      THEN('The erase method reports ok status for 3rd key ') 
+      THEN('The has method returns true for 1,2 and 3 keys') 
+      THEN('The has method returns false for [-1000]to[1000] keys')
+      THEN('Changing the value of struct elements does not change values of key 1,2 and 3')
```
ethanjli: The struct fields are correctly set by direct access using the [] operator:
```
map[1].value = 122
map[2].value = 3
map[3].value = 899
assert(map[1].value == 122)
```

The map values are correctly set by assignment on the [] operator:
```
TestStruct val4{122}
TestStruct val5{3}
TestStruct val6{899}
map[1] = val4
map[2] = val5
map[3] = val6
assert(map[1].value == 122)
```

ethanjli (commented):
TODO: review all files, exhaustively enumerate all possible internal states to set up as GIVENs for RingBuffer, e.g.:
```
[0* ] [1 ] [2 ] [3^ ] [   ]
[   ] [1*] [2 ] [3^ ] [   ]
[   ] [  ] [2*] [3^ ] [   ] 
[   ] [  ] [  ] [3*^] [   ]
```

ethanjli:
I believe there are a total of 25 cases to cover for a `RingBuffer` with `buffer_size == 5` (i.e. a capacity of 4 elements).

Notation:
- `*` indicates the element addressed by `oldest_index_`
- `^` indicates the element addressed by `newest_index_`

There are four types of "cells"/elements in the list of cases:
- `[* ^]` represents `oldest_index_ == newest_index_`, i.e. the buffer is empty
- `[*1 ]` represents "the oldest element in the buffer is 1"
- `[ n ]` represents an element in the buffer whose value is `n`
- `[  ^]` represents the element at `newest_index_`, which never has a value (it's always left as an empty space)

So a RingBuffer with template parameter `buffer_size` can store a maximum of `buffer_size - 1` elements. In either this PR or a separate one, we should consider changing the `buffer_size` template parameter to a `capacity` template parameter, where `buffer_size == capacity + 1`; to be clear, "capacity" means "the maximum number of elements we can _actually_ store in the container". This would be more consistent with our other containers, which all use `capacity` as the template parameter whenever `buffer_size != capacity`.

Cases:
```
Group 0: oldest_index_ == 0
[* ^] [   ] [   ] [   ] [   ]: newest_index == 0
[*1 ] [  ^] [   ] [   ] [   ]: newest_index == 1
[*1 ] [ 2 ] [  ^] [   ] [   ]: newest_index == 2
[*1 ] [ 2 ] [ 3 ] [  ^] [   ]: newest_index == 3
[*1 ] [ 2 ] [ 3 ] [ 4 ] [  ^]: newest_index == 4
Group 1: oldest_index_ == 1
[   ] [* ^] [   ] [   ] [   ]: newest_index == 1
[   ] [*1 ] [  ^] [   ] [   ]: newest_index == 2
[   ] [*1 ] [ 2 ] [  ^] [   ]: newest_index == 3
[   ] [*1 ] [ 2 ] [ 3 ] [  ^]: newest_index == 4
[  ^] [*1 ] [ 2 ] [ 3 ] [ 4 ]: newest_index == 0
Group 2: oldest_index_ == 2
[   ] [   ] [* ^] [   ] [   ]: newest_index == 2
[   ] [   ] [*1 ] [  ^] [   ]: newest_index == 3
[   ] [   ] [*1 ] [ 2 ] [  ^]: newest_index == 4
[  ^] [   ] [*1 ] [ 2 ] [ 3 ]: newest_index == 0
[ 4 ] [  ^] [*1 ] [ 2 ] [ 3 ]: newest_index == 1
Group 3: oldest_index_ == 3
[   ] [   ] [   ] [* ^] [   ]: newest_index == 3
[   ] [   ] [   ] [*1 ] [  ^]: newest_index == 4
[  ^] [   ] [   ] [*1 ] [ 2 ]: newest_index == 0
[ 3 ] [  ^] [   ] [*1 ] [ 2 ]: newest_index == 1
[ 3 ] [ 4 ] [  ^] [*1 ] [ 2 ]: newest_index == 2
Group 4: oldest_index_ == 4
[   ] [   ] [   ] [   ] [* ^]: newest_index == 4
[  ^] [   ] [   ] [   ] [*1 ]: newest_index == 0
[ 2 ] [  ^] [   ] [   ] [*1 ]: newest_index == 1
[ 2 ] [ 3 ] [  ^] [   ] [*1 ]: newest_index == 2
[ 2 ] [ 3 ] [ 4 ] [  ^] [*1 ]: newest_index == 3

```
This way of listing cases might make it easier to organize the test code: first you set up the empty ring buffer with `oldest_index` initialized to the correct location from 0 to 4 (by repeatedly pushing and popping an element `n` times, where `n` ranges from 0 to 4), and then you simply push between 0 and 4 elements into the ring buffer. This should initialize your GIVEN.

If there's any possible way we can unify some of the test code into reusable, parameterized functions (e.g. making functions consisting of `REQUIRE` assertions, and then each `THEN` calls a function with a parameter; and/or making a function to create an empty ring buffer with `oldest_index` initialized to the desired value; and/or making a function to create any one of our 25 initial states for the ring buffer; and/or making a function to push some specified number of elements; etc.), we should do that for maintainability.

raavilagoo:
@ethanjli for Vector.h file I am getting clang-tidy error - member initializer for 'size_' is redundant 
To suppress it, i was trying to use NOLINTNEXTLINE(bugprone-dynamic-static-initializers) , still getting the same error.
How should i suppress this?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Containers/Vector.h
```diff
@@ -34,8 +34,7 @@ class Vector {
   // initializer list (e.g. Vector<uint8_t, 5> vec{1, 2, 3})
   // Note: this makes copies of the values!
   // clang-tidy says - member initializer for 'size_' is redundant
-  // NOLINTNEXTLINE(bugprone-dynamic-static-initializers)
-  // NOLINTNEXTLINE(bugprone-dynamic-static-initializers,cert-err58-cpp)
+  // NOLINTNEXTLINE(modernize-use-default-member-init)
```
ethanjli: I thought the approach we decided to try was removing `size_` from the constructor's member initializer list and then suppressing the cppcheck as a false positive, using the same suppression statement as in https://github.com/pez-globo/pufferfish-software/blob/423136e8f9308412a72390f7b406984beb3dec5c/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Util/Containers/EnumMap.h#L42 , instead of suppressing clang-tidy - did that approach have problems?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
+SCENARIO('The EnumMap is constructed from initializer list works corretly') 
+  GIVEN('An EnumMap constructed with capacity 11, has {1, 1}, {2,2}, {3,3}, {4,4} items') 
+    WHEN('The size method is called')
+      THEN('The size method returns size as 4')
+
+    WHEN('The avaliable method is called ')
+      THEN('The avaliable method reports that 7 items are avlaible ')
```
ethanjli: It would be less ambiguous to say that there are 7 spaces available (since the map has 4 items and 7 spaces in it)

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
+SCENARIO('The EnumMap is constructed from initializer list works corretly') 
+  GIVEN('An EnumMap constructed with capacity 11, has {1, 1}, {2,2}, {3,3}, {4,4} items')
```
ethanjli: It would be clearer to specify the Key and Value types used for specifying the EnumMap template.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    WHEN('The full method is called')
+      THEN('The full method reports that the EnumMap is not full')
+
+    WHEN('The empty method is called')
+      THEN('The empty method reports that the EnumMap is not empty')
+
+    WHEN('The has method is called')
+      THEN('The has method returns true for 1-4 keys')
+      THEN('The has method returns false for [-1000]to[1000] keys')
```
ethanjli: It would be clearer to say that it returns false for every key from -1000 to 0, and separately that it returns false for every key from 5 to 1000.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    WHEN('Indexing thorugh EnumMap')
+      THEN('Indexing into the containers for keys 1, 2, 3, 4 returns the values 1, 2, 3, 4 respectively')
+
+    WHEN('The input method is called on 5th item with value as 5')
```
ethanjli: The EnumMap doesn't really have the concept of an "nth item". It would be clearer and more consistent to say that the input method is called with key 5 and value 5 (assuming that's what you meant).

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    WHEN('The has method is called')
+      THEN('The has method returns true for 1-4 keys')
+      THEN('The has method returns false for [-1000]to[1000] keys')
+
+    WHEN('Indexing thorugh EnumMap')
+      THEN('Indexing into the containers for keys 1, 2, 3, 4 returns the values 1, 2, 3, 4 respectively')
+
+    WHEN('The input method is called on 5th item with value as 5') 
+      THEN('The input method returns ok status for 5th item')
```
ethanjli: It's redundant to say "for the 5th item"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
...
+
+    WHEN('The input method is called on 5th item with value as 5') 
+      THEN('The input method returns ok status for 5th item') 
+      THEN('The max_size method reports, EnumMap has capacity 11') 
+      THEN('After the input method, the size method reports size as 5 ') 
+      THEN('The avaliable method reports that 6 items are avaliable') 
+      THEN('The full method reports that the EnumMap is not full') 
+      THEN('The empty method reports that the EnumMap is non-empty') 
+      THEN('The erase method reports ok status for 1st key ') 
+      THEN('The has method returns true for 1-5 keys')
```
ethanjli: It would be clearer to say "reports ok status when called with key 1", "returns true for keys 1-5", etc.

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/EnumMap.txt
```diff
@@ -0,0 +1,357 @@
...
+      THEN('The input method returns ok status for 5th item') 
+      THEN('The max_size method reports, EnumMap has capacity 11') 
+      THEN('After the input method, the size method reports size as 5 ') 
+      THEN('The avaliable method reports that 6 items are avaliable') 
+      THEN('The full method reports that the EnumMap is not full') 
+      THEN('The empty method reports that the EnumMap is non-empty') 
+      THEN('The erase method reports ok status for 1st key ') 
+      THEN('The has method returns true for 1-5 keys') 
+      THEN('After input method, itering through EnumMap returns expected sequence of data for 1-5 keys') 
+      THEN('The has method returns false for [-1000]to[1000] keys')
```
ethanjli: Similar to what I described above for a previous THEN, this should be split up into -1000 to 0 and 6 to 1000

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+  GIVEN('A RingBuffer with capacity 4 to which 0 bytes of data were pushed, 0 bytes were poped') 
+    // internal state should look like [*^] [] [] [] []
+    WHEN('1 byte of data is pushed ') 
+      THEN('The push method returns ok status') 
+      THEN('After push method ,the peek method returns expected data ') 
+        ////internal state should look like [0*] [^] [] [][]
+
+    WHEN('1 byte of data is poped ') 
+      THEN('The pop method returns empty status')
+        // internal state should look like [*^] [] [] [][]
```
ethanjli: We need a THEN to specify what the internal state should look like, e.g. "THEN the peek method returns empty on all indices"

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+    // internal state should look like [0*] [^] [] [][]
+    WHEN('1 byte of data is pushed') 
+      THEN('The push method returns ok status') 
+      THEN('After push method, the peek method returns expected data ') 
+        // internal state should look like [0*] [1] [^] [][]
+
+    WHEN('1 byte of data is poped') 
+      THEN('The pop method returns ok status') 
+      THEN('After the pop method,the peek method returns expected data') 
+        // internal state should look like [*^][][][][]
```
ethanjli: Shouldn't the internal state look like [][*^][][][]?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+      THEN('The push method returns ok status') 
+      THEN('After push method, peek method returns expected data') 
+        // internal state should look like [0*][1][2][^][]
+
+    WHEN('1 byte of data is poped ') 
+      THEN('The pop method returns ok status') 
+      THEN('After the pop method,the peek method returns expected data in RingBuffer') 
+        // internal state should look like [] [1*] [^] [][]
+
+  GIVEN('A RingBuffer with capacity 4 to which 3 bytes of data were pushed and 0 byte were poped')   //[][][][3*][4^]
```
ethanjli: Did you forget to delete "//[][][][3*][4^]"?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+    // internal state should look like [0*][1][2][^][]
+    WHEN('1 bytes of data are pushed ') 
+      THEN('The push method returns ok status') 
+      THEN('After the push method, peek method returns expected data') 
+        // internal state should look like [1*] [2] [3] [4][^]
+
+    WHEN('2 byte of data is poped ') 
+      THEN('The pop method returns ok status')
```
ethanjli: If you pop multiple bytes, you should specify that each pop method call returns ok status

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+
+    WHEN('2 byte of data is poped ') 
+      THEN('The pop method returns ok status') 
+      THEN('After the pop method,the peek method returns expected data in RingBuffer') 
+        // internal state should look like [][][2*][^][]
+
+  GIVEN('A RingBuffer with capacity 4 to which 4 bytes of data were pushed and 0 byte were poped')   // [0^][][][][4*]
+    // internal state should look like [0*] [1] [2] [3][^]
+    WHEN('1 byte of data is pushed ') 
+      THEN('The push method returns full status')
```
ethanjli: It may be clearer to call it "the full error code" here

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+    WHEN('2 byte of data is poped ') 
+      THEN('The pop method returns ok status') 
+      THEN('After the pop method,the peek method returns expected data in RingBuffer') 
+        // internal state should look like [][][2*][^][]
+
+  GIVEN('A RingBuffer with capacity 4 to which 4 bytes of data were pushed and 0 byte were poped')   // [0^][][][][4*]
+    // internal state should look like [0*] [1] [2] [3][^]
+    WHEN('1 byte of data is pushed ') 
+      THEN('The push method returns full status') 
+      THEN('After the push method, peek method returns expected data')
```
ethanjli, line 53: It may be clearer to also explicitly say that the pushed byte of data is not in the RingBuffer object

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+    // internal state should look like [0*] [1] [2] [3][^]
+    WHEN('1 byte of data is pushed ') 
+      THEN('The push method returns full status') 
+      THEN('After the push method, peek method returns expected data') 
+        // internal state should look like [0*] [1] [2] [3][^]
+
+    WHEN('4 bytes of data are poped ') 
+      THEN('The pop method returns ok status for 1st time') 
+      THEN('After the pop method,the peek method returns expected data in RingBuffer') 
+        // internal state should look like [*^] [] [] [][]
```
ethanjli: I believe the internal state should look like [][][][][*^]? I think it's important to say that because we should have a GIVEN where the internal state starts as "[][][][][*^]" - and so we can follow from one THEN to another GIVEN to something like an inductive proof of the correctness of the RingBuffer's behavior

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+
+  GIVEN('A RingBuffer with capacity 4 to which 2 bytes of data were pushed and 2 bytes was poped ')   // [0*][1][2^][][]
```
ethanjli: Did you mean to write "1 byte of data was pushed and 1 byte was popped"? Also, shouldn't this be a new scenario (for oldest index is 1)?

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+  GIVEN('A RingBuffer with capacity 4 to which 2 bytes of data were pushed and 1 byte was poped and 3 bytes were pushed')   // [0^][1*][][][]
+    // internal state should look like [^][1*][2][3][4]
+    WHEN('2 bytes of data are pushed ') 
+      THEN('The push method returns full status') 
+      THEN('After push method, peek method returns expected data') 
+        // internal state should look like [^][1*][2][3][4]
+
+    WHEN('1 byte of data is poped ') 
+      THEN('After the pop method,the peek method returns the data in RingBuffer') 
+        // internal state should look like [^][][2*][3][4]
```
ethanjli, line 113: As another example of THENs we need in order to break down "internal state should look like [^][][2*][3][4]" into concrete THENs: after the pop method call, it should be possible to push one more byte onto the ring buffer, but it should not be possible to push more than one byte onto the ring buffer

firmware/ventilator-controller-stm32/Core/Test/Pufferfish/Util/Containers/RingBuffer.txt
```diff
@@ -0,0 +1,592 @@
...
+  
+  GIVEN('A RingBuffer with capacity 4 to which 3 bytes of data were pushed and 3 byte was poped ')
```
ethanjli, line 116: The next scenario after oldest_index = 1 should be for oldest_index = 2; this looks like oldest_index = 3

ethanjli (approved):
I looked over some representative test cases for EnumMap and RingBuffer. There are many places where clarity could be improved, but that doesn't block this PR. There are also places where there seem to be errors which slipped past proofreading; but since this PR is so old, let's merge in the PR as soon as you feel it's ready, and you can resolve remaining issues later.

For records-keeping:

raavilagoo:
For records-keeping:

This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? - Yes
Were any of these contributions also part of work you did for an employer or a client? - No
Does this work include, or is it based on, any third-party work which you did not create? - No

PR #432, opened 2021-07-25
Change event type button style in events log

### LINKED ISSUE #427: Visually differentiate active alarms in the Events Log view
Currently, the events log view shows all active alarms the same way as regular log events. This makes it almost impossible to identify the sequence of events leading up to an alarm, and to understand the context around active alarms. We should have change the visual presentation of active alarms to make them easy to see among regular log events. It would be a good idea to not rely solely on color to mark active alarms. For example, maybe each row corresponding to an active alarm, we could encapsulate the alarm label of the "Event" column in a red rounded rectangle - this would establish a visual link to the red rounded rectangle in the topbar, to communicate that that row corresponds to an active alarm.

rohanpurohit:
This fixes #427

### DIFF
--- a/frontend/src/modules/controllers/SimpleTable.tsx
+++ b/frontend/src/modules/controllers/SimpleTable.tsx
@@ -133,7 +133,7 @@ const useStyles = makeStyles((theme: Theme) =>
       marginBottom: 'auto',
       left: 0,
       padding: '20px 24px 8px',
-      height: 'wrap',
+      height: '100%',
       display: 'flex',
     },
   }),
--- a/frontend/src/modules/controllers/SimpleTable.tsx
+++ b/frontend/src/modules/controllers/SimpleTable.tsx
@@ -133,7 +133,7 @@ const useStyles = makeStyles((theme: Theme) =>
       marginBottom: 'auto',
       left: 0,
       padding: '20px 24px 8px',
-      height: '100%',
+      height: 'wrap',
       display: 'flex',
     },
   }),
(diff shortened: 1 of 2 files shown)


rohanpurohit:
Few issues I see from testing:
- `AlarmLimits` events and `control (Ventilation etc)` had color as `${theme.palette.primary.main}` which is the blue color we have everywhere on frontend, it looks good when the button is `filled`, but is not quite `visible` after this PR, which is just a border.
1. we could maybe increase border size from `1px` to `2/4px`
2. we could change the color (to green or something that's clearly visible)

ethanjli:
Yes, let's increase the border so it looks thick. 4px would be ideal, unless it interferes with the text or makes the fill region look too thin

rohanpurohit:
> Yes, let's increase the border so it looks thick. 4px would be ideal, unless it interferes with the text or makes the fill region look too thin

I personally like how `4px` looks

ethanjli:
It looks good on the Raspberry Pi with the touchscreen display. I noticed that you've reduced the number of rows per page from 9 to 8, which is fine. Ideally we would have 10 rows per page since that's a nice round number, but since we already weren't at 10 rows per page, I wouldn't consider it a regression of this PR. A future visual improvement might be to adjust padding and margins for the event type button, since they're quite tall now that we've added 8 px of height to each one, but that might make things look visually cramped; a future solution might be to play around more with the styling in other ways so that we can make the things under the "Type" column less tall. I think this PR is fine to merge as it is right now.

For records-keeping:

rohanpurohit:

ethanjli (approved):

PR #434, opened 2021-07-29
Improve some of the driver interfaces for the Nonin sensor

rohanpurohit:
This PR includes:

- Device.h
  - split invalid_checksum status to invalid_header and invalid_checksum
- FrameBuffer.h
  - Add two different enums `BufferInputStatus` and `BufferOutputStatus`
  - add `output_ready` status for `input`
- FrameReceiver.h
  - change variable `start_of_frame_status_` to `at_start_of_frame`
  - split `validate_start_of_frame` into `validate_frame_header` and `validate_frame_checksum`
 
- Types.h
  - Add inner namespace FrameBytes for the 5 bytes in a frame: `header,status,pleth,data,checksum`
  - Add BIT0 of `status` byte mentioned as `SYNC` in datasheet in `StatusMasks`

In Cpp files resolved most of the unresolved comments from #421 except the refactor for `Sensor.cpp` which i will also do in this PR itself.


rohanpurohit:
@ethanjli I was working on the high-level driver and I'm not too sure when we should say the `sensor is connected`, and unfortunately I broke the `connector` that connects the nonin module to the MCU breadboard, so I won't be able to test the latest changes until I'm able to get the replacement for the connector wire. sorry about that!

ethanjli:
Ok, no worries! As for connectivity, I think there are a few levels and we'll need to report the different levels differently. We should consider the module connected to the MCU when we're receiving new packets/samples at the expected interval (I believe it's a few times per second). We should consider the finger sensor connected to the module when the module packets are reporting that the SNSD status flag is false. We should consider the patient's finger connected to the sensor when the module packets are reporting that SpO2 and HR are not at their "missing data" indicator valued (127 and 511, respectively) and simultaneously reporting that the OOT status flag is false.

rohanpurohit:
> Ok, no worries! As for connectivity, I think there are a few levels and we'll need to report the different levels differently. We should consider the module connected to the MCU when we're receiving new packets/samples at the expected interval (I believe it's a few times per second). We should consider the finger sensor connected to the module when the module packets are reporting that the SNSD status flag is false. We should consider the patient's finger connected to the sensor when the module packets are reporting that SpO2 and HR are not at their "missing data" indicator valued (127 and 511, respectively) and simultaneously reporting that the OOT status flag is false.

Thanks! I'm thinking that would mean changing/adding more return statuses for `Device` class or `PacketStatus` as we currently return waiting for most of these, and the logic for all of them are handled in these separate classes, we might also need to look at that again. as far as reporting missing data is concerned, If I'm not wrong I think we currently ignore if we get these bytes, I'll confirm once again.

rohanpurohit:
> Ok, no worries! As for connectivity, I think there are a few levels and we'll need to report the different levels differently. We should consider the module connected to the MCU when we're receiving new packets/samples at the expected interval (I believe it's a few times per second). We should consider the finger sensor connected to the module when the module packets are reporting that the SNSD status flag is false. We should consider the patient's finger connected to the sensor when the module packets are reporting that SpO2 and HR are not at their "missing data" indicator valued (127 and 511, respectively) and simultaneously reporting that the OOT status flag is false.

Thanks! I have made some changes to `Sensor.cpp` but it's still WIP and we can discuss it in our meeting tomorrow.

firmware/ventilator-controller-stm32/Core/Src/main.cpp
```diff
@@ -650,8 +650,12 @@ int main(void)
 
     // Independent Sensors
     fdo2.output(hfnc.sensor_vars().po2);
-    nonin_oem.output(store.sensor_measurements_raw().spo2, store.sensor_measurements_raw().hr);
-    PF::Driver::Serial::Nonin::SensorAlarmsService::transform(sensor_connections, alarms_manager);
+    auto nonin_status =
+        nonin_oem.output(store.sensor_measurements_raw().spo2, store.sensor_measurements_raw().hr);
```
ethanjli: Seems like we need to add `sensor_connections` back as an output parameter.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/SensorAlarmService.cpp
```diff
@@ -9,30 +9,32 @@
...
-
-  sensor_connections.sensor_alarm
-      ? alarms_manager.activate_alarm(Application::LogEventCode_nonin_disconnected, system)
-      : alarms_manager.deactivate_alarm(Application::LogEventCode_nonin_disconnected);
+    InitializableState status,
+    const SensorConnections &sensor_connections,
+    Application::AlarmsManager &alarms_manager) {
+  status == InitializableState::failed
+      ? alarms_manager.activate_alarm(LogEventCode::LogEventCode_nonin_disconnected, system)
+      : alarms_manager.deactivate_alarm(LogEventCode::LogEventCode_nonin_disconnected);
```
ethanjli: TBD: do we need to return early if status is `failed`?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Sensor.h
```diff
@@ -15,35 +15,43 @@
...
-  explicit Sensor(Device &device) : device_(device) {}
+  explicit Sensor(Device &device, HAL::Interfaces::Time &time) : device_(device), time_(time) {}
 
   InitializableState setup() override;
   InitializableState output(float &spo2, float &hr);
-  InitializableState measure(Sample measurements);
 
  private:
-  static bool find(const Flags &measurement, const bool &expected);
+  static const uint32_t waiting_duration = 2;  // ms
```
ethanjli: Rename to "measurement_timeout", "sample_timeout", or something similar. Also, this should be 2000 ms, not 2 ms.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
   }
 
+  if (wait_time_exceeded()) {
+    return InitializableState::failed;
+  }
+
   return InitializableState::setup;
 }
 
 InitializableState Sensor::output(float &spo2, float &hr) {
```
ethanjli: We should return early if last state was not `InitializableState::ok`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
 // Sensor
 
 InitializableState Sensor::setup() {
```
ethanjli, line 21: We should return early if last state was not `InitializableState::setup`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
+  }
+
   return InitializableState::setup;
 }
 
 InitializableState Sensor::output(float &spo2, float &hr) {
-  if (!sensor_connections_.nonin_connected) {
+  switch (device_.output(measurements_)) {
+    case PacketStatus::invalid_checksum:
+    case PacketStatus::invalid_header:
```
ethanjli, line 57: TBD: should we just handle these the same way as `waiting` and `frame_loss`?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
+      // handle error cases first
+      spo2 = NAN;
+      hr = NAN;
+      return InitializableState::failed;
+    case PacketStatus::waiting:
+    case PacketStatus::frame_loss:
+    case PacketStatus::ok:
```
ethanjli, line 66: `ok` seems like a special case to handle differently from the other cases. (TBD: same with setup method?)

ethanjli: `ok` should reset the waiting timeout timer. (same with setup?)

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-  if (measure(measurements_) != InitializableState::ok) {
+  // measurements status
+  sensor_connections_.sensor_disconnected = find(measurements_.sensor_disconnect);
+  sensor_connections_.sensor_alarm = find(measurements_.sensor_alarm);
+  sensor_connections_.out_of_track = find(measurements_.out_of_track);
+
+  if (measurements_.e_spo2_d == ErrorConstants::spo2_missing) {
     spo2 = NAN;
-    hr = NAN;
     return InitializableState::failed;
```
ethanjli: We should return `ok` in this case; we shouldn't return early.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-  sensor_connections_.hr_out_of_track = (measurements_.e_hr == ErrorConstants::hr_missing ||
-                                         measurements_.e_hr_d == ErrorConstants::hr_missing) &&
-                                        find(measurements_.out_of_track, true);
+  if (measurements_.e_hr_d == ErrorConstants::hr_missing) {
+    hr = NAN;
+    return InitializableState::failed;
+  }
 
   spo2 = measurements_.e_spo2_d;
   hr = measurements_.e_hr_d;
```
ethanjli: These will become else cases on the above if statements.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -14,65 +14,80 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
+    case PacketStatus::invalid_checksum:
+    case PacketStatus::invalid_header:
       return InitializableState::failed;
-    case InitializableState::ok:
-      sensor_connections_.nonin_connected = true;
-      return InitializableState::ok;
-    case InitializableState::setup:
+    case PacketStatus::waiting:
+    case PacketStatus::frame_loss:
+    case PacketStatus::ok:
```
ethanjli: We should return `ok` afterwards.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
ethanjli: Pseudocode:

```
output:
    if we already failed, return failed
    if we're in setup, return setup
    
    if it has been too long since we last received valid data, return failed
    
    // i.e. it hasn't been too long since we last received valid data:
    if we haven't received valid data, return ok early
    
    // i.e. we've received valid data:
    reset the timer
    write the output params
```

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/Sensor.h
```diff
@@ -15,35 +15,43 @@
...
+  static bool find(const Flags &measurement);
```
ethanjli: Let's move find to the cpp file, rename to be more descriptive (e.g. `any_true` or something)

ethanjli (commented):
Notes from today's review of high-level code:

Let's merge after changes are addressed and we do some testing. We'll save refactors/improvements of low-level driver code for a future PR. I'll review the low-level refactor code, and anything besides fixes for regressions should be addressed in future PRs.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameBuffer.h
```diff
@@ -34,12 +34,17 @@ namespace Pufferfish::Driver::Serial::Nonin {
 /* Frame */
 using Frame = std::array<uint8_t, frame_max_size>;
 
-enum class FrameBufferStatus {
+enum class BufferInputStatus {
   ok = 0,
-  waiting,
+  output_ready,
   full,
```
ethanjli, line 40: Were we planning to remove `full` and replace it with an `input_overwritten` output parameter in `FrameBuffer`, or was that only for `BufferReceiver`?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameReceiver.h
```diff
@@ -70,16 +71,16 @@ class FrameReceiver {
...
+  // if the bytes at the start of a frame are valid, set 'at_start_of_frame_' to 'true',
+  // and therefore we can read the subsequent bytes.
+  // Inititally set 'at_start_of_frame_` to 'false',
+  // so that we can 'validate' if there is noise or loss of bytes in the frame.
+  bool at_start_of_frame_ = false;
 };
 
-extern bool validate_start_of_frame(const Frame &new_frame);
+extern bool validate_frame_header(const Frame &new_frame, const uint8_t &mask);
+extern bool validate_frame_checksum(const Frame &new_frame);
```
ethanjli: We don't need `extern`, right? Also, should we move the function declarations into the `cpp` file, or would it be better to leave them in the `h` file to expose for unit tests?

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/PacketReceiver.h
```diff
@@ -96,4 +96,6 @@ class PacketReceiver {
 extern void read_status_byte(
     Sample &sensor_measurements, const size_t &frame_index, const uint8_t &byte_value);
 
+extern bool check_packet_sync(const Frame &frame);
```
ethanjli: We don't need `extern` here either, right?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -26,17 +26,16 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-  static const uint8_t mask_start_of_packet = 0x81;
-  if (new_frame[0] == 0x01 && (new_frame[1] & mask_start_of_packet) == mask_start_of_packet) {
-    if (((new_frame[0] + new_frame[1] + new_frame[2] + new_frame[3]) % (UINT8_MAX + 1)) ==
-        new_frame[4]) {
-      return true;
-    }
-  }
-  return false;
+bool validate_frame_header(const Frame &new_frame, const uint8_t &mask) {
+  return (new_frame[0] == 0x01 && (new_frame[1] & mask) == mask);
```
ethanjli, line 32: `(new_frame[1] & mask) == mask` looks a bit odd to me, is there a more concise way of expressing this, or is this the clearest approach for the set of masks we're using with this function?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -26,17 +26,16 @@ namespace Pufferfish::Driver::Serial::Nonin {
...
-bool validate_start_of_frame(const Frame &new_frame) {
-  static const uint8_t mask_start_of_packet = 0x81;
-  if (new_frame[0] == 0x01 && (new_frame[1] & mask_start_of_packet) == mask_start_of_packet) {
-    if (((new_frame[0] + new_frame[1] + new_frame[2] + new_frame[3]) % (UINT8_MAX + 1)) ==
-        new_frame[4]) {
-      return true;
-    }
-  }
-  return false;
+bool validate_frame_header(const Frame &new_frame, const uint8_t &mask) {
```
ethanjli: We can just use a `uint8_t` mask instead of a `const uint8_t`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -45,46 +44,49 @@ bool validate_start_of_frame(const Frame &new_frame) {
  */
 FrameInputStatus validate_frame(const Frame &new_frame) {
   static const uint8_t mask_start_of_frame = 0x80;
```
ethanjli: Since this mask is selecting certain bits, let's use a `0b` literal rather than a `0x` literal.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -45,46 +44,49 @@ bool validate_start_of_frame(const Frame &new_frame) {
...
 
-  return FrameInputStatus::checksum_failed;
+  if (!validate_frame_checksum(new_frame)) {
+    return FrameInputStatus::invalid_checksum;
+  }
+  return FrameInputStatus::output_ready;
 }
 
 FrameInputStatus FrameReceiver::update_frame_buffer(uint8_t new_byte) {
+  static const uint8_t mask_start_of_packet = 0x81;
```
ethanjli: Since this mask is selecting certain bits, let's use a `0b` literal rather than a `0x` literal.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -45,46 +44,49 @@ bool validate_start_of_frame(const Frame &new_frame) {
  */
 FrameInputStatus validate_frame(const Frame &new_frame) {
```
ethanjli: If we only use this function in one place, it will probably be more elegant to just remove this function and move the two `if-then-return early` checks into the `update_frame_buffer` method - as right now the `return FrameInputStatus::output_ready` is a bit weird, since frame validation should have no awareness of the concept of "output"s, and the `update_frame_buffer` method is having to do some extra checks about the value returned by `validate_frame`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -45,46 +44,49 @@ bool validate_start_of_frame(const Frame &new_frame) {
...
+  static const uint8_t mask_start_of_packet = 0x81;
   Frame frame_buffer;
 
-  if (frame_buf_.input(new_byte) == FrameBufferStatus::waiting) {
+  if (frame_buf_.input(new_byte) == BufferInputStatus::ok) {
     return FrameInputStatus::ok;
   }
 
-  if (frame_buf_.output(frame_buffer) != FrameBufferStatus::ok) {
+  if (frame_buf_.output(frame_buffer) != BufferOutputStatus::ok) {
```
ethanjli: Maybe it would be better for this to explicitly check `== BufferOutputStatus::waiting`, or else to have a `switch` statement covering both of the possible return code values as cases?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
```diff
@@ -45,46 +44,49 @@ bool validate_start_of_frame(const Frame &new_frame) {
...
 }
 
 FrameInputStatus FrameReceiver::update_frame_buffer(uint8_t new_byte) {
+  static const uint8_t mask_start_of_packet = 0x81;
   Frame frame_buffer;
 
-  if (frame_buf_.input(new_byte) == FrameBufferStatus::waiting) {
+  if (frame_buf_.input(new_byte) == BufferInputStatus::ok) {
     return FrameInputStatus::ok;
   }
```
ethanjli, line 52: It looks like if `frame_buf_.input(new_byte)` returned `BufferInputStatus::full`, we still proceed as if everything is normal. Do we want that behavior?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
ethanjli: If clang-tidy allows us, let's move this declaration to be right above the `if` check where it's needed, i.e. within the outer `if` branch, to limit the scope of this declaration.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/PacketReceiver.cpp
```diff
@@ -40,49 +40,58 @@ void read_status_byte(
...
   } else if ((byte_value & StatusMasks::rprf) == StatusMasks::rprf) {
-    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::red_perfusion;
+    signal_perfusion = SignalQuality::red_perfusion;
   } else if ((byte_value & StatusMasks::gprf) == StatusMasks::gprf) {
-    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::green_perfusion;
+    signal_perfusion = SignalQuality::green_perfusion;
   } else {
-    sensor_measurements.signal_perfusion[frame_index] = SignalQuality::no_perfusion;
+    signal_perfusion = SignalQuality::no_perfusion;
   }
```
ethanjli: I know I had suggested doing things this way, but I've just realized an even more readable approach which doesn't require us to assign to a `SignalQuality &`:

```
SignalQuality signal_perfusion = SignalQuality::no_perfusion;
if (...) {
  signal_perfusion = ...;
} else if (...) {
  signal_perfusion = ...;
} else if (...) {
  signal_perfusion = ...;
}
sensor_measurements.signal_perfusion[frame_index] = signal_perfusion;
```

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/SensorAlarmService.cpp
```diff
@@ -0,0 +1,45 @@
...
+void SensorAlarmsService::transform(
+    InitializableState status,
+    const SensorConnections &sensor_connections,
+    Application::AlarmsManager &alarms_manager) {
+  if (status == InitializableState::failed) {
+    alarms_manager.activate_alarm(LogEventCode::LogEventCode_nonin_disconnected, system);
+    return;
+  }
+
+  if (status != InitializableState::failed) {
```
ethanjli: This `if` check is redundant, right? We only reach line 24, if `status != ...::failed`

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/SensorAlarmService.cpp
```diff
@@ -0,0 +1,45 @@
...
+            LogEventCode::LogEventCode_nonin_finger_sensor_disconnected);
+
+  sensor_connections.sensor_alarm
+      ? alarms_manager.activate_alarm(LogEventCode::LogEventCode_nonin_sensor_alarm, system)
+      : alarms_manager.deactivate_alarm(LogEventCode::LogEventCode_nonin_sensor_alarm);
+
+  sensor_connections.out_of_track
+      ? alarms_manager.activate_alarm(
+            LogEventCode::LogEventCode_nonin_out_of_track_measurements, system)
+      : alarms_manager.deactivate_alarm(LogEventCode::LogEventCode_nonin_out_of_track_measurements);
```
ethanjli: Do you think the use of the ternary if expression (` test ? branch : branch`) is clearer than using explicit `if () {...} else {...}`? Usually I only use the ternary if expression as the right-hand side of variable assignments.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -9,39 +9,82 @@
...
-  float spo2 = NAN;
-  float hr = NAN;
-  return output(spo2, hr);
+  if (prev_state_ == InitializableState::failed) {
+    return InitializableState::failed;
+  }
+
+  switch (device_.output(measurements_)) {
+    case PacketStatus::ok:
+      waiting_timer_.reset(time_.millis());
```
ethanjli, line 28: Is there any kind of range checking we could/should do for the Nonin sensor's setup routine if we get measurements from the sensor?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
```diff
@@ -9,39 +9,82 @@
...
-InitializableState Sensor::output(float &spo2, float &hr) {
-  if (!setup_) {
+InitializableState Sensor::output(SensorConnections &sensor_connections, float &spo2, float &hr) {
+  if (prev_state_ == InitializableState::setup) {
+    return InitializableState::setup;
+  }
+
+  if (prev_state_ == InitializableState::failed) {
     return InitializableState::failed;
   }
```
ethanjli: Would it be equivalent to replace this with `if (prev_state_ != InitializableState::ok) {return prev_state_;}` ?

ethanjli (commented):
Here's my review of all changes in this PR!

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameReceiver.h
rohanpurohit: I think it would be good to expose them for unit tests!

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
rohanpurohit: No! you're right, I have now added a switch block to deal with all 3 cases.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/SensorAlarmService.cpp
rohanpurohit: Yes so when the `status != ...:failed`, we would still need to deactivate the alarm right? since we are returning early if it's failed and not checking any other `status codes`

rohanpurohit:
@ethanjli This pr is ready for testing. after which we can do one final review depending on test results

Tests:
- We should be testing if we get normal output or for some reason, we return failed and display `NaN` instead.
- We should see if we get `Nonin sensor disconnected` on disconnecting the sensor from mcu.
- On disconnecting the finger sensor, is there an alarm for `finger sensor disconnected`.
- If there are alarms for `out_of_track` and `sensor_alarm` which are unlikely and mostly when the sensor is incapable of providing good measurements.

anything else you might think of while testing, do let me know!
thanks.

firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/FrameReceiver.h
```diff
@@ -79,9 +79,7 @@ class FrameReceiver {
   bool at_start_of_frame_ = false;
 };
 
-extern bool validate_frame_header(const Frame &new_frame, const uint8_t &mask);
-extern bool validate_frame_checksum(const Frame &new_frame);
-
-extern FrameInputStatus validate_frame(const Frame &new_frame);
+extern bool validate_frame_header(const Frame &new_frame, uint8_t mask);
```
ethanjli: Do we still need `extern` here?

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
ethanjli: Do we want to treat `full` the same as `output_ready`? (I'm asking because I've forgotten what the desired error-handling behavior should be in this situation)

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
ethanjli: If `prev_state_` is already `ok`, then we should do an early return of `ok`; if `prev_state_` is already `failed`, then we should do an early return of `failed`; if `prev_state_` is `setup`, then we should proceed to run the setup code. In other words, we can just do `if (prev_state_ != InitializableState::setup) {return prev_state_;}`.

The `Sensor` object starts out with `prev_state_` initialized as `setup`, so `prev_state_ != InitializableState::ok` when the firmware first starts. Your change below causes `Sensor::setup` to return `failed` the first time we call `setup`

ethanjli (commented):
It doesn't work yet. Observations from testing:

- If I set the list of initializables to `PF::Driver::make_initializables(sfm3019_air, sfm3019_o2)`, then when I run the application backend and the frontend I get an "Internal sensor failed: SpO2 and HR sensor connection lost" alarm, but I'm still able to start ventilation and the firmware supplies simulated values for SpO2 and HR. I could see arguments both ways on whether we want to display this message if `nonin_oem` is commented out of the initializables list: one the one hand, it is indeed true that the sensor is not connected, so this is a safeguard against incorrect configuration of the firmware before uploading it to the MCU. On the other hand, the sensor was intentionally disconnected so maybe we don't want the alarm? If we want to remove it, perhaps we can check `nonin_oem.output(...) != InitializableState::failed` rather than `nonin_oem.output(...) != InitializableState::ok`. But maybe we want to keep this alarm in this situation.
- If I set the list of initializables to `PF::Driver::make_initializables(sfm3019_air, sfm3019_o2, nonin_oem)`, then the firmware fails setup and makes LED1 flash very rapidly on the Nucleo board. I've pointed out the cause of this and the fix under https://github.com/pez-globo/pufferfish-software/pull/434/files/780f4ea3e2c8c190d9cc38f46910465e49f86c5e..6ac7b189bd40fd0c61d26589f80f0be36314b570#r685664193 . With that fix applied, the firmware successfully passes setup/initialization. However, I still get the "Internal sensor failed: SpO2 and HR sensor connection lost" alarm, and when I start ventilation I see simulated SpO2 and HR values from the firmware. When I step through `Sensor::output` with a debugger, I notice that the first return of `device_.output(...)` is `PacketStatus::frame_loss`, which is entirely reasonable; however, the waiting timer reports that we're past the timeout. Because using the debugger messes with the timing, I'm not completely confident - but my guess is that we may have some problems with initialization of `waiting_timer_`, as `main.cpp`'s setup code may be waiting for a longer time for other sensors to finish initialization, even after the Nonin `Sensor::setup()` reset its `waiting_timer_` and returned `InitializableState::ok`. Because other code is allowed to call `time_.delay(...)` for relatively long times (e.g. 2 ms, maybe more), I think we'll have to perform a `waiting_timer_.reset(...)` after setup has completed on all `Initializable`s and before we enter the main control loop. For example, maybe in our `main.cpp`, before we enter our `while (true) {...}` "normal loop", we'll first need to call `initializables.after_setup()` which iterates through all initializable objects and calls each object's `after_setup()` method, which should be non-blocking and run exactly once for each object. Let's discuss this more in today's call, there are probably other viable or cleaner approaches (and better method names) which I haven't thought of yet.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/FrameReceiver.cpp
ethanjli: It would be helpful to add a code comment explaining why we want to break after `BufferInputStatus::full`, and how this interacts with the behavior of `FrameBuffer`.

firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp
rohanpurohit: will be tracked in https://github.com/pez-globo/pufferfish-software/issues/440

ethanjli:
Here are my results from testing:
- The changes made have not yet fixed the problem. However, they did make it possible for me to better understand what's happening via the STM32 debugger. If I set two breakpoints - one inside the `post_setup_reset` method, and the other for `prev_state_ = InitializableState::failed` inside the `output` method, then I see that I first hit the `prev_state_ = InitializableState::failed` before I hit the breakpoint inside `post_setup_reset`. This would suggest that we're calling the `output` method as part of the main function's Setup loop, i.e. after the Nonin Sensor has completed its `setup` routine but before the other sensors have completed their `setup` routines (during which they may sleep for > 2 ms), the Nonin `Sensor::output` method is being called. It turns out that this is caused by the main function's "Configure the simulators" section, specifically https://github.com/pez-globo/pufferfish-software/blob/10cabafc474ea24cc2d6da6d12a0f698a7a6c1f0/firmware/ventilator-controller-stm32/Core/Src/main.cpp#L621 , being before the "Reset nonin timer" section. If I swap the order of the two sections, then I am able to get some data from the Nonin sensor. The rest of this test report is made after applying this fix.
- Then what I observe is that on the dashboard, with my finger not in the Nonin sensor, SpO2 is displayed as "--" good, while HR is displayed as "127" (this looks suspiciously like a missing data value, except the Nonin datasheet says the HR missing data indicator is 511 while the SpO2 missing data indicator is 127). Once I put my finger in the sensor, after the expected delay I start getting reasonable SpO2 and HR readings (e.g. 98%, 83 BPM). Once I remove my finger from the sensor, after the expected delay the SpO2 goes back to "--" - and then the HR goes back to 127. I'm not sure what's happening here.
- I also did 15 trials of unplugging the Nucleo board from USB and plugging it back in. In 5 of those trials, the Nucleo board failed setup. The seemingly random pattern suggests to me that maybe our measurement timeout is too low for some reason. When I increase the measurement timeout from 2 sec to 5 sec and repeat 25 trials, the Nucleo board fails setup in 0 of those trials. For this PR we should just increase the measurement timeout; however, once you have the Nonin hardware working on your end we'll need to investigate to understand why a Nonin measurement timeout of 2 sec could be causing the setup routine to randomly fail sometimes. It's probably some kind of interaction with the setup methods of the other drivers, so you could begin by trying to combine the Nonin driver with each of the other drivers in the initialization list, one at a time, to see if the random occurrence of Nonin measurement timeouts can be reproduced with just a single other driver.

Conclusion: just a few small changes and I'm pretty confident that it will work enough for us to merge this PR.

For records-keeping: 

rohanpurohit:
@ethanjli one interesting thing to note is in the previous code that you wrote for Sensor.cpp, for both HR and SpO2 you checked for `127` which is the missing data for `SpO2` [Sensor.cpp](https://github.com/pez-globo/pufferfish-software/blob/49328d7e49bcc75f267609cbb9d34c713fc9eed1/firmware/ventilator-controller-stm32/Core/Src/Pufferfish/Driver/Serial/Nonin/Sensor.cpp#L17), and it matches to what we are observing, which would definitely fix that bug! but of course, we'll need to figure out why that happens.

I guess this is the code block that's doing that change [PacketReceiver L63](https://github.com/pez-globo/pufferfish-software/blob/95833b1adcf3d9255a406c8b876d4819acc2f9c4/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/PacketReceiver.h#L63) and looks like that is what we want since the LSB is for SpO2
 
ethanjli:
TODO for me: test, and see if it's easy to fix the suspicious HR MSB processing at https://github.com/pez-globo/pufferfish-software/blob/95833b1adcf3d9255a406c8b876d4819acc2f9c4/firmware/ventilator-controller-stm32/Core/Inc/Pufferfish/Driver/Serial/Nonin/PacketReceiver.h#L63

TODO for you (this work does not block this PR): investigate whether FrameBuffer should behave more like RingBuffer or more like ChunkSplitter in its interface for reporting `input_overwritten` (i.e. as a return code or as an output parameter)

ethanjli:
I've pushed up a commit completing the fix of the HR parsing code, as well as adding definitions to the frontend's logs/constants.ts for handling the new events you added. With these fixes, everything seems to work; there's a very noticeable delay between when I put my finger in the sensor and when the frontend shows numbers, but I haven't checked whether this is the same delay as before or a larger delay.

For records-keeping on my contributions:

ethanjli (approved):
You'll need to update the unit test code, but after it's done this is ready to merge.

ethanjli:
We concluded that FrameBuffer shouldn't modify any internal state if it's already full, it should just return early. This is like the interface of RingBuffer. By contrast, ChunkSplitter clears the internal buffer if it's full and another input is overwritten. There it's reasonable because dropping one frame doesn't affect the state of the protocol, since the protocol doesn't really have connection state at the levels above frames. For the Nonin sensor, there is connection state at the packet level, because each packet is a sequence of 25 consecutive frames and dropping one will invalidate the packet anyways - so clearing the internal buffer wouldn't help error recovery.

PR #442, opened 2021-09-13
Reorganize some selectors and reducers

rohanpurohit:
This PR includes:
- Resolves some TODO'S in redux store
- implement screen lock in the simulator
- implement screen lock in the firmware

### DIFF
--- a/frontend/src/store/controller/selectors/frontend_pb.ts
+++ b/frontend/src/store/controller/selectors/frontend_pb.ts
@@ -1,0 +1,60 @@
+import { createSelector } from 'reselect';
+import {
+  SystemSettings,
+  SystemSettingsRequest,
+  FrontendDisplaySetting,
+} from '../../proto/frontend_pb';
+import { getController } from '../selectors';
+import { ControllerStates, RotaryEncoderParameter, SystemSettingsRequestResponse } from '../types';
+
+// RotaryEncoder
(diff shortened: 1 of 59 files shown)


frontend/src/store/controller/selectors/frontend_pb.ts
```diff
@@ -0,0 +1,60 @@
+import { createSelector } from 'reselect';
```
ethanjli, line 1: The naming of this file is a bit awkward; but because in the future we'll want to move selectors out to feature folders (e.g. a feature folder for the rotary encoder behaviors, maybe a feature folder for system settings, etc.), the filename you chose is fine for this PR.

ethanjli (commented):
Code changes look good to me!

Things to implement in this PR or a future PR:
- Screen locking simulator in the firmware
- For screen locking simulators, also lock and unlock the screen from the firmware at regular or random times (simulating the effect of pressing the screen lock toggle membrane button connected to the interface board for the MCU)

TODO: test the screen locking behavior so I understand how it behaves in the backend simulator and in the firmware, and then approve for merging.

rohanpurohit:

rohanpurohit:
Tested this branch on the backend simulator, while checking the redux console for any regressions and accessed standard features of the frontend and there are no regressions that I faced, merging this now.

PR #446, opened 2021-09-28
Misc refactor

rohanpurohit:
This PR includes:
- Rename eventLogs to eventsLog
- Rename AppConstants.tsx to constants.ts
- embed OverlayScreen at App level
- decompose audio alarms from overlayscreen
- rename overlayscreen to screenlock
- update props for paramValueSpinner to include valuespinners in displayTab as well

PR #447, opened 2021-10-08
Feature/frontend remove rxjs

rohanpurohit:
This PR:
- Remove RxJs from our stack and use redux instead for all those functionalities.


rohanpurohit:
@ethanjli current status of this PR:
- ScreenLock which used RxJs now works with redux
- Discard Alarm Limits Modal popup works with redux
- MultiStep Wizard opening and closing works (**but it clearly has bugs**, seems like submit, cancel are not discarding state, I did not get time to properly understand the bug!)
so maybe it can be fixed by someone else/you or it could be a blocker for this PR.
- Rotary encoder highlighting (I could not test it) so the status is unknown! which could again be a blocker!

I guess changes in this PR could be reviewed more carefully and worked on by someone once #446 is merged, which is maybe good to merge from my side, if there are no regressions introduced in that PR.

anyway, I guess this will be the last change by me on this PR.

PR #449, opened 2021-12-02
I2CDeviceStatus added in read and write method of I2CDevice.cpp

raavilagoo:
Changes in I2CDevice.cpp file, add_read method modified, status is given as input parameter. Status is stored in read status queue.
In case of write method, add_write_status is called, to add status, which gets stored in write status queue.
Checks added , 1. queue size of read buf and read status buf should be same. 2. Count and buf size should be equal. 
Test cases of Mock I2CDevice.
Modified test cases for Sensor.cpp and Device.cpp according to I2CDevice latest changes.For Device.cpp and Sensor.cpp following error was seen, "fatal error {Unknown expression after the reported line}" , after changing count from 254 to number of bytes of data in read method, this issue was resolved. Couldn't recreate this issue.


raavilagoo:
This PR passes clang-tidy and clang-format checks.
For records-keeping:

This project is licensed under Apache License v2.0 for any software, and Solderpad Hardware License v2.1 for any hardware - do you agree that your contributions to this project will be under these licenses, too? - Yes
Were any of these contributions also part of work you did for an employer or a client? - No
Does this work include, or is it based on, any third-party work which you did not create? - No

